PR #1, opened 2017-02-09
Overhaul for dynamic inventory

thoraxe:
There is a lot of cleanup and formatting here to eliminate deprecation warnings, however, the meat of what was done was to "fix" the `main.yml` and other plays to fully leverage EC2 dynamic inventory.

For this to work with the existing opentlc-shared OpenShift installation, you will need to ensure that certain tags exist on EC2 instances. You can look at the bu-workshop cloudformation template to see what tags should be there and what their formats are.

These tags make dynamic inventory work to its fullest potential.


thoraxe:
Whoops... not quite there yet. Probably one more commit incoming.

thoraxe:
Testing this now.

thoraxe:
OK -- my install completed successfully. I think this would work for opentlc-shared if you adjusted the CF template @sborenst

PR #52, opened 2017-07-25
create role opentlc-integration + fix ansible-provisioner

fridim:
- Use existing role that was in configs/ocp-workshop/roles
- move it to top directory
- use it only in ocp-workshop for now
- replace all shell tasks with their ansible equivalent
- add install_opentlc_integration variable (default true, keep behavior)


fridim:
use the role opentlc-integration in ansible-provisioner +  fix ansible-provisionner + DRY role opentlc-integration

- fix CloudFormation template: remove DependsOn InternetGatewayAttachment,
  use var vpcid_name_tag
- DontRepeatYourself: use a variable symlinks for role opentlc-integration
- provisioner: fix post-software  (replace shell with ansible equivalent)

PR #37, opened 2017-05-26
Overhaul of the ssh config file generation

thoraxe:
* Stops including the bastion as a host that should be proxied (causes a
loop)

* adds a config directive to the proxycommand (to properly pick up the
bastion host config)

I'm testing this right now on our provisioning host. I copied the /opt/ansible_agnostic_deployer folder to /opt/dev_dep and checked out this branch. It's not having SSH issues and it is up to the installation of openshift node packages.

I'll wait for the complete installation to finish, but it's looking promising right now.

PR #53, opened 2017-07-26
make ec2_infrastructure_deployment agnostic

fridim:
The play 'Configure local ssh config for bastion proxy use' is specific to OCP
environments. If there is no bastion host, the playbook fails with the message:

```
TASK [Store bastion hostname as a fact] ***************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute u'tag_ansible_provisioner_gucoretest_bastion'\n\nThe error appears to have been in '/home/user/Git/ansible_agnostic_deployer/ansible/cloud_providers/ec2_ssh_config_setup.yml': line 14, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n     var: \"{{hostvars}}\"\n  - name: Store bastion hostname as a fact\n    ^ here\n"}
```

Fix:
- add conditional to include based on the presence of bastion host

### DIFF
--- a/ansible/cloud_providers/ec2_infrastructure_deployment.yml
+++ b/ansible/cloud_providers/ec2_infrastructure_deployment.yml
@@ -72,6 +72,7 @@
 
 - name: Configure local ssh config for bastion proxy use
   include: "{{ ANSIBLE_REPO_PATH }}/cloud_providers/{{cloud_provider}}_ssh_config_setup.yml"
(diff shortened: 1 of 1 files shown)


fridim:
related to #37

PR #55, opened 2017-07-28
opentlc-integration: fix symlinks

fridim:
restore the original path for symlinks before PR #52

### DIFF
--- a/ansible/configs/ansible-provisioner/post_software.yml
+++ b/ansible/configs/ansible-provisioner/post_software.yml
@@ -19,9 +19,12 @@
   gather_facts: False
   vars:
     symlinks:
-      - provision_workshop_env.sh
-      - provision-ose-projects.sh
-      - deploy_scripts
+      - src: /opt/OPEN_Admin/OPENTLC-OCP3/provision_workshop_env.sh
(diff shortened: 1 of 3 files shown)


fridim:
@sborenst  The playbook currently fails with message:

```
TASK [/home/user/Git/ansible_agnostic_deployer/ansible/roles/opentlc-integration : Create symlinks to ~opentlc-mgr/bin] *******************************************************************************************
failed: [ec2-52-28-76-182.eu-central-1.compute.amazonaws.com] (item=provision-accounts.sh) => {"failed": true, "gid": 0, "group": "root", "item": "provision-accounts.sh", "mode": "0777", "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /opt/OPEN_Admin/OPENTLC-OCP3/provision-accounts.sh", "owner": "root", "path": "/home/opentlc-mgr/bin/provision-accounts.sh", "secontext": "unconfined_u:object_r:home_bin_t:s0", "size": 54, "src": "/opt/OPEN_Admin/OPENTLC-OCP3/provision-accounts.sh", "state": "link", "uid": 0}
```
which can be **blocking** for users. It's a bug introduced by PR #52 
This PR fixes that.

PR #71, opened 2017-08-23
Fix report and conditional

fridim:
-  gather facts before generating report
  - This should fix the 'ansible_process' not defined error.
- fix conditional '== true'
  - even when running the playbook with -e ocp_report=true, the report generation was always skipped. This commit fix the syntax and those conditional. Also fix other similar conditional.
- add the guid in the report generated filename


fridim:
merge to test on admin.na

PR #72, opened 2017-08-23
wrapper.sh: pass additional args to the ansible-playbook command

fridim:


fridim:
merge to test on admin.na

PR #80, opened 2017-09-05
use variables for root fs size in ec2 templates + add one for bastion as well

fridim:
Use variables to be able to simply change size of the device attached for rootfs.
* ocp-workshop
* ocp-ha-lab


fridim:
- use var `docker_device` in ec2 templates instead of hardcoded `/dev/xvdb`
- add `nfs_size` var to allow adjusting the device size for NFS on support instance

PR #81, opened 2017-09-07
Ansible Tower Workshop via Terraform

dischord01:
I got this to work and added a Terraform cloud_provider with instructions on how to use in my README in `configs/ansible-tower-terraform`. Well probably need to chat on exactly how you want this merged but it works well creating and tearing down infrastructure and instances.


sborenst:
@dischord01 Ping me this week/next to see if we can make this more silky smooth

PR #91, opened 2017-10-13
add scaleup action to end-user's wrapper.sh

fridim:
now used in our process for long-running clusters (DEV1, NA1, RHPDS...)


fridim:
closed by 
ae44e5a

PR #93, opened 2017-10-16
add tag 'owner' to CloudFormation stacks

fridim:
Do the same as for instances.

Tags applied to stack are applied to all the resources of the stack:

Revert "add 'owner' tag on ec2 instances (cloudformation templates)"
This reverts commit d9abab9.
But keeps changes made on README.org(s)


fridim:
tested from laptop with email set and unset, works.

fridim:
ec2_infrastructure_deployment.yml  now has this change. I close this PR.

PR #95, opened 2017-10-27
Ec22 new provider

sborenst:


sborenst:
Changed ec2.py provider to use add_host

PR #97, opened 2017-10-31
Replaces nexus 2 role with generic nexus role and RH maven repos

siamaksade:
The new Nexus role accepts the `nexus_version` variable to determine to install Nexus 2 or 3. In both cases, the Red Hat maven repos are configured by default on the official Nexus image downloaded from Docker Hub.

Nexus 2 is used for `ocp-demo-lab` and `opentlc-shared` like before but `ocp-workshop` is changed to Nexus 3 now.


siamaksade:
The conflict is just Git messing up. It's an easy merge:
https://www.diffchecker.com/a80LwpGK

ansible/roles/nexus-container/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: Download Nexus template to master
+    get_url:
+      url: https://raw.githubusercontent.com/OpenShiftDemos/nexus/master/nexus{{nexus_version}}-persistent-template.yaml
+      dest: /root/nexus-template.yaml
+      mode: 0440
+      
+  - name: Check if Nexus was already provisioned
+    command: "oc get service nexus -n {{desired_project}}"
+    register: install_nexus
+    ignore_errors: true
```
fridim: can you add `changed_when: false`, as this is read-only ?

ansible/roles/nexus-container/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+### We should add a check if a pv is already created (nexus requests a 10Gi pv)
+  - name: Instantiate Nexus from template
+    command: "oc process -f /root/nexus-template.yaml --param=VOLUME_CAPACITY=10Gi --param=MAX_MEMORY=2Gi -n {{desired_project}} | oc create -f - -n {{desired_project}}"
```
fridim: just a comment for later: wouldn't be better to have params as variables defined in playbook vars (with defaults in `roles/nexus/defaults/` set to those values) ?

fridim:
```
command: "oc process [...] | oc create [...]"
```

This won't work. Command is for a single command and you are using a pipe. You need to change it to shell.

ansible/roles/nexus-container/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  # looks like we need a better check - it seems we're ready up to several
+  # seconds before the router finds out about us, so we might want another
+  # http check to make sure nexus is responding
+  - name: Wait for Nexus to be running
+    command: "oc get dc/nexus -o yaml -n {{desired_project}}"
+    register: result
+    until: '"availableReplicas: 1" in result.stdout'
+    retries: 5
+    delay: 60
```
fridim: same as previously: can you add `changed_when: false` ?

ansible/roles/nexus-container/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+## Sonatype Nexus Install
+
+##
+  - name: Download Nexus template to master
+    get_url:
+      url: https://raw.githubusercontent.com/OpenShiftDemos/nexus/master/nexus{{nexus_version}}-persistent-template.yaml
```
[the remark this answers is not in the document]
siamaksade: Sure

ansible/roles/nexus-container/tasks/main.yml
siamaksade: 👍

ansible/roles/nexus-container/tasks/main.yml
siamaksade: Possibly. I keep them as is for the PR but definitely something to consider for future.

siamaksade:
Updated

PR #107, opened 2017-12-13
Adding support for tower post deploy

bennojoy:
This commit adds the support for configuring tower post install with projects and job templates so that it is ready for Demo

Just add extra var "deploy_tower_demo=true" to the deploy script to enable configuring the tower.


ansible/roles/tower_demo/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+    src: "{{ ANSIBLE_REPO_PATH }}/files/tower_cli.j2"
+    dest: "~/.tower_cli.cfg"
+    mode: 0640
+
+- name: copy the license file
+  copy:
+   src: "{{ ANSIBLE_REPO_PATH }}/files/tower_license"
+   dest: "/tmp/tower_license"
+
+- name: copy the private key
```
fridim: careful with private keys.
We usually allow copying private key depending of variables (`use_own_key`)

see https://github.com/sborenst/ansible_agnostic_deployer/blob/development/ansible/roles/set_env_authorized_key/tasks/main.yml

and 

https://github.com/sborenst/ansible_agnostic_deployer/blob/development/ansible/roles/bastion/tasks/main.yml

ansible/roles/tower_demo/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+  copy:
+    src: "{{ ANSIBLE_REPO_PATH }}/workdir/{{ guid + 'key' }}"
+    dest: /tmp/id_rsa
```
fridim: in `/tmp/` ?  does it mean it won't work after a reboot ?

ansible/roles/tower_demo/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+    mode: 0640
+  
+- name: generate ansible hosts file
+  template:
+    src: "{{ ANSIBLE_REPO_PATH }}/configs/{{ env_type }}/files/hosts_template.j2"
+    dest: "/etc/ansible/hosts"
+
+- name: Add the license of tower
+  command: |
+     tower-cli setting modify LICENSE '{"eula_accepted" : "true", "company_name": "Red Hat Internal", "contact_email": "ber@redhat.com", "contact_name": "Jon Bersuder",  "instance_count": 16, "license_date": 1514782800, "license_key": "2e2995c0e36fa1846c34db031138d44e15dcf5b04c414c462d159f2648f579aa", "license_type": "enterprise", "subscription_name": "Ansible Tower by Red Hat, Standard (16 Managed Nodes)",  "trial": false }' -u admin -p "{{ tower_admin_password }}"
```
fridim: can you split this line for readability ? i think you can do multilines using `\`

ansible/roles/tower_demo/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+
+
+- name: Add tower scm credential
+  shell: |
+        tower-cli credential create --organization="{{ org_name }}" --kind=scm --username=test --password=test --name="{{ scm_cred_name }}"
+
+- name: Create Tower project
+  shell: |
+        tower-cli project create --organization="{{ org_name }}" --scm-type=git --scm-branch="{{ scm_branch }}" --scm-credential="{{ scm_cred_name }}" --scm-url="{{ scm_url }}" --scm-update-on-launch=True -n "{{ proj_name }}"
```
fridim: same, multiline would be better

ansible/roles/tower_demo/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+
+- name: create the user ssh credential
+  shell: |
+       tower-cli credential create --name="{{ ssh_cred_name }}" --username="{{ user_cred_name }}" --organization="{{ org_name }}" --kind=ssh --ssh-key-data="{{ user_cred_path }}" --become-method=sudo
+
+- name: Create the job template
+  shell:  |
+        tower-cli job_template create --name 3tier_app_demo  --project="{{ proj_name }}" --inventory="{{ inv_name }}" --job-type=run --playbook='{{ playbook_name }}' --machine-credential="{{ ssh_cred_name }}"
+
+#- name: Add tower scm credentail
```
fridim: Are those commented lines the ansible version of the previous shell tasks ?

If yes, i think we should prefer the ansible-like task

ansible/roles/tower_demo/tasks/main.yml
bennojoy: this is just temporary to upload to tower

ansible/roles/tower_demo/tasks/main.yml
fridim: "license_key": "2e2995xxxxxxxxx59f2648f579aa"

Shouldn't this be a variable  ? what is this license key ?

PR #111, opened 2018-01-17
Added Let's Encrypt Wildcard Support

wkulhanek:
Added Let's Encrypt Wildcard Support for cop-workshop. This will automatically request certificates for Masters and Routers and deploy them upon installation. The certificates are good for 90 days after which time they need to be renewed. But for our purposes that's most likely fine.

The intermediate certificate (https://letsencrypt.org/certs/fakeleintermediatex1.pem) needs to be imported into the browsers to make this work.

@sborenst @fridim Please review and approve.


fridim:
awesome! thanks @wkulhanek

PR #121, opened 2018-02-14
Updated openshift-demos verify and image build playbooks

siamaksade:


siamaksade:
Will create a new PR from development branch

PR #124, opened 2018-02-15
Verify deployments after demos are deployed

siamaksade:


ansible/configs/openshift-demos/msa-cicd-eap-full.yml
```diff
@@ -61,3 +61,58 @@
         gogs_hostname: "gogs-{{ project_cicd }}.{{ apps_hostname_suffix }}"
         prune_deployments_selector_prod: ""
         prune_deployments_selector_stage: "comp-required!=true,app!=inventory"
+
+    # verify deployments in ci/cd project
+    - name: wait for deployments in project {{ project_cicd }} to complete (succeed or fail)
+      shell: "oc get pods -n {{ project_cicd }} | grep '\\-deploy' | grep 'Running'"
```
fridim: I would rather take advantage of json output here. `oc` can output json. and then use jinja2 filters like `from_json` and `json_query` instead of grep

```
command: "oc get pods -n {{ project_cicd }} -o json"```

File: ansible/configs/openshift-demos/msa-cicd-eap-full.yml [hunk 2]
```diff
@@ -61,3 +61,58 @@
...
+      until: deployment_running|failed
+      retries: 20
+      delay: 30
+      ignore_errors: true
+
+    - include_tasks: verify_tasks.yml project_name={{ project_cicd }}
+
+    # verify deployments in prod project
+    - name: wait for deployments in project {{ project_prod }} to complete (succeed or fail)
+      shell: "oc get pods -n {{ project_prod }} | grep '\\-deploy' | grep 'Running'"
```
fridim: same as previous comment

ansible/configs/openshift-demos/msa-cicd-eap-full.yml
```diff
@@ -61,3 +61,58 @@
         gogs_hostname: "gogs-{{ project_cicd }}.{{ apps_hostname_suffix }}"
         prune_deployments_selector_prod: ""
         prune_deployments_selector_stage: "comp-required!=true,app!=inventory"
+
+    # verify deployments in ci/cd project
+    - name: wait for deployments in project {{ project_cicd }} to complete (succeed or fail)
+      shell: "oc get pods -n {{ project_cicd }} | grep '\\-deploy' | grep 'Running'"
+      register: deployment_running
+      until: deployment_running|failed
```
fridim: This would then be something like

```
until: deployment_running|from_json|json_query(...) )  is empty
```

determining '(...)' is the hard part. I think `.spec.name == 'deployment'  && .status.phase == 'Running'` for '-deploy'  pods should do it.

PR #134, opened 2018-03-09
Updated workload.yml - Fuse Ignite

honghuac:
A version of the workload.yml for Fuse Ignite playbook is now available.


sborenst:
@jbride Would you like to approve this PR? (looks good to me)

PR #135, opened 2018-03-21
Client VM Config and Role

wkulhanek:
@fridim @sborenst @NonyNo3lle 
Added a config to create a Client VM. Also added Role to install and configure the Client VM with Docker and all packages.

Tested locally. Needs testing from Cloud Forms with OPENTLC IPA Integration.

Create a rc file from the example in the README of the config. Then use the wrapper script to create it.


ansible/configs/ocp-clientvm/destroy_env.yml
```diff
@@ -0,0 +1,31 @@
...
+      until: cloudformation_result|succeeded
+      retries: {{ cloudformation_retries }}
+      delay: 60
+      ignore_errors: yes
+
+    - name: report Cloudformation error
+      fail:
+        msg: "FAIL {{ project_tag }} Destroy Cloudformation"
+      when: not cloudformation_result|succeeded
+      tags: [ destroying, destroy_cf_deployment ]
```
fridim: here you could just use the common ec2 destroy playbook:

```---
- name: Import default CloudFormation (aws) destroy playbook
  import_playbook: "{{ANSIBLE_REPO_PATH}}/cloud_providers/{{cloud_provider}}_destroy_env.yml"
```

ansible/configs/ocp-clientvm/files/cloud_providers/ec2_cloud_template.j2
```diff
@@ -0,0 +1,247 @@
...
+  HostTCPPorts:
+    Type: "AWS::EC2::SecurityGroupIngress"
+    Properties:
+      GroupId:
+        Fn::GetAtt:
+          - HostSG
+          - GroupId
+      IpProtocol: tcp
+      FromPort: 0
+      ToPort: 65535
```
fridim: we only need port 22, no ?

ansible/configs/ocp-clientvm/post_software.yml
```diff
@@ -0,0 +1,65 @@
...
+      ec2_remote_facts:
+        aws_access_key: "{{ aws_access_key_id }}"
+        aws_secret_key: "{{ aws_secret_access_key }}"
+        region: "{{ aws_region }}"
+      when: ocp_report
+    - name: Generate report
+      template:
+        src: "{{ ANSIBLE_REPO_PATH }}/configs/{{ env_type }}/files/ocp_report.adoc.j2"
+        dest: "{{ ANSIBLE_REPO_PATH }}/workdir/ocp_report_{{ env_type }}-{{ guid }}.adoc"
+      when: ocp_report
```
fridim: do you really need the ocp report for this single-instance config ?

ansible/configs/ocp-clientvm/files/cloud_providers/ec2_cloud_template.j2
wkulhanek: Right. I'll change that.

Do we need the internal DNS zone for just one VM? I don't think so but am not sure what parts of the Cloudformation Template I can remove...

ansible/configs/ocp-clientvm/destroy_env.yml
wkulhanek: Didn't know that existed. That one doesn't work anyway... doesn't work for ocp-workshop either. I'll change it here. If that works maybe we should change ocp-workshop as well.

ansible/configs/ocp-clientvm/post_software.yml
wkulhanek: Probably not. Don't know what it does - so I can take the whole section out.

wkulhanek:
@fridim all updates made. Even made sure that Mosh was there. :-) So only question remaining for me is if we need the internal DNS zone.

PR #136, opened 2018-03-22
Initial import of the fsi-client-onboarding demo.

DuncanDoyle:
@sborenst I've tested this against my local OpenShift instance using Ansible and it seems to work fine. Can you have a look at the resource quotas and limits I've set? Let me know if you need me to do more fine-tuning on those.


sborenst:
@DuncanDoyle Merged, Please test this with 'Dev37', let me or @NonyNo3lle know if you need help with that.

PR #139, opened 2018-04-06
Added support for m5.4xlarge instances and CRI-O 

wkulhanek:
Added support for m5.4xlarge instances (**need to be same** on masters, infranodes and nodes).

Also added support for CRI-O (crio_enabled=false/true). The Garbage Collection Daemon Set still crashes though preventing cleanup of build docker containers... (missing fix for bug https://bugzilla.redhat.com/show_bug.cgi?id=1564271). Somehow I needed to pre-install atomic on all nodes and pre-pull the cri-o image - that weirdly failed during running of the prerequisites playbook. With the pre-install/pull it works.

### DIFF
--- a/ansible/configs/ans-tower-lab/env_vars.yml
+++ b/ansible/configs/ans-tower-lab/env_vars.yml
@@ -104,6 +104,13 @@ security_groups:
         rule_type: Ingress
   - name: TowerSG
     rules:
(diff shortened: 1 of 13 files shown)


wkulhanek:
@sborenst @fridim Please review and comment. :-)

PR #140, opened 2018-04-12
Updated HA Workshop for OCP 3.9

wkulhanek:
@newgoliath @sborenst @fridim Please review. Updated HA Workshop for 3.9. I also removed all logic for releases before 3.7. We don't run any classes on earlier versions any more.

### DIFF
--- a/ansible/configs/ocp-ha-lab/env_vars.yml
+++ b/ansible/configs/ocp-ha-lab/env_vars.yml
@@ -38,7 +38,7 @@ config_nfs_uservols: "true"
 user_vols: 200
 user_vols_size: 4Gi
 master_api_port: 443
-osrelease: 3.7.9
+osrelease: 3.9.14
 openshift_master_overwrite_named_certificates: true
 deploy_openshift: true
 deploy_openshift_post: true
@@ -70,7 +70,7 @@ software_to_deploy: "openshift"
 #### OCP IMPLEMENATATION LAB
 ################################################################################
 
-repo_version: '3.7'
+repo_version: '3.9'
 cloudapps_dns: '*.apps.{{subdomain_base}}.'
 master_public_dns: "loadbalancer.{{subdomain_base}}."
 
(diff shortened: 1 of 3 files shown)


ansible/configs/ocp-ha-lab/files/repos_template.j2
```diff
@@ -34,4 +34,11 @@ baseurl={{own_repo_path}}/rhel-7-fast-datapath-rpms
 enabled=1
 gpgcheck=0
 
+[rhel-7-server-ansible-2.4-rpms]
```
fridim: do we have this repo on admin host ? satellite?

ansible/configs/ocp-ha-lab/env_vars.yml
```diff
@@ -169,7 +175,7 @@ ansible_ssh_private_key_file: ~/.ssh/{{key_name}}.pem
 
 project_tag: "{{ env_type }}-{{ guid }}"
 
-docker_version: "1.12.6"
+docker_version: "1.13.1"
```
fridim: don't you want to use your magic line here?

    docker_version: "{{ '1.12.6' if repo_version | version_compare('3.9', '<')  else '1.13.1' }}"

ansible/configs/ocp-ha-lab/files/hosts_template.3.9.14.j2
```diff
@@ -210,16 +159,13 @@ openshift_logging_curator_nodeselector={"env":"infra"}
 # Configure additional projects
 openshift_additional_projects={'openshift-template-service-broker': {'default_node_selector': ''}}
 
-
 ###########################################################################
 ### OpenShift Router and Registry Vars
 ###########################################################################
 
-openshift_hosted_router_selector='env=infra'
```
fridim: Do we want to remove this line ?

ansible/configs/ocp-ha-lab/env_vars.yml
sborenst: We don't really need to keep ocp-ha-lab backward compatible.

ansible/configs/ocp-ha-lab/files/hosts_template.3.9.14.j2
sborenst: let's keep it for now, we can remove it when we come up with a better line. 
Eric S suggested we change it to "type" rather than "env"

wkulhanek:
Actually hosted-infra-selector is now for,router, registry, tsb and asb (last one doesn’t work yet). So,we can remove individual selectors for those components.

newgoliath:
I'm confused - what's the difference between ocp-ha-lab and this "HA Workshop"?

PR #141, opened 2018-04-17
Workload for bu added

jorgemoralespou:
Can you @sborenst and @fridim review this PR and make comments so I can adjust?

These are the still pending issuses/caveats:

- In main.yml needs become:true as the user that has oc configured is root.

- In workload.yml when openshift_common_facts need to specify path to .kube/config in /root

- Users are defined with user%02d pattern

- Input variables needed:
  - admin_project
  - user_count


ansible/roles/ocp-workload-starter-workshop/NOTES.txt
```diff
@@ -0,0 +1,9 @@
+- In main.yml needs become:true as the user that has oc configured is root.
```
fridim: On shared cluster we set up an admin user `opentlc-mgr` that has system:admin user configured for `oc` commands.

ansible/roles/ocp-workload-starter-workshop/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+user_count_ini: 1
+user_count_end: 50
+user_format: user%02d
+
+##
+# VALUES SPECIFIC TO THIS WORKLOAD
+# Uncomment and set a value to take effect. Else, will use the defaults
+#
+user_gogs_admin: "gogsadmin"
+user_gogs_password: "r3dh4t1!"
```
fridim: please have this in env_secret_vars.yml (not versionned) instead and an example using `-e user_gogs_password=...` in readme

ansible/roles/ocp-workload-starter-workshop/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Running Pre Workload Tasks
+  include: ./pre_workload.yml
```
fridim: `include` is deprecated since ansible 2.4 IIRC.

can you use one of those instead of `include`:
- import_tasks  (static)
- include_tasks (runtime)

prefer static if possible

ansible/roles/ocp-workload-starter-workshop/tasks/per_user.yml
```diff
@@ -0,0 +1,77 @@
...
+- set_fact:
+    my_user: "{{ item }}"
+
+- name: per_user {{my_user}} Tasks Started
+  debug:
+    msg: "per_user {{my_user}} Tasks - Started"
+
+# Create user project and make him admin
+- name: Create workshop user's project for user "{{ my_user }}"
+  shell: oc adm new-project "{{ my_user }}"
```
fridim: you don't need the shell module most of the time, command is enough:

```
command: oc adm new-project "{{ my_user }}"
```

Use shell only when you need Shell features: pipes, input/output redirects, subshell, or multiple commands, etc.

ansible/roles/ocp-workload-starter-workshop/tasks/per_user.yml
```diff
@@ -0,0 +1,77 @@
...
+        --hard limits.cpu="{{quota_limits_cpu}}"  \
+        --hard requests.memory="{{quota_requests_memory}}" \
+        --hard limits.memory="{{quota_limits_memory}}" \
+        --hard configmaps="{{quota_configmaps}}" \
+        --hard pods="{{quota_pods}}" \
+        --hard persistentvolumeclaims="{{quota_persistentvolumeclaims}}"  \
+        --hard services="{{quota_services}}" \
+        --hard secrets="{{quota_secrets}}" \
+        --hard requests.storage="{{quota_requests_storage}}"
+  ignore_errors: true
```
fridim: Just checking: did you mean each one of the `ignore_errors` in the playbooks ?

don't we want to know if the command for quotas failed ?

ansible/roles/ocp-workload-starter-workshop/tasks/pre_workload.yml
```diff
@@ -0,0 +1,39 @@
...
+# - name: Add user to developer group (allowed to create projects)
+#   shell: "oadm groups add-users {{item}} {{ocp_username}}"
+#   register: groupadd_register
+#   with_items: "{{ocp_user_groups}}"
+#   when: ocp_username is defined and ocp_user_groups is defined
+#
+# - name: test that command worked
+#   debug:
+#     var: groupadd_register
+#     verbosity: 2
```
fridim: if those 2 tasks are commented maybe they can be removed ?

ansible/roles/ocp-workload-starter-workshop/defaults/main.yml
jorgemoralespou: where would env_secret_vars.yml go? Is this file picked automatically like defaults/main.yml?

ansible/roles/ocp-workload-starter-workshop/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
+---
+# PROVIDED BY THE INFRA
+# Uncomment and set a value to take effect. Else, will use the defaults
+#
+# ocp_apps_domain:
+admin_project: 
+user_count_ini: 1
+user_count_end: 50
+user_format: user%02d
```
fridim: why do you want to use `user%02d` ? why not be backward compatible with what instructors are used to (user1, user2, ..)

why not `user%d` ?

jorgemoralespou:
@fridim  Can you do another review?

PR #153, opened 2018-06-26
Add Industry 4.0 IoT Demo

ccustine:


ansible/roles/ocp-workload-iot-demo/files/certs/readme.md
```diff
@@ -0,0 +1,3 @@
+# Warning!
+The certs in this directory are self generated placeholders for the initial installation. It is expected that 
+these will be replaced soon after installation by updating the ec_cert secret in the Openshift project.
```
fridim: it seems always strange to have cert keys versioned in the code. Isn't it possible to generate cert and key before the installation instead ?

ansible/roles/ocp-workload-iot-demo/files/redhat-iot-rhpds-secret.yml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: redhat-iot-rhpds-pull-secret
+data:
+  .dockerconfigjson: ewogICJhdXRocyI6IHsKICAgICJxdWF5LmlvIjogewogICAgICAiYXV0aCI6ICJjbVZrYUdGMExXbHZkQ3R5YUhCa2N6cElUVWREV0ZsWFMwUTVNVXRGTjFwSVFUQlRVa0pVTkV0V1MweFhUREJTVVROTVJFaFFNRlpDU0ZSUlFVTktRbFJQU1ZJek1WZEdPVGd3TmsweFUweEUiLAogICAgICAiZW1haWwiOiAiIgogICAgfQogIH0KfQ==
```
fridim: same question here, can we generate the secret ?

ansible/roles/ocp-workload-iot-demo/files/certs/readme.md
ccustine: @fridim I factored this out and borrowed some code from another similar role. everything under /certs has been removed and is generated now.

PR #154, opened 2018-07-02
Workshops menu

jorgemoralespou:
This role adds a workshops menu in the Service Catalog.

### DIFF
--- a/ansible/configs/ocp-workshop/post_software.yml
+++ b/ansible/configs/ocp-workshop/post_software.yml
@@ -383,6 +383,19 @@
     - env-specific
     - install_nexus
 
+- name: Customize Service Catalog UI for workshops
+  hosts: masters
+  run_once: true
+  gather_facts: False
+  become: yes
+  vars_files:
+    - "{{ ANSIBLE_REPO_PATH }}/configs/{{ env_type }}/env_vars.yml"
+  roles:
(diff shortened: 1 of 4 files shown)


jorgemoralespou:
cc @siamaksade  @fridim 
Not sure if a generic role is the best way as this is very specific to the ocp_workshop config

siamaksade:
Agree that it's better if we could add it only to OCP Workshop. For example when RHPDS shared cluster is deployed, this role shouldn't be included

fridim:
@jorgemoralespou , I can't pull from your repo. Did you delete the branch ?

Regarding the role, we are already using one to configure the broker: openshift-ansible-broker. For the moment this role only adds openshiftapb docker hub in the config.

Maybe we can add your code into the openshift-ansible-broker role, as it's related to openshiftapb too. Wdyt?

FYI, it's now not installed by default. `install_openshiftapb: false`: https://github.com/sborenst/ansible_agnostic_deployer/commit/5f069b922cd9c3bdaa1b3db90e9914283769bc02

sborenst:
We need to add:
"enable_workshops_catalog: false" in the env_vars.yml file as the default
and add a "when: enable_workshops_catalog | bool" to the task you want to run
Role should be called "infra-ocp-enable-custom-catalog"

jorgemoralespou:
Close in favor of #178

PR #158, opened 2018-07-19
Fix role import

jorgemoralespou:
@fridim In the AAD you import the role in this way


fridim:
i miss clicked.. wanted to merge this PR.

PR #166, opened 2018-08-03
Single node gpu

thoraxe:
Adds a single-node GPU config to support doing OpenShift demos using NVIDIA GPU-enabled EC2 instances.


thoraxe:
Wrong branch.

PR #170, opened 2018-08-20
Fix cicd server

ericzolf:
It was agreed between @sabre1041 and myself that the CI/CD server would be a separate, so here the necessary changes (I hope); in compensation, the bastion host has been made smaller. Additionally I've started to add a high-level description of the environment, which we should complete as we progress.


NonyNo3lle:
Merged.

PR #172, opened 2018-08-20
https://github.com/redhat-gpe/bxms_process_automation_foundations/iss…

diego-torres:
…ues/28

https://github.com/redhat-gpe/bxms_process_automation_foundations/issues/28
Default for user cloudforms provision, when using for workshop use the -e MAVEN_REPO_URL option to override during playbook execution. Value for workshop cluster: http://nexus3.default.svc.cluster.local:8081/repository/maven-public/


jbride:
thanks

PR #188, opened 2018-08-28
WiP: Use Software Collections to install pip and Molecule

drhelius:
Ansible CI/CD lab:
This fixes an issue installing pip on RHEL 7 by using rhel-server-rhscl-7-rpms repo


ericzolf:
If it's required, I'm fine to merge it but SCL is really my last resort. Would it work and wouldn't it be easier to use Python 3 instead of Python 2 from SCL?

ericzolf:
After discussion with @drhelius the PR might not be necessary, please don't merge before there is a comment from him asking explicitly for merging.

drhelius:
Ignore this PR.
It looks like it's working nice as it is now.

PR #206, opened 2018-09-05
Updated ocp-client-vm workload - added script for AMT and EAP installation - AAD Development Branch

honghuac:
Instructions for App Migration Toolkit and JBoss EAP installation added to ocp-client-vm


wkulhanek:
@honghuac @sborenst This can't be in the master client vm. This would make it to every single provisioned client vm world wide.
This should be a separate role that gets included from the Client VM when a specific variable is set.

PR #249, opened 2018-10-03
Enable the use of Ansible Galaxy roles

fridim:
Allow the use of Galaxy Roles in a config. If a 'requirements.yml' file exist
then automatically install the roles in the configs/{{env_type}}/roles directory.

Create HowTo doc.


ericzolf:
I had started to use a similar approach, but the issue was lack of version pinning (I saw your warning) and also the additional time it requires. Another issue you might get is that, during development, the roles will be deployed and get in your way git-wise, so there should be a rule in `.gitignore` to ignore those roles. It's not so easy because you might have roles you have patched and don't want to download dynamically, those shouldn't be ignored.

Few ideas (none perfect):

1. document this and leave it to the config-owner to add the correct .gitignore file in their config-directory.
2. add in the `ansible/configs` a `.gitignore` that ignores `roles`, document it, and leave it to the config-owner to add the correct .gitignore file in their config-directory to _not_ ignore their static roles (doable with an exclamation mark in front of the line).
3. as you can rename roles while downloading them, you could add a naming convention allowing to ignore such dynamic roles easily without touching others.
4. or introduce the convention that all static roles are to be in `ansible/roles` and dynamic roles under `ansible/configs/{{ config-name }}/roles` and add in the `ansible/configs` a `.gitignore` that ignores `roles`.

All in all, many possibilities, rather equivalent, just needs to be properly documented. Point 4 is properly the easiest one (exceptions can be made using approach Nr. 2).

tonykay:
I think the easiest, and arguably Best Practice, is to allow a requirements.yml which **uses** verion pinning:

```
- src: nginxinc.nginx
  version: 0.8.0
```

It also provides good encapulation. Although its not used I've started putting a requirements.txt into configs like the linklight wrapper so I know what libs are needed. Then the day we start to use virtualenvs more we have that.

PR #264, opened 2018-10-17
ocp-client-vm :  default variables for maven and non_root_user

jbride:
A few changes to ocp-client-vm based on running it in the following environments:

1)  RHEL7 vm managed in local libvirtd environment
2)  RHEL7 vm managed in remote Azure environment


ansible/roles/ocp-client-vm/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+non_root_os_user: "ec2-user"
```
fridim: there is another variable `remote_user` and `ansible_ssh_user` don't you want to use that instead ?

ansible/roles/ocp-client-vm/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
...
+maven_version: "3.5.4"
+docker_device: /dev/xvdb
```
wkulhanek: @jbride  Where is this docker_device being used? I don't see this anywhere in your changes.

ansible/roles/ocp-client-vm/tasks/packages.yml
```diff
@@ -16,36 +16,49 @@
...
-    url: https://www-us.apache.org/dist/maven/maven-3/3.5.4/binaries/apache-maven-3.5.4-bin.tar.gz
+    url: https://www-us.apache.org/dist/maven/maven-3/{{maven_version}}/binaries/apache-maven-{{maven_version}}-bin.tar.gz
     dest: /root/apache-maven.tar.gz
   register: r_geturl
   retries: 5
   delay: 20
   until: r_geturl is succeeded
   tags:
   - install_openshift_client_vm_packages
+  when: 'existing_mvn_version.stat.exists == False'
```
wkulhanek: @jbride  I'd prefer a block around the "Maven Install". That way the when condition is only evaluated once. See the section "Install everything for student user" as an example.

ansible/roles/ocp-client-vm/defaults/main.yml
wkulhanek: @jbride @fridim Agreed. Any reason we couldn't use `remote_user`

fridim:
@jbride any update on this PR ?

PR #265, opened 2018-10-18
Adding user info to Integreatly workload

pmccarthy:
**Summary**
Adding user info to Integreatly workload
Updating release tag


pmccarthy:
@NonyNo3lle @fridim Hey guys, mind taking a quick look please?

ansible/roles/ocp-workload-integreatly/tasks/workload.yml
```diff
@@ -7,8 +7,29 @@
...
+  register: openshift_master_config
+
+- name: Retrieve Web App URL
+  shell: oc get route/tutorial-web-app -o template --template \{\{.spec.host\}\} -n "{{ webapp_namespace }}"
+  register: webapp_route
+
+- set_fact:
+    openshift_master_url: "{{ (openshift_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
+
+- name: Print Overview
```
fridim: what is this used for ?

Can we not print password please ?

ansible/roles/ocp-workload-integreatly/defaults/main.yml
```diff
@@ -4,4 +4,10 @@ become_override: True
...
 inventory_hosts_file: inventories/hosts
-release_tag: release-0.0.1
\ No newline at end of file
+release_tag: release-1.0.0
+webapp_namespace: webapp
+openshift_master_config_path: /etc/origin/master/master-config.yaml
+admin_username: admin@example.com
+admin_password: Password1
+evals_username: evals@example.com
+evals_password: Password1
```
fridim: can we avoid to hardcode the password ?
You can set it by default to another password if there is any, like `student_password` ?
Or leave it undefined, and we have to pass it to the playbook run

ansible/roles/ocp-workload-integreatly/tasks/workload.yml
fridim: the logs of the deployment will not be automatically accessible, so i'm not sure, what the `print Overview` task is for here.

pmccarthy: Ok, maybe I misunderstood. I thought that the user.info messages would populate a section of the confirmation email that gets sent, hence the print Overview.

ansible/roles/ocp-workload-integreatly/defaults/main.yml
pmccarthy: Happy to leave out passwords for now, these credentials are referenced in our public README so are not sensitive. All the same I agree that we might be better off not specifying them.

ansible/roles/ocp-workload-integreatly/tasks/workload.yml
pmccarthy: In terms of getting the `user.info` messages into the email template, is this something that needs to be configured afterwards?

fridim: OK my bad then.
I'll merge your PR as is, i was not aware you agreed on this. That's cool.

pmccarthy: Thanks. It was @NonyNo3lle who had pointed me to the user.info examples after I asked to have a custom section in the confirmation email for Integreatly

pmccarthy:
@fridim Just pushed up original PR

PR #266, opened 2018-10-23
Setting TLS flag for Integreatly workload

pmccarthy:
**Summary**
Now that the Integreatly catalog item has LetsEncrypt enabled by default, we need to set a flag as part of the Integreatly install.

Also bumped Integreatly release tag


pmccarthy:
@fridim Mind taking a look when you get a chance please? Thanks

PR #267, opened 2018-10-25
Reboot vms

wkulhanek:
@fridim @sborenst @newgoliath Added logic to reboot VMs after the "yum -y update" step.

Currently this works for Ansible 2.6 (which is what have on our Admin hosts). When upgrading to Ansible 2.7 it will throw an error to remove the 2.6 logic and use the (commented out) 2.7 logic as well.

I tested the 2.7 logic on my Mac and it works fine. But on 2.6 on Admin Dev it throws an error that it finds an empty task (since it doesn't understand 'reboot:') even when there is a "when" clause. So I had to comment it out.


ansible/configs/ocp-workshop/env_vars.yml
```diff
@@ -406,7 +406,7 @@ instances:
 
   - name: "node"
     count: "{{node_instance_count}}"
-    public_dns: false
```
fridim: Is this related to this "reboot the vms" topic ? It doesn't look like it.

ansible/roles/common/tasks/main.yml
```diff
@@ -10,6 +10,39 @@
...
+  - update_packages|bool
+  - ansible_version.full is version_compare("2.7", "<")
+
+- name: Wait for VMs to come back up
+  wait_for_connection:
+    delay: 30
+    timeout: 300
+  when:
+  - update_packages|bool
+  - ansible_version.full is version_compare("2.7", "<")
```
fridim:
```#- name: Reboot all VMs after updating to the latest release
#  reboot:
#    connect_timeout: 300
#    msg: "Ansible updates installing. Rebooting now."
#    pre_reboot_delay: 60
#    post_reboot_delay: 10
#  when:
#  - update_packages|bool
#  - ansible_version.full is version_compare("2.7", ">=")

- name: Check if right module is used
  fail:
    msg: Please use the reboot module in Ansible > 2.7
  when:
  - ansible_version.full is version_compare("2.7", ">=")
```

You commented the ansible-2.7 region here. If i understand correctly, if a user has ansible 2.7 and is using the `common` role, his playbook will fail. Then he/she will have to uncomment the 2.7 region of the code, and comment the 2.6 region of the code, right ?

I would prefer if the common module could work for both 2.6 and 2.7, without manual steps for the user.

I think this is achievable by using dynamic includes so ansible 2.6 doesn't throw a syntax error if it sees `reboot` module somewhere. For example, have you tried:

```
- name: Reboot all VMs after updating to the latest release (ansible 2.7)
  include_tasks: reboot_27.yml
  when: ansible_version.full is version_compare("2.7", ">=")

- name: Reboot all VMs after updating to the latest release (ansible 2.6)
  include_tasks: reboot_26.yml
  when: ansible_version.full is version_compare("2.7", "<")

File: ansible/roles/common/tasks/packages.yml [hunk 3]
```diff
@@ -3,9 +3,8 @@
 ######################### Install Basic Packages
 - name: install basic packages
   yum:
-    name: "{{ item }}"
     state: present
-  with_items: "{{common_packages}}"
+    name: "{{common_packages}}"
```
fridim: this is good, we need to change it everywhere, `with_items` + `yum` being deprecated.
Does this work with all ansible versions ?
No other places ? Can we safely search and replace everywhere ?

( Not directly related to this "reboot vms" topic though. )

ansible/configs/ocp-workshop/env_vars.yml
wkulhanek: Yes. It needs an EIP because the IPs change and then AAD doesn't know who to talk to any more. Now what we REALLY should be doing according to @sborenst is to run a dynamic inventory after the reboot. Sha doesn't want to do the reboot - I however strongly feel we should do it. If you can tell me how to update the inventory after the reboot I think he'd be OK with it and we wouldn't need EIPs for the nodes. Of course this is tactical until we have the Packer images. So maybe we postpone until Packer.

ansible/roles/common/tasks/packages.yml
wkulhanek: I have not tried include_tasks. But that's a good idea. @fridim

sborenst:
Can we merge this? Is there something in our way @wkulhanek @fridim

wkulhanek:
@sborenst @fridim This should be good to go. I'm doing the include tasks as recommended by Guillaume (and that works) and then I'm doing a gather facts as recommended by Sha to avoid having EIPs on the Nodes. In my testing this works.

sborenst:
Issue resolved in other ways. 

In future we should reboot after a 'needs_reboot' test.

fridim:
Do we really need to reboot all the time ?

What if the image used is up to date or have minor package updates ?


@wkulhanek Can you remind us what is the issue you are trying to fix by rebooting the instances ?

Today, most of the configs work without rebooting. I understand the change in this PR wants to be pro-active, but i would prefer if it's an option we activate when we need it, ie:

```yaml
reboot_after_update: fase
```

fridim:
@wkulhanek Have you tried to update the AMI instead, does it fix the issue ?

ansible/configs/ocp-workshop/env_vars.yml
```diff
@@ -450,7 +450,7 @@ instances:
 
   - name: "support"
     count: "{{support_instance_count}}"
-    public_dns: false
+    public_dns: true
```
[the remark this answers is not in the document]
newgoliath: OCS

newgoliath:
> Do we really need to reboot all the time ?
> 
> What if the image used is up to date or have minor package updates ?
> 
> @wkulhanek Can you remind us what is the issue you are trying to fix by rebooting the instances ?
> 
> Today, most of the configs work without rebooting. I understand the change in this PR wants to be pro-active, but i would prefer if it's an option we activate when we need it, ie:
> 
> ```yaml
> reboot_after_update: fase
> ```

with `yum update` there are kernel improvements that require reboot... for OCS.

ansible/configs/ocp-workshop/env_vars.yml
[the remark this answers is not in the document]
fridim: Should we do 

```yaml
public_dns: {{ true if install_glusterfs | bool  else false }}
```

fridim:
I added a test to check if a reboot is needed before rebooting the instance

PR #277, opened 2018-11-09
Update parksmap with instructions valid for 3.10 and 3.11

jorgemoralespou:
Wayne Boxall reported on Oct 23rd that the guides were incomplete. I missed to fix this sooner. This is a patch that should fix the issue.


jorgemoralespou:
@fridim @NonyNo3lle This fixes the demo

PR #283, opened 2018-11-20
Custom images using packer

fridim:
Enable use of custom images:

The AMI is detected before the cloudformation stack is created.
There is 2 detection strategies:
- using the image name
  variable: `custom_image_filter`
  example: `custom_image_filter: RHEL 7.5 GOLD packer 1542702393`
- specific image using tags and the tuple `(env_type, version, stage)`
  variable: `custom_image_stage`
  This allows us to enable or disable an AMI easily by adding or not the stage into its tag `stages`. env_type and version have to match too.
  example: `custom_image_stage: PROD`

If no AMI is found, then it defaults to what is used in the template.

### DIFF
--- a/ansible/cloud_providers/ec2_destroy_env.yml
+++ b/ansible/cloud_providers/ec2_destroy_env.yml
@@ -34,7 +34,7 @@
       fail:
         msg: "FAIL {{ project_tag }} Destroy Cloudformation"
       when:
-        - not cloudformation_result|succeeded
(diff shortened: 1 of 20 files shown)


sborenst:
@fridim, looks good to me, feel free to merge. We should take the description you wrote in the PR (above) and add it as a "readme.adoc" file or as a comment somewhere close to this code. Can be in the env_vars or anything you see fit.

PR #292, opened 2018-11-30
Get rid of ANSIBLE_REPO_PATH and vars_files includes and Be Ansible Tower friendly

fridim:
Making AgnosticD work with Tower.

This PR contains the changes necessary to:

- Get rid of the `ANSIBLE_REPO_PATH` variable
  This variables was convenient but never necessary. The imports are now relative.
- Get rid of `vars_files` includes.
  Since 1926912a16cbcaa86a45a6858e5e8ebcb92e0d67, `env_vars.yml` files are now dynamically included at the beginning of playbooks. No more `vars_files: env_vars.yml` in every play!

This PR will be merged as separate (thus incomplete) commits to keep authorship. It was hard work.

We usually prefer small and atomic changes but this change is a huge refactoring of the code, and touches a lot of places. Some sanitizing commits are included in this PR for simplicity:

- Use role names directly instead of absolute or relative path: clean, easy. :raised_hands: 
  Example:
    ```yaml
    # before
    include_role:
      name: "{{ ANSIBLE_REPO_PATH }}/roles/openshift-node"
    
    # after
    include_role:
      name: host-ocp-node
    ```
- Cleanup of `ansible.cfg` files: they are now identical (symlink). We keep one at the top level for Tower.
  The `ansible.cfg` file is used to specify roles_path. Another solution is to use the `ANSIBLE_ROLES_PATH`  environment variable.
- Destroy roles are used when possible (DRY code)
- Add more syntax tests (travis CI), with & without setting `ANSIBLE_REPO_PATH`
- Add `scaleup.yml` playbooks to syntax tests too
- Add detection of the aws region in the `infra-ec2-template-destroy` role when fallback regions are used
- Create `setup_runtime.yml` playbook to be run as first step. Its job for the moment is to:
  - include `env_vars.yml` vars
  - set `output_dir` if it's not set already, create the dir if it doesn't exist
  - install galaxy roles.  If a `requirements.yml` file exists in the config dir, ansible galaxy roles are automatically downloaded and installed in the roles path. See [documentation on how to use galaxy roles in AAD](https://github.com/sborenst/ansible_agnostic_deployer/blob/development/docs/How_to_use_galaxy_roles_in_your_config.adoc).


fridim:
I just rebased from development and fixed conflicts.
@tonykay i had to get rid of your commits because of conflicts from development.

 About the `tools/archive` directory in this PR, i'm not a fan.

Why don't we just delete the configs? It's a git repository, if we need a previous configs, we can just `git checkout COMMIT|Branch` and then look at the files.
What is the point in keeping it ?

PR #293, opened 2018-12-03
Adding Istio/Prometheus CI/CD Demo

cricci82:


ansible/roles/ocp-workload-istio-prometheus-demo/templates/istio_cleanup.sh
```diff
@@ -0,0 +1,7 @@
+for i in `oc get clusterroles | grep istio`; do
```
fridim: You are not using `istio_cleanup.sh` anywhere. Can it be removed ?

fridim: @cricci82 did you see my comment ?
Are you done with the PR ? I can remove the istio_cleanup.sh myself if you want.

cricci82: Hi sorry for the delayed response.  Yes I saw and as a result I’m making some additional cleanups but haven’t had an opportunity to submit the update yet.

Thanks for the input

PR #299, opened 2018-12-18
Adding support for LE Staging Cert

pmccarthy:
##### SUMMARY
Adding support for LetsEncrypt staging Root Certificate

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
Integreatly Workload

##### ADDITIONAL INFORMATION
This change allows Integreatly to support LE staging environments for both Dev and Test while continuing support in Prod. 

It will mean that `--skip-tags production` will need to be set for both Dev and Test catalog items in Cloudforms

### DIFF
--- a/ansible/roles/ocp-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp-workload-integreatly/defaults/main.yml
@@ -12,3 +12,6 @@ admin_username: admin@example.com
 admin_password: Password1
 evals_username: evals@example.com
 evals_password: Password1
+letsencrypt_staging_root_cert_url: https://letsencrypt.org/certs/fakelerootx1.pem
+letsencrypt_staging_root_cert_path: /etc/origin/master/fakelerootx1.pem
+rhsso_identity_provider_ca_cert_path: "{{ letsencrypt_staging_root_cert_path }}"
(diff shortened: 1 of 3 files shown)


fridim:
We are using OCP Workshop config for letsencrypt and just pass the integreatly workload. This is a common config for several deployment. 

> It will mean that --skip-tags production will need to be set for both Dev and Test catalog items in Cloudforms

@pmccarthy Can you add a condition instead ? We pass the `lets_encrypt_production` variable, so you could do:

```yaml
- yourtask:
    ...
  when: lets_encrypt_production|bool
```

This way we don't have to skip tags.

pmccarthy:
@fridim Thanks for the review, I wasn't aware of the `lets_encrypt_production` var which should make things a lot easier. So to be clear, are you saying that for Dev and Test environments `lets_encrypt_production` is set to `false` while in Production it is set to `true`?

fridim:
> So to be clear, are you saying that for Dev and Test environments lets_encrypt_production is set to false while in Production it is set to true?

@pmccarthy correct

pmccarthy:
@fridim Excellent, thanks for confirming. I'll issue another PR in the next few minutes with these changes.

pmccarthy:
@fridim Updated changes added to follow up PR: https://github.com/sborenst/ansible_agnostic_deployer/pull/300

PR #300, opened 2018-12-19
Replacing prod tag with lets_encrypt_production var

pmccarthy:
##### SUMMARY
Updating LetsEncrypt logic in Integreatly role by using `lets_encrypt_production` instead of tags

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
Integreatly


pmccarthy:
@fridim FYI

PR #309, opened 2019-01-09
Changes for OSBA install and base Azure cloud provider configuration

vincepower:
##### SUMMARY

When on Azure configure the cloud provider configuration and install the open service broker for azure.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Updating host template and adding one new ocp-infra-azure-service-broker role

##### ADDITIONAL INFORMATION

```paste below

```


ansible/configs/ocp-workshop/files/hosts_template.3.10.45.j2
```diff
@@ -308,6 +308,26 @@ template_service_broker_install=true
...
 
+###########################################################################
+### Cloud Provider Specific Vars
+###########################################################################
+
+{% if cloud_provider == 'ec2' %}
+{% elif cloud_provider == 'azure' %}
+openshift_storageclass_parameters={'kind': 'managed', 'storageaccounttype': 'Premium_LRS'}
+openshift_cloudprovider_kind=azure
+openshift_cloudprovider_azure_client_id="{{ azure_service_principal }}"
```
fridim: Can we create a service principal on-the-fly when provisioning, and then use it here instead of our main SP ?
This way, if a user has SSH access to the environment, he/she cannot use it once the environment is gone.

vincepower: Yes, I can build that into the role for creation and then add it to the destroy_env role as well

vincepower: I'll reuse the same service principal in the install osba role

vincepower: OK, I've been fighting with this off and on for the last month and I don't think it is possible to have a new account created via automation as a service principal. You need to be logged in as a regular user account. I've asked Microsoft and internal Red Hat people and no one has a way to do it.

fridim: waw.. thanks for looking into this for so long.

Just out of curiosity, why can't we simply do:

```az ad sp create-for-rbac```

?

Comment (vincepower) [reply ansible/configs/ocp-workshop/files/hosts_template.3.10.45.j2]: To do automation they recommend a service principal be used and that is how all the scripts we have are setup.
If you run that command as a normal user account with Owner access on the subscription it will work.
If you run that command as a service principal, even if the role is Owner, it returns an error about not enough access, and for every doc I read and every person I talk to, there doesn't seem to be a way to do it.

File: ansible/cloud_providers/azure_destroy_env.yml [hunk 2]
```diff
@@ -10,4 +10,7 @@
     - name: Run infra-azure-template-destroy
       include_role:
         name: infra-azure-template-destroy
+    - name: Remove the service principal
+      include_role:
+        name: infra-azure-create-service-principal
```
fridim: i guess it should be `infra-azure-delete-service-principal` here ?

vincepower: You are correct, and I fixed the reference

PR #313, opened 2019-01-16
0.10 installer is out

thoraxe:
Updates OCP4 to use 0.10 installer


wkulhanek:
Thanks @thoraxe

wkulhanek:
@thoraxe Of course that wasn't enough..... install_config.yaml format changed in 0.10 as well...

PR #322, opened 2019-02-08
Added env_secret_vars creation

### LINKED ISSUE #73: automatically create env_secret_vars.yml file
playbooks fails if this files is not present because of `vars_file` in plays. Can we add a test and create it (empty) automatically if it does not exists?

gmontalvoy:
##### SUMMARY
Added a simply test and creation env_secret_vars.yml file into config/{{ env_type }} into main.yml

Resolves #73  

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
config

##### ADDITIONAL INFORMATION
As requested on #73 is just a env_secrets_vars.yml empty file in case it is not populated by user.


fridim:
Thanks @gmontalvoy for looking into this!

I commented on the original (obsolete) issue.
The issue is a non-issue now that we have dynamic inclusion of those files.

PR #368, opened 2019-04-02
ocp4 logging workload

thoraxe:
##### SUMMARY

OCP4 logging workload.
Dependent on infrastructure nodes existing (for now)

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME

ocp4-workload-logging


thoraxe:
Rebased and ready for merge.

wkulhanek:
LGTM. The only question which need to be tested if we run the infra role and then immediately this one will it work. Because I'm thinking that it takes a while (5 min) for the infra nodes to actually exist. Will need to test.

thoraxe:
@wkulhanek it will "work" in the context that the nodes don't need to exist yet to create a CR that specifies a nodeselector. However it might fail if the nodes take a really long time to come up, because I don't wait very long for the pods.

My guess is that we can add a wait state to the beginning of the logging playbook that checks to see if any nodes exist with the infra role and waits for up to 10 minutes for them to exist?

PR #369, opened 2019-04-03
OCP4 client vm 0.1

pavelanni:
##### SUMMARY

For the Advanced Development course we need some tools to be installed on the client VM: skopeo, java, Docker, etc. I've created a copy of `ocp-client-vm` role and added it to `ocp4-coreos-deployer`.


wkulhanek:
@fridim @pavelanni I think this is the wrong place. We don't need this on the OCP 4 Shared Cluster bastion VM. We only have one of those in a class. Students will get their own Client VMs from CloudForms.

We need to update the ocp-clientvm configuration to offer OCP 4 as a deployment target. Or create an ocp4-clientvm config.

pavelanni:
I agree with @wkulhanek . Essentially a Client VM is a Client VM, the only difference should be the version of the `oc` command. And we can even copy more than one version of `oc` into the VM and ask the student to set (or just use the preset) environment variable (`OC_CLI_VERSION` or something) to call the right one. Not an overcomplicated exercise. And a useful one, BTW.

fridim:
Ok sounds good. Let's remove from the config the line that loads the role

PR #370, opened 2019-04-05
Adds cluster autoscale for ocp4 and cleans some descriptions

thoraxe:
##### SUMMARY

This will create a cluster autoscaler (max size 9 nodes) and create a machineautoscaler for each existing machineset. In other words, it finds all the current machinesets in the cluster, and creates one machine autoscaler for each one.

This means that this role should come before any other roles that might create additional machinesets (eg: infra nodes).

It also means that for regions like us-east-1 where the cluster gets 6 sets but only 3 nodes (default), autoscalers will be created for machinesets that effectively end up unused. That's not really a problem, so w/e.

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME

ocp4-workload-cluster-autoscale

##### ADDITIONAL INFORMATION
none

### DIFF
--- a/ansible/roles/ocp4-workload-cluster-autoscale/readme.adoc
+++ b/ansible/roles/ocp4-workload-cluster-autoscale/readme.adoc
@@ -1,0 +1,118 @@
+= ocp4-workload-cluster-autoscale - Sets up cluster autoscaling and creates machineautoscalers
+
+== Role overview
+
+* This role configures the cluster autoscaler and creates machine autoscalers
+  for the default worker pool. It assumes a default cluster size of 3 workers
+  and will allow the cluster to scale up to 6 workers. It consists of the
+  following playbooks:
+** Playbook: link:./tasks/pre_workload.yml[pre_workload.yml] - Sets up an
+ environment for the workload deployment.
+*** Debug task will print out: `pre_workload Tasks completed successfully.`
+
+** Playbook: link:./tasks/workload.yml[workload.yml] - Used to set up autoscalers
+*** Debug task will print out: `workload Tasks completed successfully.`
+
+** Playbook: link:./tasks/post_workload.yml[post_workload.yml] - Used to
+ configure the workload after deployment
+*** This role doesn't do anything here
+*** Debug task will print out: `post_workload Tasks completed successfully.`
+
+** Playbook: link:./tasks/remove_workload.yml[remove_workload.yml] - Used to
+ delete the workload
+*** This role removes the infrastructure nodes (DANGER!!!)
+*** Debug task will print out: `remove_workload Tasks completed successfully.`
+
+== Review the defaults variable file
+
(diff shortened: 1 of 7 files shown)


ansible/roles/ocp4-workload-cluster-autoscale/tasks/workload.yml
```diff
@@ -0,0 +1,47 @@
+# vim: set ft=ansible
+---
+# Implement your Workload deployment tasks here
+- name: get current machinesets
+  shell: "oc get machineset -o name -n openshift-machine-api | awk -F/ '{print $2}'"
```
fridim: The `|awk` bothered me and i was curious: how to do it without a shell task ? i found those:

Use jsonpath:

```yaml
command:  >-
  oc get machineset -n openshift-machine-api
  -o jsonpath='{range .items[*]}{@.metadata.name}{"\n"}{end}'
```

or go template:

```yaml
command: >-
  oc get machineset -n openshift-machine-api
  -o go-template
  --template='{{range .items}}{{printf "%s\n" .metadata.name}}{{end}}'
```

or use lookup:

```yaml
- name: Fetch a list of namespaces
  set_fact:
    machinesets: "{{ lookup('k8s', api_version='machine.openshift.io/v1beta1', kind='MachineSet') | json_query('[].metadata.name') }}"

# then
- loop: "{{ machinesets }}"
  ...
```

ansible/roles/ocp4-workload-cluster-autoscale/readme.adoc
```diff
@@ -0,0 +1,118 @@
+= ocp4-workload-cluster-autoscale - Sets up cluster autoscaling and creates machineautoscalers
+
+== Role overview
+
+* This role configures the cluster autoscaler and creates machine autoscalers
+  for the default worker pool. It assumes a default cluster size of 3 workers
+  and will allow the cluster to scale up to 6 workers. It consists of the
```
fridim: `s/6/9/` ?

ansible/roles/ocp4-workload-cluster-autoscale/tasks/workload.yml
thoraxe: So are you saying you will not accept it as awk?

ansible/roles/ocp4-workload-cluster-autoscale/readme.adoc
thoraxe: Will fix

wkulhanek:
Well it’s very hacky when there is a proper way of retrieving the field you want....

--
Wolfgang

fridim:
@thoraxe we think the `lookup` version is cleaner.

Could you make the change ?

thoraxe:
@fridim @wkulhanek switched to k8s_facts from awk and fixed the node count number

fridim:
LGTM

PR #373, opened 2019-04-11
Rename ocp4-coreos-deployer to ocp4-ha-lab

fridim:
##### SUMMARY

- Rename ocp4-coreos-deployer to ocp4-ha-lab
- introduce `ocp4_base_domain` variable to all ocp4 configs to override top domain when deploying


wkulhanek:
@fridim LGTM. Should be good to merge. We just need to remember to update the deployed scripts when we tag..

PR #381, opened 2019-04-22
Deploys Istio control plane into OCP4

thoraxe:
##### SUMMARY
Deploys Istio control plane into OCP4

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME
Istio.

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/ansible/roles/ocp4-workload-istio-controlplane/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-istio-controlplane/defaults/main.yml
@@ -1,0 +1,7 @@
+---
+become_override: False
+ocp_username: opentlc-mgr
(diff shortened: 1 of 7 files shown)


wkulhanek:
@thoraxe why not use k8s to check/create the istio-operator project?
Also: doesn't the operator create the istio-system project? So we shouldn't need to create that one, should we?

thoraxe:
@wkulhanek the operator does not create istio-system. This follows the docs exactly.

updated to use `k8s` module for project creation and restores the defaults which are needed

wkulhanek:
LGTM

PR #390, opened 2019-04-24
Added fail tasks if admin_password is not defined

### LINKED ISSUE #374: Default ocp-workload-terminal fails late because of undefined variable
**Describe the bug**
When deploying ocp4-workshop, the playbook fails at the end if the variable `admin_password` is not set.

```
TASK [ocp-workload-terminal : Create OpenShift objects for Terminal workload] ***
Friday 12 April 2019  12:43:51 +0000 (0:00:00.061)       0:40:38.484 ********** 
changed: [clientvm.e435.internal] => (item=./templates/project.j2)
changed: [clientvm.e435.internal] => (item=./templates/service.j2)
changed: [clientvm.e435.internal] => (item=./templates/service_account.j2)
changed: [clientvm.e435.internal] => (item=./templates/role_binding.j2)
fatal: [clientvm.e435.internal]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'admin_password' is undefined\n\nThe error appears to have been in '/tmp/ocp4-workshop-e435/ansible_agnostic_deployer/ansible/roles/ocp-workload-terminal/tasks/workload.yml': line 8, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Create OpenShift objects for Terminal workload\n  ^ here\n"}
```

**Expected behavior**
* The password should be generated and printed using `user.info`
* Or the playbook should fail *early* with an explicit error message to let user know that the variable is missing.
* Or the workload should not be a default workload.

gmontalvoy:
##### SUMMARY
Deploy ocp_workload will fail late if admin_password is not defined
Fixes #374 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-workshop

##### ADDITIONAL INFORMATION
A simple task to early failing if admin_password is not defined, maybe it could serve as an approach to fix issue #374


fridim:
Thanks for the PR @gmontalvoy 

few things:

1. I want this issue to be address more globally too, with something like:
   * load `pre_checks.yml` if it exists in `configs/CONFIGNAME` dir
   * Or maybe have a `mandatory_variables` list, but that would require some python module code
   * I don't know yet what's the cleanest, but you got the idea.
1. The goal here is to fail early. Why did you put it in the software playbook ?  why not `pre_infra.yml` ?

Thanks

gmontalvoy:
Thank you for your answer @fridim 

- I will take a look to load the pre_reqs only if it fits on the config name.

- I wouldn't do it using a mandatory_variables list, We are currently using a list of variables that should be configured previously on a env_vars file to get a successful deployment, this would be considered as a kind of "mandatory vars", having different places with mandatory variables would be a little confusing, to make it short, IMHO "the simplest/cleanest, the better"

- I considered admin_password as part as software step, but in a second though I could include that into the infra pre (maybe a part of prerequisites check to make it fail as soon as possible?)

Germán

wkulhanek:
@gmontalvoy We ended up implementing this differently. If admin_password is not set then it is automatically generated and printed via user.info to be included in the provisioning e-mail. Therefore I'm closing this Pull Request

gmontalvoy:
Thank you @wkulhanek 
 
I assume this was discussed at any moment I missed, it is already done?

wkulhanek:
Yes. It is implemented.

PR #404, opened 2019-05-01
Move to registry redhat io

DuncanDoyle:
##### SUMMARY
Due to the shutdown of registry.access.redhat.com, all the RHPAM and RHDM demos started to fail. This PR fixes this by:
- Checking whether the required ImageStreams are in the "openshift" namespace
- If not present, provision the ImageStreams in the "openshift" namespace

I need to use the "openshift" namespace instead of the project namespace, as I need the credentials to pull from "registry.redhat.io", which are present in the "openshift" namespace, but not in the project namespace.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- rhpam-order-it-hw
- rhdm-qlb-loan
- fsi-cc-dispute

##### ADDITIONAL INFORMATION
Talked to Sha to discuss possible fixes, and this seemed the best option (unitil we move to a dedicated Quay instance).

### DIFF
--- a/ansible/roles/ocp-workload-dm7-qlb-demo/defaults/main.yml
+++ b/ansible/roles/ocp-workload-dm7-qlb-demo/defaults/main.yml
@@ -33,7 +33,7 @@ kie_server_user: kieserver
 kie_server_pwd: kieserver1!
 pv_capacity: 512Mi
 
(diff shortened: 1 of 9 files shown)


wkulhanek:
LGTM. Only hits those workloads so effect should be contained.

PR #406, opened 2019-05-01
Additional vars for workshopper role

thoraxe:
adds additional vars (detected from environment) to the workshopper workload role


wkulhanek:
LGTM

PR #407, opened 2019-05-01
Update  ocp-workload-odo-workshop to support new deploy scripts

jankleinert:
With a change to how we deploy this workshop, the workload needed to be updated.


wkulhanek:
LGTM. One comment I would make is that there are a lot of Shell commands for which there are Ansible Modules (like file instead of rm). But it'll work...

PR #413, opened 2019-05-04
Bumping Integreatly release version v1.4.0-rhpds

trepel:
##### SUMMARY

Updating Integreatly release version to v1.4.0-rhpds

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME

Integreatly

##### ADDITIONAL INFORMATION

There is this `-rhpds` suffix because the release is intended to be used on RHPDS only.

```paste below

```

### DIFF
--- a/ansible/roles/ocp-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp-workload-integreatly/defaults/main.yml
@@ -4,7 +4,7 @@ become_override: True
 silent: False
 install_dir: /tmp/integreatly
 inventory_hosts_file: inventories/hosts.template
-release_tag: release-1.3.0
+release_tag: release-1.4.0-rhpds
 webapp_namespace: webapp
 self_signed_certs_enabled: True
 openshift_master_config_path: /etc/origin/master/master-config.yaml


trepel:
@fridim @NonyNo3lle Mind taking a look when you get a chance, please? Thanks.

fridim:
@trepel is this for RHPDS or for summit or both ?
It seems late if it's for summit ?

tremes:
@fridim This is primarily for RHPDS AFAIK.

trepel:
@fridim @NonyNo3lle would you be able to spare some time on this. It would be good to have this soon after summit. Thanks.

trepel:
@fridim @NonyNo3lle we found out that some people still depend on older Integreatly version so we decided to postpone the 1.4 release to RHPDS until there is either version selector or brand new Integreatly workshop with 1.4 version (the older 1.3 workshop will be kept alongside this new one).

Thus there is no need to merge now

fridim:
@trepel sure. Let us know when you want to bump the version.

pmccarthy:
@trepel We've since changed how release tags get passed into the integreatly workload playbook where we now read in the version that gets selected from the cloud forms dropdown menu. The process from now on will be to update a json file [1] which gets used to populate the dropdown.

Based on the above mentioned, we can probably close out this PR. 

[1] https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/ocp-workload-integreatly/files/supported_versions.json

PR #416, opened 2019-05-04
ocp4-workload-build-your-own-operator

ryanj:
##### SUMMARY
ocp4-workload-build-your-own-operator

### DIFF
--- a/ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
+++ b/ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
@@ -1,0 +1,31 @@
+---
+- environment:
+    KUBECONFIG: "{{ tmp_kubeconfig }}"
+  block:
+    - name: post_workload Tasks Complete
+      debug:
+       msg: "Pre-Software checks completed successfully - Removed"
+  
+    - name: Git clone the repo if it doesn't exist
+      command: git clone --branch v1.0 https://github.com/openshift-labs/lab-build-your-own-operator {{ tmp_git_location }}
+      args:
+        creates: "{{ tmp_git_location }}"
+
+    - name: Delete spawner
+      command: "{{ tmp_git_location }}/.workshop/scripts/delete-spawner.sh"
+
+    - name: Delete workshop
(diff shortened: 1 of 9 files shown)


ansible/roles/ocp4-workload-build-your-own-operator/tasks/workload.yml
```diff
@@ -0,0 +1,29 @@
...
+      shell: "oc get project {{user_project_name}}"
+      register: workshop_project
+      ignore_errors: true
+
+    - name: Create project for user workshop if it doesn't exist
+      shell: "oc new-project {{user_project_name}} --display-name='user workshop'"
+      when: workshop_project is failed
+
+    - name: Git clone the repo if it doesn't exist
+      command: git clone --branch v1.0 https://github.com/openshift-labs/lab-build-your-own-operator {{ tmp_git_location }}
```
fridim: if i understand correctly, you're using an external repo and a script to deploy the workload: https://github.com/openshift-labs/lab-build-your-own-operator/blob/v1.0/.workshop/scripts/deploy-spawner.sh

That's out of our process. We are using tags in agnosticd to pin a working state.
If you're using a branch you're bypassing this. If you push changes to this v0.1 branch, then deployments may fail.
I'm not against using files from an external repo, but can we use a commit ID instead of a branch ? So the version is pinned ?
(`git clone` + `git checkout COMMITID`)

ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
```diff
@@ -0,0 +1,31 @@
+---
+- environment:
+    KUBECONFIG: "{{ tmp_kubeconfig }}"
+  block:
+    - name: post_workload Tasks Complete
+      debug:
+       msg: "Pre-Software checks completed successfully - Removed"
+  
+    - name: Git clone the repo if it doesn't exist
+      command: git clone --branch v1.0 https://github.com/openshift-labs/lab-build-your-own-operator {{ tmp_git_location }}
```
fridim: same here

jankleinert:
This can be closed in favor of https://github.com/redhat-cop/agnosticd/pull/418

fridim:
fixed by #418

PR #418, opened 2019-05-05
Ocp4 workload build your own operator (replaces PR 416)

jankleinert:
This is a replacement for https://github.com/redhat-cop/agnosticd/pull/416, which addresses the issues raised there. 

The workload is now referencing a tagged release rather than just a branch https://github.com/openshift-labs/lab-build-your-own-operator/releases/tag/v1.1

Also fixed another task that needed chdir added to it.

### DIFF
--- a/ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
+++ b/ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
@@ -1,0 +1,31 @@
+---
+- environment:
+    KUBECONFIG: "{{ tmp_kubeconfig }}"
+  block:
+    - name: post_workload Tasks Complete
+      debug:
+       msg: "Pre-Software checks completed successfully - Removed"
+  
+    - name: Git clone the repo if it doesn't exist
+      command: git clone --branch v1.0 https://github.com/openshift-labs/lab-build-your-own-operator {{ tmp_git_location }}
+      args:
+        creates: "{{ tmp_git_location }}"
+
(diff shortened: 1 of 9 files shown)


ansible/roles/ocp4-workload-build-your-own-operator/tasks/remove_workload.yml
```diff
@@ -0,0 +1,35 @@
+---
+- environment:
+    KUBECONFIG: "{{ tmp_kubeconfig }}"
+  block:
+    - name: post_workload Tasks Complete
+      debug:
+       msg: "Pre-Software checks completed successfully - Removed"
+  
+    - name: Git clone the repo if it doesn't exist
+      command: git clone --branch v1.1 https://github.com/openshift-labs/lab-build-your-own-operator {{ tmp_git_location }}
```
fridim: @jankleinert you changed the name of the branch from `v1.0` to `v1.1`.
I'll re-post my previous comment as it's still valid in this case:

> That's out of our process. We are using tags in agnosticd to pin a working state.
If you're using a branch you're bypassing this. If you push changes to this v0.1 branch, then deployments may fail.
I'm not against using files from an external repo, but can we use a commit ID instead of a branch ? That way the version would be pinned ?
(git clone + git checkout COMMITID)

jankleinert: @fridim commented in chat as well, but to have the note here too, v1.1 is a tag. So this git clone --branch v1.1 <repo url> command is cloning that release now

PR #420, opened 2019-05-06
IoT next gen workshop

sabre1041:
Automation for IoT Next Generation Technologies lab

### DIFF
--- a/ansible/roles/ocp4-workload-iot-next-gen-technologies/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-iot-next-gen-technologies/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+
+become_override: false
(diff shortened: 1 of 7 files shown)


ansible/roles/ocp4-workload-iot-next-gen-technologies/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+become_override: false
+repository_archive_url: https://gitlab.com/redhatsummitlabs/next-gen-technologies-at-scale-building-solutions-to-manage-tomorrows-workloads/-/archive/master/next-gen-technologies-at-scale-building-solutions-to-manage-tomorrows-workloads-master.tar.gz
```
fridim: Can we can pin a version on this ?
(commit ID)

The goal is that if you push something to redhatsummitlabs/next-gen-technologies... it doesn't break deployments.

PR #424, opened 2019-05-14
Adding OpenShift Ansible Better Together workshop

deweya:
Adding OpenShift Ansible Better Together workshop for dev track.

Is a copy of the ocp4-workshop with our stuff added on top. Ours config is called "openshift-ansible-better-together"

Essentially, upped the user count to 30, are installing docker and operator-sdk on the bastions, and created a workload called "ocp4-workload-openshift-ansible" that will create CRDs to the cluster.


wkulhanek:
Is it really necessary to fork this config rather than just using workshop with the right workloads and parameters? How will you keep up with future changes?

deweya:
From what I can tell, the only thing we really need on top of the ocp4 base config is docker and operator-sdk installed on the bastion. Part of the lab will include testing and building an Ansible operator with the operator-sdk. 

Everything else we should be able to configure with env_vars.yaml and a custom workload. Perhaps we could further configure the bastions with a post-provisioning script, but out of the box the config doesn’t provide everything we need.

deweya:
@wkulhanek Could you please review again? Now that I understand a little better, I realized I could do what I needed with just a custom workload. So I removed the new config and moved the logic I had there into a custom workload.

wkulhanek:
LGTM. Thanks @deweya for making these changes. MUCH less risky.... I have a feeling we'll be doing some surgery on ocp4-workshop shortly and this will keep you whole. :-)

PR #365, opened 2019-03-29
First try for including Hybrid Edge Deployment demo in to the roles

alezzandro:
##### SUMMARY

As discussed with RHPDS team this is the first try for including Hybrid Edge Deployment demo in to the roles
##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
ocp-workload-edge-deployments

##### ADDITIONAL INFORMATION
*PLEASE NOTE*: the role is actually partial complete. This demo requires an All-In-One standalone OpenShift installation with *also* an additional standalone RHEL system to work with. Actually this pull request only refers to creating the environment needed for the OpenShift part, after that I need to verify with RHPDS team how I can create and configure a RHEL host (outside from OCP), setup all the pre-requisites and deploy the SSH key that I create through this Ansible Role.

PR #430, opened 2019-05-20
Pull Request for 'Edge Deployments' role

alezzandro:
## SUMMARY
As discussed with RHPDS team this is the PR for including Hybrid Edge Deployment demo in to the roles

## ISSUE TYPE
New role Pull Request
COMPONENT NAME
ocp-workload-edge-deployments

## ADDITIONAL INFORMATION
**PLEASE NOTE**: This demo requires an All-In-One standalone OpenShift installation with also an additional standalone RHEL system to work with: the demo will install additional stuff on the bastion node. This pull request will build the OpenShift part, after that it will also configure the bastion RHEL host (outside from OCP), setup all the pre-requisites and deploy the SSH key that it creates through this Ansible Role.

This a brand new PR after that we decided to close the previous one with so many commits in it: https://github.com/redhat-cop/agnosticd/pull/365

### DIFF
--- a/ansible/roles/ocp-workload-edge-deployments/tasks/remove_workload.yml
+++ b/ansible/roles/ocp-workload-edge-deployments/tasks/remove_workload.yml
@@ -4,38 +4,79 @@
   debug:
     msg: "Cleaning up the environment.."
 
+- name: Check if project IoT Development exists
+  shell: "oc get project iot-development 2>/dev/null | wc -l"
+  register: result
+
 - name: Delete project IoT Development
   shell: "oc delete project iot-development"
-  ignore_errors: true
+  when: result.stdout != "0"
+
+- name: Check if project IoT Testing exists
+  shell: "oc get project iot-testing 2>/dev/null | wc -l"
+  register: result
 
 - name: Delete project IoT Testing
   shell: "oc delete project iot-testing"
-  ignore_errors: true
+  when: result.stdout != "0"
+
+- name: Check if project IoT Hub exists
+  shell: "oc get project iot-hub 2>/dev/null | wc -l"
+  register: result
 
 - name: Delete project IoT Hub
   shell: "oc delete project iot-hub"
-  ignore_errors: true
+  when: result.stdout != "0"
+
+- name: Check BuildConfig in openshift namespace exists
+  shell: "oc get bc deploy-containers-apb -n openshift 2>/dev/null | wc -l"
+  register: result
 
 - name: Delete BuildConfig in openshift namespace
   shell: "oc delete bc deploy-containers-apb -n openshift"
-  ignore_errors: true
+  when: result.stdout != "0"
+
+- name: Check ImageStream in openshift namespace exists
+  shell: "oc get is deploy-containers-apb -n openshift 2>/dev/null | wc -l"
+  register: result
 
 - name: Delete ImageStream in openshift namespace
   shell: "oc delete is deploy-containers-apb -n openshift"
-  ignore_errors: true
+  when: result.stdout != "0"
+
+- name: Check that the temp dir exists
+  stat:
+    path: /tmp/{{guid}}
+  register: stat_result
+
+- name: Getting the generated public-key into this playbook
+  slurp:
+    src: /tmp/{{guid}}/deploy-containers-apb/id_rsa.pub
+  register: publickey
+
+- name: Removing SSH public key to the authorized host for the machine
+  become: true
+  authorized_key:
+    user: root
+    state: absent
+    key: "{{ publickey['content'] | b64decode }}"
 
 - name: Cleaning up temp dir
   shell: "rm -rf /tmp/{{guid}}"
-  ignore_errors: true
+  when: stat_result.stat.exists == True
+
+- name: Check that the backup file exists
+  stat:
+    path: /etc/containers/registries.conf.bak
+  register: stat_result
 
 - name: Restoring up /etc/containers/registries.conf
   become: true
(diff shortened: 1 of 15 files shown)


ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
```diff
@@ -0,0 +1,133 @@
...
+- name: Deleting the fis-karaf ImageStream
+  shell: "oc delete -n openshift is fis-karaf-openshift"
+
+- name: Creating the fis-karaf ImageStream
+  shell: "oc create -n openshift -f /tmp/{{guid}}/fis-karaf-openshift-is-registry.yaml"
+
+# Starting the configuration for the External Machine. We suppose that the External machine is the bastion where we're launching this ansible playbook.
+- name: Adding SSH public key to the authorized host for the machine
+  become: true
+  shell: "cat /tmp/{{guid}}/deploy-containers-apb/id_rsa.pub >> /root/.ssh/authorized_keys; chmod 700 /root/.ssh/authorized_keys"
```
fridim: you can use the `authorized_key` module https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
```diff
@@ -0,0 +1,133 @@
...
+  shell: "oc new-project iot-hub"
+
+- name: Create resources for IoT Hub
+  shell: "oc process -f /tmp/{{guid}}/project-iot-hub.yaml | oc create -f -"
+
+- name: Create policy for letting Jenkins ServiceAccount Access to iot-testing Project
+  shell: "oc policy add-role-to-user edit system:serviceaccount:iot-development:jenkins -n iot-testing"
+
+- name: Generate a brand new SSH key for connecting to the remote machine
+  shell: "cd /tmp/{{guid}}/deploy-containers-apb; ssh-keygen -N '' -f ./id_rsa"
```
fridim: Do you really need to create a new ssh key ?
the bastion should be able to access all the instances, as root and using the internaldns name.

Also, you're using `id_rsa.pub` later, so you need

```shell
ssh-keygen -f ./id_rsa -y > id_rsa.pub
```

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
fridim: You can use the `authorized_key` ansible module : https://docs.ansible.com/ansible/latest/modules/authorized_key_module.html

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
fridim: Do you really need a new key ?
The bastion should be able to ssh to any machine, as root and using the internaldns for remote host.

Also, you're using `id_rsa.pub` later so you probably need:

```shell
ssh-keygen -f id_rsa > id_rsa.pub
```

ansible/roles/ocp-workload-edge-deployments/tasks/remove_workload.yml
```diff
@@ -0,0 +1,47 @@
...
+  ignore_errors: true
+
+- name: Cleaning up temp dir
+  shell: "rm -rf /tmp/{{guid}}"
+  ignore_errors: true
+
+- name: Restoring up /etc/containers/registries.conf
+  become: true
+  shell: mv /etc/containers/registries.conf.bak /etc/containers/registries.conf
+  ignore_errors: true
```
fridim: i'm not a big fan of all those `ignore_errors`. It looks like cheap idempotency.
They will appear as error in the logs, and will catch the eyes if someone is trying to troubleshoot.

Ideally (if time permits) i would test for existence before trying to delete.

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
alezzandro: I prefer to create and use a brand new key because the default key will allow Demo users to connect to any OCP nodes. Generating a brand new key, adding it ONLY to the authorized_keys of the bastion will ensure that user can use the key only for connecting to the bastion.
Is that ok for you?

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
alezzandro: yes, as I said, I'll fix it

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
alezzandro: replied before

ansible/roles/ocp-workload-edge-deployments/tasks/remove_workload.yml
alezzandro: What do you suggest? Continue using `shell` module in a way like this?
`oc get projects|grep iot-development && oc delete project iot-development`

Or using [register variables](https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#register-variables)?

fridim: "register variables" is a more common ansible pattern.

ansible/roles/ocp-workload-edge-deployments/tasks/workload.yml
fridim: Sure

alezzandro:
@fridim made all modifications as per your request. I've also tested them, now it should be fine.

PR #455, opened 2019-06-12
Change limits and quotas

matskiv:
##### SUMMARY
a) changed minimal requested memory to allow Launcher deployment ([it's currently requesting 8M](https://github.com/fabric8-launcher/launcher-operator/blob/d0124534296c3d1239fb0110ce70876bfe9dd3fa/templates/fabric8-launcher.yaml#L225-L229)). I picked "6Mi" value to match min requested memory for a pod.

b) **dropped**, see discussion below ~~bumped UserQuota for OCP4 workshop. This will enable opentlc-mgr user to utilize all available resources (I calculated CPU and memory based on 3x m4.2xlarge worker nodes available).~~
~~This is mainly motivated by the problems discovered during Integreatly installation. We are starting to test new MW products on OpenShift 4.~~

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp-workshop, ocp4-workshop

##### ADDITIONAL INFORMATION
**dropped**, see discussion below ~~P.S: This change is motivated by 3Scale Redis deployment -~~`requests.memory: 32Gi`

### DIFF
--- a/ansible/configs/ocp-workshop/files/project-template.j2
+++ b/ansible/configs/ocp-workshop/files/project-template.j2
@@ -36,7 +36,7 @@ objects:
           max:
             memory: 6Gi
           min:
-            memory: 10Mi
+            memory: 6Mi
           default:
             cpu: 500m
             memory: 1.5Gi
--- a/ansible/roles/ocp4-workload-project-request-template/templates/project_request_template.j2
+++ b/ansible/roles/ocp4-workload-project-request-template/templates/project_request_template.j2
@@ -16,7 +16,7 @@ objects:
           max:
             memory: 6Gi
           min:
-            memory: 10Mi
+            memory: 6Mi
           default:
             cpu: 500m
             memory: 1.5Gi


matskiv:
@fridim @NonyNo3lle @rut31337 can you take a look, please?
Thank you! :)

wkulhanek:
@matskiv I dont' want to set these for every single workshop. Can you just override them in your integreatly workload? You can just apply a new UserQuota or ProjectRequestTemplate using the k8s module.

matskiv:
@wkulhanek If any other workshop will want to install new Launcher or 3Scale, they will bump into the same limits.
What potential problem do you expect from this changes?

wkulhanek:
You are asking to change every single workshop ever deployed. We don't want or need these big quotas on regular workshops. If anyone wants/needs bigger quota they can just use the mechanism to overwrite what they request in their specific environment.

matskiv:
@wkulhanek Ok, I will drop a commit that's increasing the quota.

Are you okay with lowering min.memory in LimitRange? This is safe change IMHO.

wkulhanek:
Yes. Lowering the minimum is definitely fine.

PR #498, opened 2019-07-15
configs/ocp-workshop : Making NFS server to use public DNS

pranavgaikwad:
##### SUMMARY
This change makes the NFS server to use a public DNS name instead of an internal name. This change allows the NFS server to be used from other OpenShift cluster without relying on a transient IP address. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
configs/ocp-workshop


fridim:
I'd like to discuss this.

IIUC, one consequence of this PR is that now support instances will always have an Elastic IP associated to it.
- +1 EIP per stack
- +public DNS entry for support

Also what are the security implications for opening the NFS server ?

I would prefer if we implement security groups properly before opening NFS servers to the internet.
When NFS is on the private network, it's fine to have lines like:

```
/srv/nfs/{{ item }} *(insecure,rw,root_squash,no_wdelay,sync)
```

But if it's open to the internet it can quickly get out of control.

wkulhanek:
Agreed. I don’t think we should do this at all.

pranavgaikwad:
@fridim @wkulhanek  I get your point. The current NFS in AgnosticD (without the DNS) is still open to internet. It just does not have an EIP and dns. That's the reason I thought it would be okay to have a public dns record for it. I would like to know if we can have an extra variable to decide whether to use public dns or not. We can keep the default set to false. I can also look into making NFS more secure. Any thoughts?

jwmatthews:
@wkulhanek the reasoning behind this request is we'd like to support a scenario in OCP App Migration Labs where user has a PV using NFS on a source cluster, we do a migration and have the destination cluster use the same NFS share as the source used for the PV.    

We call this 'Swinging the PV', it's one way a user may do a migration if they have a remote shared volume and they don't want to copy data, they would just unhook it from source side and reattach on destination side....that's the scenario we'd like to allow in a lab.

fridim:
> The current NFS in AgnosticD (without the DNS) is still open to internet. 

that's wrong and should be fixed in the security groups.

The `instances` variable allows user to specify for each instance if it has an eip or not:

```
instances:
  - ...
    public_dns: '{{ true if install_glusterfs | bool else false }}'
```

You can just override this to true for your deployment, but i don't want to make it true by default with `install_nfs`.

We can have another variable `support_public_dns: '{{ true if install_glusterfs | bool else false }}'` and use it in the `instances` list. That way it's easy for you to override the value when deploying.

But keep in mind that we will want to fix:

> The current NFS in AgnosticD (without the DNS) is still open to internet. 

So you will probably also have to update the security groups.

PR #502, opened 2019-07-19
Added a pause to wait for the actual builds to be created before we c…

DuncanDoyle:
…heck whether they're finished.

##### SUMMARY
Bugfix ocp-workloap-optaweb-vrp deployment. The task that checks for the builds to be finished was executed before the system had the time to actually create the builds, which caused the deployment to continue and fail on the "wait for deployment".

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp-workload-optaweb-vrp

##### ADDITIONAL INFORMATION
n.a.


wkulhanek:
It would be better to check if the build has finished than just adding a random pause. On heavy load the time it takes might vary.

PR #505, opened 2019-07-22
multi-region-tower config

miteshrh:
##### SUMMARY
multi-region-tower config is ready for review

##### ISSUE TYPE
New config
New roles

PR #506, opened 2019-07-22
multi-region-tower WIP

fridim:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New config Pull Request
- New role Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


fridim:
Before merging this PR, some of the points from the previous #505 :


The fact that workers needs ssh port 22 access to it, without bastion, make the whole infra wonky:

i don't see the point of having a bastion if we can ssh to the worker to reach private network of all the environment?

We should try the following:

1. Create the primary stack
1. generate the ssh conf
1. create the secondary stacks
1. Create all the VPC peering
1. only then, generate the ssh config again by specifying `bastion_hostname` = `groups.bastions[0]`

----
```diff
- pg_host='tower1.{{chomped_zone_internal_dns}}'
+ pg_host='tower1.{{target_regions[0].name}}.{{guid}}{{subdomain_base_suffix}}'
```

Why are we using the public dns for the database ?

Regarding the DB, i thought we wanted to try RDS so we don't have to manage it?

----
The templates worker.j2 and default.j2 have only one line diff

```diff
-{% for instance in instances %}
+{% for instance in instances_worker %}
```

We should not need two nearly identical templates just for that.

I would prefer is we add the capability to filter instances per region
(easy fix in `default.j2` template)

----
fix the potential mismatch in jinja2 templates between `loop.index` and `host`

sborenst:
Please, let's schedule a meeting to approve the PR. (With me) 

Yeay, Multi region tower. @prakhar1985 @miteshrh

prakhar1985:
@sborenst I have scheduled a meeting at 2:30 PM 25th July Thursday. This time will be good for @miteshrh as well.

sborenst:
Removing this merge request. we would need to rebase as it has gone stale. 
@prakhar1985

PR #516, opened 2019-07-26
New config for ARO

vincepower:
##### SUMMARY
Adding a config that will configure and remove Azure Red Hat OpenShift

##### ISSUE TYPE
- Feature Pull Request
- New config Pull Request

##### COMPONENT NAME
ARO config that has everything needed to deploy using already generated Application Registrations and administration groups in Azure's Active Directory. The App Registrations are management by an Azure Function and Cosmos DB. 

##### ADDITIONAL INFORMATION
This has user.info for output of the URL


ansible/configs/aro/env_vars.yml
```diff
@@ -0,0 +1,34 @@
...
+key_name: id_rsa
+
+# Setting default remote_user
+remote_user: azure
+
+# Setting some defaults that are checked for
+HostedZoneId: none
+
+# Setting this variable that specifies the group with AD admins in it
+az_group_customer_admin: b61baaf0-da29-4d17-836d-00e205f02aa6
```
wkulhanek: Is publishing this a security risk?

vincepower: The object id is just the reference to the a group that contains the "admin user" for ARO. They need access to our subscription to query the object and it just contains a pointer to a single user object.

I guess we could pass it in as a parameter from RHPDS and if we are doing that then I'd also like to pass in the base URL also included in the env_vars (az_function_*). Both could/would change if someone was to have a second Azure subscription.

vincepower: I'm submitting a change right now that will address this.

rut31337: @bosebc once Wolfgang accepts the PR please start the process of creating the catalog item in RHPDS.

vincepower: @wkulhanek I've updated this and added a note in README.md to reflect the new parameter needed.

PR #583, opened 2019-08-15
user passwords are incorrect in the email

philbrookes:
##### SUMMARY
These password variables had different values every time they were read :man_facepalming: 

Using a filename makes them sticky values.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
integreatly

@mikenairn @pmccarthy


sborenst:
@philbrookes Does this workload only run once? If so, this would be fine, if it runs multiple times, would is overwrite the file? Is that an issue? 

Let me  know and we can approve.

NonyNo3lle:
This runs only once, it's an infra workload.

PR #601, opened 2019-08-19
Add ocp-student-workloads.yml

kaovilai:
##### SUMMARY

Add ocp-student-workloads.yml

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME

ocp-student-workloads.yml
##### ADDITIONAL INFORMATION
User will be able to run multiple student-roles on their local machine in one go by providing CSV of student_workloads

```paste below

```

### DIFF
--- a/ansible/configs/ocp-student-workloads/ocp-student-workloads.yml
+++ b/ansible/configs/ocp-student-workloads/ocp-student-workloads.yml
@@ -1,0 +1,15 @@
+---
+- name: Deploy a workload role on a master host
+  hosts: all
+  become: false
+  gather_facts: False
+  tags:
+    - step007
+  tasks:
+  - name: Deploy ocp_workloads
+    include_role:
+      name: "{{ one_ocp_workload }}"
+    become: yes
+    loop: "{{ ocp_workloads.split(',') | list }}"
+    loop_control:
+      loop_var: one_ocp_workload


wkulhanek:
This isn't student workloads though? The way this is written it's infra-workloads. For student workloads it would also need a loop var to set up the students (user{1..x}), wouldn't it?

Infra Workload: Run a loop over a CSV list of workloads
Student Workload: Run a double loop over CSV list of workloads and users {1..num_users}

kaovilai:
Just something to quickly test. Is infra workloads a better name?
Just as extension of ocp-workload.yml

wkulhanek:
Yes, I think infra-workloads would be better.

stencell:
@wkulhanek I don't think we should take this as it is. Instead, a more formal review and updating of the [original ocp-workloads config](https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/ocp-workloads/ocp-workload.yml) should happen. 

Since we have 3 different kinds of workloads right now for OCP4 (student-, infra-, default-), just have one config that can handle all 3 instead of 3 that all do the same thing for fairly arbitrarily named workloads. 

Of course, that brings up the question of why and if we really actually need to differentiate all 3 of these workload types? Why can't they just be "workloads"? If someone wants to add or remove from that workloads list for their deployment, that is their call...I don't know that we have any defaults that are mandatory? And even if we do, they can just go under a mandatory-workload so then we are at least down to 2x types instead of 3x.

thoraxe:
High level user story --

> As a developer of *-workloads for agnosticd/v
> I want to be able to easily deploy *-workloads into an existing cluster
> From my local machine or from the bastion host

Let's not lose the forest for the trees here.

stencell:
I fail to see how having a single type of workload defeats that user story. It makes it easier IMO. I no longer have to care about infra-workload or student-workload or default-workload or understand what each is used for. I just have a single var (in a file, command line, or agnosticv) that has my list of workloads. Don't want logging or TSB or quota or infra or whatever...just delete it from that var list to get as close to vanilla or fully tweaked out cluster that you need.

Adding this as it is, now we would have 2 configs - ocp-workloads and ocp-student-workloads. Plus we have an ocp-workloads playbook embedded in the 3 OCP4 configs (shared, ha-lab, workshop) that are all at slightly different versions.

I don't think fixing this right and having a unified single config or shared playbook is missing the forest for the trees vs spending a bit of extra effort to make it more future proof and easy to use for the people in the story.

kaovilai:
@sborenst Want to chime in?
Stability of not messing with current ocp-workloads vs making one unified file that does it all.

fridim:
Hello,

There are two things going on here.

1/ is the PR itself, which offers a student-workload playbook, but miss half the loops (it loops workloads but not users).

2/ naming and design discussion


When developing a student workload, i see how it can be useful to have a config to test it against a running cluster.

So, if you think it is useful and fix this PR by adding the user{1..x}, it can be considered. I would also not create another config directory but put this under `configs/ocp-workloads/ocp-student-workload.yaml` instead.


Regarding the design and naming of workloads, i think we have a simple but naive implementation. It's helped us achieve a lot, but now we need to improve on that.

For example, we want to implement the capability to select a git tag for each workload versus today "deploy config + workload using the same git tag".

```yaml
ocp_workloads:
  - name: ocp-workload-coolstuff
    type: infra
    scm_ref: ocp-workload-coolstuff-prod-1.2
  - name: ocp-workload-sampleapp
    type: student
    scm_tag_prefix: ocp-workload-sampleapp-prod
    users: "{{ ... }}"
```

(only giving this as an example to illustrate)

I would prefer if we move this discussion to a dedicated thread or issue/RFE.

fridim:
What if user deploy his own cluster, and then test his *student* workload on it ?

Then it makes sense to have such a playbook, no ?

kaovilai:
> What if user deploy his own cluster, and then test his _student_ workload on it ?
> 
> Then it makes sense to have such a playbook, no ?

That was the original intent. Assume there maybe only one ocp_username i.e. kubeadmin since the cluster for development may only have one user.

This PR avoids looping around creating users so that the user can themselves export their desired existing ocp_username for development which is guaranteed to already be created and exist in OpenShift.

fridim:
We recently updated the [ocp-workloads config](https://github.com/redhat-cop/agnosticd/tree/development/ansible/configs/ocp-workloads). It is now a config like any other, see its Readme.

It should now properly address the user story:

> As a developer of *-workloads for agnosticd/v
> I want to be able to easily deploy *-workloads into an existing cluster
> From my local machine or from the bastion host

I'm closing this PR. Freel free to re-open or discuss.

PR #617, opened 2019-08-21
ocp-workload-migration : Sync with latest changes in operator

pranavgaikwad:
##### SUMMARY
Getting new changes in migration operator which this workload depends on. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ocp-workload-migration

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/ocp-workload-migration/tasks/pre_workload.yml
+++ b/ansible/roles/ocp-workload-migration/tasks/pre_workload.yml
@@ -49,7 +49,7 @@
       owner: "{{ student_name }}"
       mode: "u+rwx"
     loop:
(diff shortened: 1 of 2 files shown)


ansible/roles/ocp-workload-migration/tasks/pre_workload.yml
```diff
@@ -75,7 +75,7 @@
     
 - name: "Applying variables"
   replace:
-    path: "{{ repodir.path }}/controller.yml"
+    path: "{{ repodir.path }}/controller-4.yml"
```
stencell: I only see a `controller.yml` in the master & rhte2019 branches of the repo you are cloning...are you sure you want to change this here to reference a file that doesn't currently exist?

pranavgaikwad: @stencell Thank you for pointing that out. But, we already have a PR which will eventually make that file available at that location.

PR #636, opened 2019-08-22
Adding openshift-applier workload

oybed:
##### SUMMARY

This PR adds support for utilizing [openshift-applier](https://github.com/redhat-cop/openshift-applier) as a workload as part of the various configs - e.g.: part of the `default_workloads`. 

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME

ocp-workload-openshift-applier

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/ocp-workload-openshift-applier/readme.adoc
+++ b/ansible/roles/ocp-workload-openshift-applier/readme.adoc
@@ -1,0 +1,71 @@
+= ocp-workload-openshift-applier - execute an openshift-applier run 
+
+== Role overview
+
+* This role uses the [openshift-applier](https://github.com/redhat-cop/openshift-applier)
+ to manage content within an OpenShift Container Platform. It consists of the following playbooks:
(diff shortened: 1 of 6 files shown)


wkulhanek:
Where does the OpenShift applied role that is being included come from?

oybed:
@wkulhanek yeah, good point. It will require a `requirements.yml` in `ansible/configs` to make galaxy pull in the dependency. However, not sure how to document this without creating a new entry in `ansible/configs`. Any suggestions? Maybe just add the info to the role's `readme.adoc`? 

Sample content for `requirements.yml`:
```
---

- src: https://github.com/redhat-cop/openshift-applier
  name: openshift-applier
  version: v2.1.1

```

wkulhanek:
So I'll merge. :-) But it would be good to have the above in the README.

What's the use case for this? Run applier on top of a vanilla installed cluster? Or something else?

PR #685, opened 2019-08-31
updates workflows workshop to work with changes to other roles

willb:
##### SUMMARY

Fixes the ML Workflows Workshop to work with recent changes to the analytics/ODH workshop roles it depends on.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp4-workload-ml-workflows-workshop

##### ADDITIONAL INFORMATION

This includes fixes to the ML Workflows Workshop to work with recent changes to the analytics/ODH workshop roles it depends on.

NB:  all roles must be run with the following variable settings:  ```-e"jupyter_notebook_memory=6Gi" -e"deploy_odh_cr=true" -e"rgw_service_ip=rook-ceph-rgw-my-store" -e"rgw_service_port=8000"```  Since we don't control the ODH roles, we can't set them there.

### DIFF
--- a/ansible/roles/ocp4-workload-ml-workflows-workshop/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-ml-workflows-workshop/defaults/main.yml
@@ -1,4 +1,4 @@
 ---
 become_override: False
-ocp_username: wibenton-redhat.com
+ocp_username: opentlc-mgr
 silent: False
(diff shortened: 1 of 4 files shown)


fridim:
> NB: all roles must be run with the following variable settings: -e"jupyter_notebook_memory=6Gi" -e"deploy_odh_cr=true" -e"rgw_service_ip=rook-ceph-rgw-my-store" -e"rgw_service_port=8000" Since we don't control the ODH roles, we can't set them there.

Can you explain this part, i'm not sure to understand.
I don't see those variables in your role

willb:
Right!  For RHTE, this role depends on other roles (both infra and student) that need these variables to be set. (This lab reuses infrastructure from another lab.) Because those variables shouldn’t be set in the parent roles (they should be something else for the other lab), and because this role runs last, we need to set these variables while deploying all the roles. 

Let me know if this doesn’t make sense or if there’s a better way to accomplish this!

PR #702, opened 2019-09-11
pipeline waits for ODH operator deploy in S2020

willb:
##### SUMMARY

This commit both fixes long deploy times in S2020 and sets variables in the ODH role that are specific to S2020.  It improves deploy times by pipelining waits for the ODH operator to be available (looping over all of the tasks before waiting for each user before looping over waits for each user and then looping over post-wait tasks for each user).

It is necessary to pipeline rather than provision each user in parallel because Ansible [doesn't support running include files asynchronously](https://github.com/ansible/ansible/issues/22716).

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

NB:  After merging this commit, the student workload for S2020 should change to `ocp4-workload-rhte-analytics_data_ocp_workshop_s2020` instead of `ocp4-workload-rhte-analytics_data_ocp_workshop`.

### DIFF
--- a/ansible/roles/ocp4-workload-ml-workflows-workshop/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-ml-workflows-workshop/defaults/main.yml
@@ -2,3 +2,7 @@
 become_override: False
 ocp_username: opentlc-mgr
 silent: False
+
+user_count_start: 1
+# The last user number to start with when creating projects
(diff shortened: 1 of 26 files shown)


willb:
(I'll submit another PR applying this technique to S1003)

PR #703, opened 2019-09-11
pipeline waits for ODH CSV and operator across all users in S1003

willb:
##### SUMMARY

This commit improves deploy times in S1003 by pipelining waits for the ODH CSV and operator pod to be available (looping over all of the pre-wait tasks for each user, looping over waits for each user, and then looping over post-wait tasks for each user).

It is necessary to pipeline rather than provision each user in parallel because Ansible [doesn't support running include files asynchronously](https://github.com/ansible/ansible/issues/22716).

(See also #702) 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- `ocp4-workload-rhte-analytics_data_ocp_workshop`


willb:
@LaVLaS ptal

LaVLaS:
Just tested and it works great. Thx @willb

PR #730, opened 2019-09-23
Move Workloads before cluster shutdown enabled

wkulhanek:
##### SUMMARY

Move calling of workloads to before the cluster shutdown enablement logic.

That way the workloads should not be affected by rotating cluster operators (like API Server operators)

##### COMPONENT NAME

ocp4_workshop


wkulhanek:
@fridim Can we merge and tag this, please?

PR #733, opened 2019-09-25
ocp-workshop : Support for enabling cloud provider

pranavgaikwad:
##### SUMMARY
Adds support for enabling AWS cloud provider on OCP3 clusters. 

The default `ocp-workshop` uses custom DNS as OpenShift node names. When AWS cloud provider is installed, the custom DNS names do not work. We need a way to tell AgnosticD to use the default EC2 DNS names for access within the VPC. The external DNS names are not a problem. 

**Changes :** 
1. Do not set hostnames on Linux hosts when AWS cloud provider is to be installed. Use the default EC2 hostnames. 
2. Add `kubernetes.io/cluster/{{ guid }} = owned` tags to all EC2 instances, at least one security group directly applied to all nodes, subnet of those security groups.
3. Add Cloud Provider configuration in OpenShift Ansible hosts file.
4. In the inventory, use EC2 dns addresses for all nodes instead of the route 53 addresses. 

**About deletion:** 
Cloud Provider creates resources in the existing cloud formation stack which need to be deleted before attempting to delete CloudFormation template.

The changes are backward compatible

##### ISSUE TYPE

- Feature Pull Request (config)

##### COMPONENT NAME
ocp-workshop

##### ADDITIONAL INFORMATION

```paste below

```
cc: @jwmatthews

### DIFF
--- a/ansible/cloud_providers/ec2_infrastructure_deployment.yml
+++ b/ansible/cloud_providers/ec2_infrastructure_deployment.yml
@@ -95,6 +95,7 @@
     - name: Run infra-ec2-linux-set-hostname Role
       import_role:
         name: infra-ec2-linux-set-hostname
+      when: not install_cloud_provider  # ec2 cloud provider uses default hostnames
 
 - name: Step 001.4 Configure Windows Hosts and Wait for Connection
   gather_facts: false
--- a/ansible/configs/ocp-workshop/destroy_env.yml
+++ b/ansible/configs/ocp-workshop/destroy_env.yml
@@ -33,6 +33,29 @@
         - s3_result is not succeeded
         - cloud_provider == 'ec2'
 
+    - name: Delete ec2 cloud provider created resources
+      environment:
+        AWS_ACCESS_KEY_ID: "{{ aws_access_key_id }}"
+        AWS_SECRET_ACCESS_KEY: "{{ aws_secret_access_key }}"
+        AWS_DEFAULT_REGION: "{{ aws_region_final|d(aws_region) }}"
+      block:
+      - name: Finding ELBs
+        action:
+          module: ec2_elb_facts
+        register: elbs
+
+      - set_fact:
+          elb_names: "{{ elbs | json_query(query) }}"
+        vars:
+          query: "elbs[?tags.\"kubernetes.io/cluster/{{ guid }}\" == 'owned'].name"
+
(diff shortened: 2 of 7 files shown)


wkulhanek:
@fridim I'm not touching this. Feel free to merge if you feel comfortable.

ansible/cloud_providers/ec2_infrastructure_deployment.yml
```diff
@@ -95,6 +95,7 @@
     - name: Run infra-ec2-linux-set-hostname Role
       import_role:
         name: infra-ec2-linux-set-hostname
+      when: not install_cloud_provider  # ec2 cloud provider uses default hostnames
```
fridim: You need to specify a default for this variable.

```yaml
install_cloud_provider | default(false)
```

This playbook is common to all configs, but you define `install_cloud_provider` only in ocp-workshop `env_vars.yml`.

Also, this variable only means something in the context of ocp-workshop config.
It should be done the other way around, in the ocp-workshop config:

- if install_cloud_provider is `true` then set `infra_set_hostname` to `false`. Or simply, in `env_vars.yml`:

```yaml
install_cloud_provider: false
infra_set_hostname: "{{ not install_cloud_provider }}"
```
And replace the test in `ec2_inrastructure_deployment.yml` with:

```yaml
when: infra_set_hostname | default(true)
```

ansible/configs/ocp-workshop/destroy_env.yml
```diff
@@ -33,6 +33,29 @@
...
+        vars:
+          query: "elbs[?tags.\"kubernetes.io/cluster/{{ guid }}\" == 'owned'].name"
+
+      - name: Deleting ELBs provisioned by cloud provider
+        ec2_elb_lb:
+          name: "{{ item }}"
+          state: absent
+        loop: "{{ elb_names }}"
+      when: cloud_provider == 'ec2' and install_cloud_provider|bool
+
```
fridim: what about other kind of resources ? Are you sure this will cleanup **everything** ?

persistent volumes as EBS ?

pranavgaikwad: @fridim That's a good point. I have not thought about EBS volumes...I'll investigate more...

pranavgaikwad:
Closing this in favor of new config PR opened here : https://github.com/redhat-cop/agnosticd/pull/808

PR #738, opened 2019-10-02
Ocp workloads with openshift applier

fridim:
##### SUMMARY

Patch ocp-workloads config to work with openshift-applier.

- add openshift-applier in `requirements.yml`
- add `become_override` to `env_vars.yml`  (default false)

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/ansible/configs/ocp-workloads/examples/cakephp-openshift-applier.yml
+++ b/ansible/configs/ocp-workloads/examples/cakephp-openshift-applier.yml
@@ -1,0 +1,36 @@
+---
+cloud_provider: none
+env_type: ocp-workloads
+
+ocp_workloads:
+  - ocp-workload-openshift-applier
+
+ocp_username: gucore-redhat.com
+guid: testgucore
+
+openshift_cluster_content:
+
+  - object: projectrequest
+    content:
+    - name: Create namespace
+      template: https://raw.githubusercontent.com/redhat-cop/openshift-applier/v2.1.1/tests/files/templates/projectrequest.yml
+      params_from_vars:
+        NAMESPACE: "{{ guid }}"
+        NAMESPACE_DESCRIPTION: Sample namespace for student {{ ocp_username }}
+        NAMESPACE_DISPLAY_NAME: Sample namespace for student {{ ocp_username }}
+
+  - object: cakephprequest
+    content:
+    - name: Create cake-php mysql
+      template: https://raw.githubusercontent.com/openshift/origin/v4.1.0/examples/quickstarts/cakephp-mysql.json
+      params_from_vars:
+        NAME: "cakephp-mysql-{{ guid }}"
+      namespace: "{{ guid }}"
+
+target_host:
+  ansible_host: bastion.dev.openshift.opentlc.com
+  ansible_port: 22
+  ansible_user: ec2-user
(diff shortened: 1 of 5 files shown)


ansible/configs/ocp-workloads/examples/cakephp-openshift-applier.yml
```diff
@@ -0,0 +1,38 @@
...
+      namespace: "{{ guid }}"
+      action: "{{ applier_action }}"
+
+target_host:
+  ansible_host: bastion.dev.openshift.opentlc.com
+  ansible_port: 22
+  ansible_user: ec2-user
+  #ansible_ssh_private_key_content: "{{ target_host_ssh_private_key_content }}"
+  ansible_ssh_private_key_file: ~/.ssh/opentlc_admin_backdoor.pem
+  #ansible_ssh_extra_args:
```
jkupferer: Minor point, but the list indenting is inconsistent in this file. I prefer not adding the extra indent before lists, but at least it should be consistent.

ansible/configs/ocp-workloads/examples/cakephp-openshift-applier.yml
```diff
@@ -0,0 +1,38 @@
...
+        NAME: "cakephp-mysql-{{ guid }}"
+      namespace: "{{ guid }}"
+      action: "{{ applier_action }}"
+
+target_host:
+  ansible_host: bastion.dev.openshift.opentlc.com
+  ansible_port: 22
+  ansible_user: ec2-user
+  #ansible_ssh_private_key_content: "{{ target_host_ssh_private_key_content }}"
+  ansible_ssh_private_key_file: ~/.ssh/opentlc_admin_backdoor.pem
```
jkupferer: This is not going to work on tower, right? Unless we copy the key to every tower node?
Do you want to merge as-is and then solve the SSH key issue in another place?

fridim: private key *file* won't but private key *content* will. So we will use something like:

```yaml
ansible_ssh_private_key_content: "{{ target_host_ssh_private_key_content }}"
```

And set the variable in a secret.

In the ocp-workloads config, there is a step that create the key in output_dir, and use it to connect to the bastion. Then it's cleaned up.

ansible/configs/ocp-workloads/examples/cakephp-openshift-applier.yml
fridim: ok, i just pushed a change. I prefer using extra-indent. It's also the default in yamllint IIRC.

oybed:
Just curious - with the introduction of the requirements.yml file, it will now always pull in the openshift-applier - even when not intended to be used. Are you guys good with this? I intentionally didn't do this in the original PR to avoid the extra overhead, but maybe you see it useful enough to actual always pull it in?

fridim:
i think it's worth it yes, and i'm pretty sure it's not even noticeable.

Plus, if the overhead is an issue, we can rename  `requirements.yml` and set a variable to point to it, only in catalog items that need it:

```yaml
requirements_path: configs/ocp-workloads/requirements-openshift-applier.yml
```

(already implemented, see https://github.com/redhat-cop/agnosticd/blob/development/ansible/install_galaxy_roles.yml#L13 )

PR #783, opened 2019-10-28
Removed a bad ref to nonexistent velero workload

jwmatthews:
Cleaned up doc to remove a bad reference
cc @newgoliath


newgoliath:
LGTM

PR #806, opened 2019-11-07
Hardcode python2 openshift

wkulhanek:


ansible/configs/ocp4-workshop/software.yml
```diff
@@ -51,22 +51,25 @@
...
-          become: yes
-          package:
-            name:
+            - https://gpte-public.s3.amazonaws.com/python2-pip-8.1.2-9.el7.noarch.rpm
             - golang
             - unzip
 
+        - name: Install Python Packages
+          become: yes
+          pip:
```
stencell: Wouldn't it be better to pin these to a specific versions?

wkulhanek: Well I couldn't use the ones that I had saved from EPEL - they were too old and AWS "stuff" was failing during the OpenShift install.

I could pin those again to the versions installed. But when AWS updates "things" and updates the packages we'd be left behind and would have breakage again.

stencell: The alternative is a broken or "enhanced" library could be pushed to PyPI that doesn't work with our code. You are the primary owner of this config, so if you are comfortable essentially using 'latest' for these libraries, then ok. I guess there is challenges both ways, although I typically prefer to pin versions and deal with an update if the break backwards compatibility.

Let me know if you want to proceed as is.

PR #834, opened 2019-11-13
Workload to deploy gogs

jorgemoralespou:
##### SUMMARY
Workload to deploy gogs. This workload is used in OpenShift BU workshops (for now) and it's meant to replace the role `https://github.com/siamaksade/ansible-openshift-gogs` used in the current APB

### DIFF
--- a/ansible/roles/ocp-workload-gogs/tasks/workload.yml
+++ b/ansible/roles/ocp-workload-gogs/tasks/workload.yml
@@ -1,0 +1,89 @@
+---
+# Implement your Workload deployment tasks here
+
+- name: Setting up workload for user
+  debug:
+    msg: "Setting up workload for user ocp_username = {{ ocp_username }}"
+
+- name: Create Project {{ gogs_project }}
+  k8s:
+    state: present
+    definition: "{{ lookup('template', item ) | from_yaml }}"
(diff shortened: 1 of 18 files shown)


ansible/roles/ocp-workload-gogs/tasks/workload.yml
```diff
@@ -0,0 +1,89 @@
...
+  k8s:
+    state: present
+    definition: "{{ lookup('template', item ) | from_yaml }}"
+  loop:
+  - ./templates/project.j2
+  register: r_createproject
+  until: r_createproject is succeeded
+  retries: 5
+
+- name: Get apps route suffix
```
wkulhanek: I just submitted a PR #836 for the ocp-workload-terminal role to change that logic since it keeps failing when deleting the dummy route for some reason. You don't actually need that dummy route. You can just create the real route and then retrieve it.

If you only need to support OCP 4 there is an even easier way:
https://github.com/redhat-gpte-devopsautomation/ansible-operator-roles/blob/master/roles/gogs-ocp/tasks/main.yml#L16

jorgemoralespou: Changed

PR #865, opened 2019-11-21
[WIP] OSP cloud provider refine

fridim:
##### SUMMARY

This PR contains my changes after trying the OSP cloud-provider roles and playbooks

- three-tier-app
- ocp-clientvm

Try the actions:

- provision
- destroy
- stop
- start
- status

##### ISSUE TYPE

- Bugfix Pull Request


ansible/lifecycle.yml
```diff
@@ -68,4 +75,74 @@
...
+
+            - when: r_osp_facts.openstack_servers | length > 0
+              block:
+                - set_fact:
+                    all_instances: >-
+                      {{ r_osp_facts.openstack_servers
+                      | json_query('[*].id') }}
+
+                - name: Stop all servers
+                  command: nova stop {{ all_instances | join(' ') }}
```
stencell: We should try to use the `openstack` client when possible, unless it is lacking a piece of functionality. This is core functionality that is available.

ansible/roles/infra-osp-template-generate/defaults/main.yaml
```diff
@@ -8,3 +8,71 @@ opentlc_admin_pub_keys:
...
+
+      - name: SSHPublic
+        description: "Public Access for bastions"
+        from_port: 22
+        to_port: 22
+        protocol: tcp
+        cidr: "0.0.0.0/0"
+        rule_type: Ingress
+
+  - name: DefaultSG
```
stencell: Instead of creating another SG, why don't we just use the default one that comes automatically when we provision a new project & network? We'd just need to add a task that changes the inbound from being limited to a remote SG to 0.0.0.0/0

ansible/roles/infra-osp-template-generate/defaults/main.yaml
```diff
@@ -8,3 +8,71 @@ opentlc_admin_pub_keys:
...
 
 provider_network: external
+
+default_security_groups:
+  - name: BastionSG
+    description: >-
+      Bastion security group allows basic ICMP
+      and SSH adn Mosh ingress and egress to *
+    rules:
+      - name: MoshPublic
```
stencell: We do not need this in a default. I'd suggest most people don't even know what mosh is or use it.

tools/builds/Dockerfile-openstack
```diff
@@ -18,10 +19,14 @@ RUN mkdir -p /etc/ansible \
...
 RUN pip3 install --no-cache \
-    ansible \
-    openstacksdk \
     python-heatclient \
     python-openstackclient \
+    python-neutronclient \
+    python-keystoneclient \
+    python-novaclient \
+    python-glanceclient \
+    ansible \
```
stencell: Is this Dockerfile going to be used for actual deployments or just testing? If actual, we should pin to specific versions in pip. Some module names change for OpenStack between Ansible 2.9 and earlier.

ansible/configs/three-tier-app/env_vars.yml
```diff
@@ -99,15 +99,23 @@ appdb_instance_count: 1
 support_instance_count: 1
 rootfs_size_bastion: 50
 
+bastion_instance_image: RHELAMI
+frontend_instance_image: RHELAMI
+app_instance_image: RHELAMI
+appdb_instance_image: RHELAMI
+support_instance_image: RHELAMI
```
stencell: These seem very specific to EC2. Are people expected to change these vars either here or in agnosticV? If so, can we add some comments to explain why they are set like this?

ansible/lifecycle.yml
```diff
@@ -68,4 +75,74 @@
...
+        OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
+        OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
+      block:
+        - when: ACTION == 'stop'
+          block:
+            - name: Gather instance facts
+              os_server_info:
+                filters:
+                  metadata:
+                    Project: "{{ project_tag }}"
```
stencell: I think this key currently gets set lowercase when instances are created. In case it is case sensitive, either change the filter here or we need to change the meta we add to match.

I prefer all lowercase to avoid issues like this.

ansible/lifecycle.yml
```diff
@@ -68,4 +75,74 @@
...
+
+            - when: r_osp_facts.openstack_servers | length > 0
+              block:
+                - set_fact:
+                    all_instances: >-
+                      {{ r_osp_facts.openstack_servers
+                      | json_query('[*].id') }}
 
+                - name: Start all servers
+                  command: nova start {{ all_instances | join(' ') }}
```
stencell: Please consider using the openstack client here instead of nova.

ansible/roles/infra-osp-template-generate/tasks/locate_template.yml
```diff
@@ -21,16 +21,15 @@
...
   when: stat_nested_template.stat.exists
 
-# Until there is a default template, leave this disabled and just fail instead
-# - name: Use the default infra template
-#   set_fact:
-#     cloud_infra_template_src: "templates/cloud_template.j2"
-#   when: not stat_local_template.stat.exists
+- name: Use the default infra template
+  set_fact:
+    cloud_infra_master_template_src: "templates/cloud_template_master.j2"
```
stencell: I'm not seeing this folder or file in the role directory?

ansible/roles/infra-osp-template-generate/tasks/locate_template.yml
```diff
@@ -21,16 +21,15 @@
...
+  when: not stat_master_template.stat.exists
 
-# - name: Fail if infra template not in config
-#   fail:
-#     msg: "No infra template found in ../configs/{{ env_type }}/files/cloud_providers/ directory"
-#   when: not stat_local_template.stat.exists
+- name: Use the default infra template
+  set_fact:
+    cloud_infra_nested_template_src: "templates/cloud_template_nested.j2"
```
stencell: I'm not seeing this folder or file in the role directory?

ansible/roles/infra-osp-template-generate/defaults/main.yaml
```diff
@@ -8,3 +8,71 @@ opentlc_admin_pub_keys:
 all_ssh_authorized_keys: "{{ opentlc_admin_pub_keys + [user_pub_key|d()] }}"
 
 provider_network: external
+
+default_security_groups:
```
stencell: I'm confused how this is being used. In place of SGs defined for a config? Merged with it? Did you make changes to the Heat templates for this?

tools/builds/Dockerfile-openstack
fridim: It's currently only used to deploy from the laptop.

I want us to start populating docker images, and have something mature for the day we need them.

Like mentioned by @wkulhanek we could use UBI here

ansible/roles/infra-osp-template-generate/defaults/main.yaml
fridim: We have it in default for aws, and i kind of like Mosh.

Does it hurt ?

ansible/roles/infra-osp-template-generate/defaults/main.yaml
fridim: That way it's explicite in agnosticd ?

ansible/lifecycle.yml
fridim: Thanks, will change

ansible/lifecycle.yml
fridim: In AWS tags, it's `Project` :/

I'll change it to use `guid` and `env_type` like i did later.

ansible/lifecycle.yml
fridim: Will do, thanks

ansible/roles/infra-osp-template-generate/tasks/locate_template.yml
fridim: Because i forgot to push. Done.

ansible/roles/infra-osp-template-generate/defaults/main.yaml
fridim: yep, pushed now

ansible/configs/three-tier-app/env_vars.yml
fridim: Yes they are expected to change it.

We can provide meaning full examples (but idk if openstack comes with images by default so it may just be specific anyway).

see `sample_vars_osp.yml`

ansible/roles/infra-osp-template-generate/defaults/main.yaml
stencell: What exactly does that buy you? is there a reason that having it explicit in agnosticd is useful? 

You have a `default` SG 100% of the time in OpenStack. This would just create another `DefaultSG` SG that is intended to do the same thing. Less is more IMO, but if you really feel another SG is warranted, I am not going to stop it.

fridim: i'm not a big fan of adding ansible tasks to just add rules in the default SG.
Here it's very self-documenting, and it's just configuration.
No more code i mean.

Unless we can just rename DefaultSG with 'default' and the heat template will update the default SG without complaining ?
I can try.

ansible/roles/bastion-student-user/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
 
 - name: Generate student_password if not defined
   set_fact:
-    student_password: "{{ lookup('pipe', 'openssl rand -base64 12') }}"
+    student_password: "{{ lookup('password', '/dev/null length=12 chars=ascii_letters,digits') }}"
```
wkulhanek: Great minds.... I've been meaning to make this exact change for a while now...

ansible/roles/infra-osp-template-generate/defaults/main.yaml
stencell: Per our conversation...

Does it hurt? No. Is it needed? No. These are defaults. Defaults should be needed either for functionality of a config/role or by a lab. Nothing requires this, so it should not be included going forward just because it was included in the past.

If there is a guaranteed repository for this package - or any like it - that is consistently maintained for updates, security fixes, etc. I'd be more okay with it. In the past we used EPEL, which can break things easily. Recently, we started hosting some packages from an s3 bucket, but I don't know who is ensuring that stays up to date.

ansible/configs/ocp-clientvm/env_vars.yml
```diff
@@ -177,3 +180,31 @@ instances:
       - device_name: "{{docker_device}}"
```
stencell: Can we consider dropping this extra device and making the size of the root disk like 30GB? We don't need this much storage on a client VM for anything and we have finite amounts of storage available.

ansible/configs/ocp-clientvm/env_vars.yml
```diff
@@ -177,3 +180,31 @@ instances:
...
+
+#### OSP ####
+# See cloud_providers/osp_default_vars.yml
+# See roles/infra-osp-project-create/defaults/main.yml
+
+# Set this to true if you need to create a new project in OpenStack
+# This should almost always be set to true for OpenShift installations
+# If it is set to false, the {{ osp_project_name }} must already exist and
+# should be able to run whatever you are deploying
+osp_project_create: true
```
stencell: I think this is still something to be discussed. 

Why don't we create a common project for client VMs. There is no reason that I can think of to have a separate project for each. It is just going to be a single VM with a single disk deployed on a single (existing) network with a single FIP.

ansible/roles/infra-osp-resources-destroy/tasks/delete_osp_resources.yml
```diff
@@ -30,19 +30,36 @@
...
-            | xargs openstack volume delete
-        
-        - name: Delete any trunk ports in project
-          shell: >-
-            set -o pipefail;
-            openstack port list --project {{ osp_project_name }} -f json -c trunk_details
-            | jq '.[].trunk_details.trunk_id | select (.!=null)'
-            | xargs openstack network trunk delete
-        
+        - name: Get all remaining volumes in project
```
stencell: I'm really confused why we turned 12 lines of code with 2 commands into over double the amount for this? This seems like overcomplicating what can be done much simpler. If there are volumes or trunks to delete, it deletes them. If not, it moves on to the remainder.

is there a reason we need the information collected, added as facts, and then essentially running a very similar delete command?

ansible/roles/infra-osp-template-generate/defaults/main.yaml
fridim: I just removed it from the default security groups.

ansible/configs/ocp-clientvm/env_vars.yml
fridim: I'm happy to reduce this, but i think it's out of the scope of this PR.

i created: https://github.com/redhat-cop/agnosticd/issues/882

ansible/configs/ocp-clientvm/env_vars.yml
fridim: I don't know if that's how we will do.
I like the separation per project. TBD with our OSP team.

But if we choose to use one project for a kind of deployment, when deploying from CloudForms + agnosticV we can hardcode the project name:

```yaml
osp_project_name: LABS-clientvms
admin_user: opentlc-mgr
# set quotas globally for this catalog item  
quota_num_instances: 200
...
```

We don't even have to set `osp_project_create: false`, since it should be idem potent.

From an agnosticD user/dev point of view, it makes things easier since everything is done by ansible. I like to minimize the amount of manual things to do.

If by default we set `osp_project_create` to false, then we have to tell in the documentation to login in OpenStack, create a project, etc...

ansible/roles/infra-osp-resources-destroy/tasks/delete_osp_resources.yml
fridim: I got errors when trying ocp-clientvm with that "shell" tasks.

The first problem was that when no volume was found, the task was failing

```
TASK [infra-osp-resources-destroy : Delete any remaining volumes in project] ******************************************************************************************************************
Friday 22 November 2019  07:53:42 +0000 (0:00:12.699)       0:00:24.214 ******* 
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "set -o pipefail; openstack volume list --project guid01-project -f value -c ID | xargs openstack volume delete", "delta": "0:00:05.741398", "end": "2019-11-22 07:53:48.226543", "msg": "non-zero return code", "rc": 123, "start": "2019-11-22 07:53:42.485145", "stderr": "usage: openstack volume delete [-h] [--force | --purge]\n                               <volume> [<volume> ...]\nopenstack volume delete: error: the following arguments are required: <volume>", "stderr_lines": ["usage: openstack volume delete [-h] [--force | --purge]", "                               <volume> [<volume> ...]", "openstack volume delete: error: the following arguments are required: <volume>"], "stdout": "", "stdout_lines": []}
```


The second error was:

```

TASK [infra-osp-resources-destroy : Delete any trunk ports in project] ************************************************************************************************************************
Friday 22 November 2019  12:30:46 +0000 (0:00:00.021)       0:00:19.720 ******* 
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "set -o pipefail; openstack port list --project guid01-project -f json -c trunk_details | jq '.[].trunk_details.trunk_id | select (.!=null)' | xargs openstack network trunk delete", "delta": "0:00:03.332714", "end": "2019-11-22 12:30:50.296071", "msg": "non-zero return code", "rc": 123, "start": "2019-11-22 12:30:46.963357", "stderr": "/bin/sh: jq: not found\nusage: openstack network trunk delete [-h] <trunk> [<trunk> ...]\nopenstack network trunk delete: error: the following arguments are required: <trunk>\nException ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>\nBrokenPipeError: [Errno 32] Broken pipe", "stderr_lines": ["/bin/sh: jq: not found", "usage: openstack network trunk delete [-h] <trunk> [<trunk> ...]", "openstack network trunk delete: error: the following arguments are required: <trunk>", "Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>", "BrokenPipeError: [Errno 32] Broken pipe"], "stdout": "", "stdout_lines": []}
```

`jq` binary was not available in my docker images.  I can easily add it, but it's also a new dependency for anybody deploying from their laptop since this is running from localhost.

It's difficult to think of everything when using a `shell` task, spliting into separate tasks is good practice.

You noticed i added:

```bash
set -o pipefail
```

Tells the shell to fail if any command from the pipe fails.
we want to know if `openstack volume list` failed. Because if it fails silently, then we can't be sure we deleted everything.

But you have to be sure of every return code of every commands in the pipe.
And they are not always consistent.

On the other hand, you don't necessary want to remove `set -o pipefail`, you want to know if you "read" command failed. So for something apparently very easy, it quickly become a puzzle.

When you split the commands into separate tasks, it gives you flexibility.

You can even then start defining:

- `failed_when`  (using rc and what not)
- `changed_when` (read-only tasks should have `changed_when: false`
- rescue block... 
- etc

It's more "lines of code" but it's more robust and flexible.

ansible/configs/ocp-clientvm/env_vars.yml
fridim: Actually, since deletion is impacted by this choice, i made some changes:
- always delete heat stack  (first step)
- delete resources of the project only if `osp_project_create` is `true`.

@stencell let me know if you can think of any problem when deleting the heat stack first.

and let me update my previous comment:

> But if we choose to use one project for a kind of deployment, when deploying from CloudForms + agnosticV we can hardcode the project name:
```yaml
osp_create_project: false
osp_project_name: LABS-clientvms
admin_user: opentlc-mgr
# set quotas globally for this catalog item  
quota_num_instances: 200
```
> We **have to set** `osp_project_create: false`, because when deleting, we don't want to delete everything in the project.

stencell: > Actually, since deletion is impacted by this choice, i made some changes:
> 
> * always delete heat stack  (first step)
> * delete resources of the project only if `osp_project_create` is `true`.
> 
> @stencell let me know if you can think of any problem when deleting the heat stack first.
> 

I think this will cause a problem. The heat stack creates resources like network and router. Over the course of the lifetime of a project, users may create instances, ports, etc. that are outside of the heat stack. If you try to tear the heat stack down without deleting those extra ports, for instance, the heat stack teardown will fail.

This is why i originally did the purge(s) first. Then deleting the heat stack cleaned up the remainder that was originally created. Deleting the project always comes at the end regardless of the order of the previous two.

@fridim

ansible/roles/infra-osp-resources-destroy/tasks/delete_osp_resources.yml
stencell: We'll have to agree to disagree on some of this since I think there is a happy medium between simplicity and robustness, but I accept your changes and arguments to make this work as needed (even if adding a little complexity!).

ansible/roles/infra-osp-resources-destroy/tasks/main.yml
```diff
@@ -1,2 +1,15 @@
 ---
-- import_tasks: delete_osp_resources.yml
\ No newline at end of file
+- import_tasks: detect_project.yml
+
+- when: osp_project_info | length > 0
```
stencell: As mentioned in other comment, heat stack deletion needs to come after `project.yml` or the heat stack will sometimes fail to delete due to dependencies created outside heat stack.

ansible/roles/infra-osp-template-generate/defaults/main.yaml
stencell: Did you get a chance to try with this? If not, I suggest we move forward as is. While we don't need another SG, I don't think it will hurt at this point and with the new purge tasks, anything created with our outside of Heat will get cleaned up prior to project deletion.

It is always something we can iterate on later since it would be "more right" to just use the default SG that already exists.

ansible/configs/ocp-clientvm/env_vars.yml
fridim: OK will change

ansible/roles/infra-osp-template-generate/defaults/main.yaml
fridim: I haven't tried it yet.
I agree, let's move this task outside of this PR.

PR #867, opened 2019-11-21
allow setting project limits

newgoliath:
##### SUMMARY
users can now set the limits they want in their project-request-template

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4-workload-project-request-template

##### ADDITIONAL INFORMATION

NOT TESTED - PLEASE TEST! :-)


ansible/roles/ocp4-workload-project-request-template/templates/project_request_template.j2
```diff
@@ -11,21 +11,7 @@ objects:
...
-        cpu: 500m
-        memory: 1.5Gi
-      defaultRequest:
-        cpu: 50m
-        memory: 256Mi
-    - type: Pod
-      max:
-        cpu: 2
-        memory: 12Gi
+ {{ project_limits | to_json }}
```
fridim: I would write:

```yaml
spec: {{ project_limits | to_json }}
```

stencell:
@newgoliath Do you want to put any changes in based on GC feedback? close this?

fridim:
This is a good idea, it just needs testing and the proper syntax which is, i think, 

```yaml
  spec: {{ project_limits | to_json }}
```

not 

```yaml
  spec:
 {{ project_limits | to_json }}
```

wkulhanek:
@newgoliath Is this still necessary?

newgoliath:
Necessary?  I thought it would be good to parameterize the project requests and limits so different workloads could set their own with a minimum of fuss.

fridim:
It's not a bad idea, but it would require to fix the syntax as suggested, and testing :)

Also now there are conflicts with current `development` branch

fridim:
I made some changes to make this happen.

wkulhanek:
Great. That’ll break a number of configs now.

wkulhanek:
Never mind. This is the old role. As long as the defaults are the same it should be OK.

PR #883, opened 2019-11-25
[WIP] Updated Quay Operator to 0.0.7 and quay to 3.0.5

wkulhanek:


stencell:
We should wait to merge this, or wait to tag this, until the labs are verified with these new versions.

wkulhanek:
@stencell agreed. Especially since the operator doesn’t work with the current config..... hence the WIP label.

PR #896, opened 2019-12-02
ensure backwards compatibility ready for v1.6.x

aidenkeating:
##### SUMMARY
in v1.6.x of integreatly we want to use a new inventory file, however the
older releases of integreatly should still maintain using the older hosts
inventory file.

this change adds a version check to ig_version, if it is within the current
known legacy versions that are installable from pds, the older hosts
inventory will be used. Else, the new pds inventory will be used.

the aim of this is to ensure the removal of this logic will be simple once
it is no longer needed.

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME
Integreatly (RHMI)

@trepel @pmccarthy would you mind taking a look, i'm not familiar with how this repo can be tested from an intly perspective?

### DIFF
--- a/ansible/roles/ocp-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp-workload-integreatly/defaults/main.yml
@@ -3,7 +3,8 @@
 become_override: True
 silent: False
 install_dir: /tmp/integreatly
-inventory_hosts_file: inventories/hosts.template
+inventory_hosts_file_legacy: inventories/hosts.template
+inventory_hosts_file: inventories/pds.template
 release_tag: "release-{{ ig_version }}"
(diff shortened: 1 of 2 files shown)


aidenkeating:
@NonyNo3lle @fridim Mind taking a look? This will unblock the RHMI v1.6.0-rhpds release

trepel:
I managed to verify that pds.template is used for 1.6.0-rhpds version and hosts.template is used for 1.5.2 version of Integreatly Workshop using DEV catalog.

I think this can be tagged for TEST and PROD catalogs.

@rut31337 @fridim @NonyNo3lle please, have a look if you can spare some time. Thanks.

PR #897, opened 2019-12-02
Fix App Deploy Homework for Users with "_" in their user name

wkulhanek:
Some IBMers have a "_" in their user name. For example benoit_godin@be.ibm.com.

"_" is an invalid character for a Kubernetes resource - so creating a cluster resource quota including the user name fails. Fixed by removing the user name from the name of the quota.

### DIFF
--- a/ansible/roles/ocp4-workload-app-deploy-homework/templates/cluster_resource_quota.j2
+++ b/ansible/roles/ocp4-workload-app-deploy-homework/templates/cluster_resource_quota.j2
@@ -1,7 +1,7 @@
 apiVersion: quota.openshift.io/v1
 kind: ClusterResourceQuota
 metadata:
-  name: "clusterquota-{{ ocp_username }}-{{ guid }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/ocp4-workload-app-deploy-homework/templates/cluster_resource_quota.j2
```diff
@@ -1,7 +1,7 @@
 apiVersion: quota.openshift.io/v1
 kind: ClusterResourceQuota
 metadata:
-  name: "clusterquota-{{ ocp_username }}-{{ guid }}"
```
fridim: You probably want to keep the old way too, and make sure it's absent. (Or cleanup quotas manually).

Is it the only resource having the username in its name?

wkulhanek: If I do the remove will fail...

PR #927, opened 2019-12-10
Modidy RHSM playbook and fix issues for RHEL 8 based bastion

makentenza:
##### SUMMARY
The RSHM playbook for the `set-repositories` role had lot of issues and was not reliable. This is an effort to make it idempotent while using `repo_method: rhn` variable.

This has been tested several times in the Novello OSP environment with no errors.

At the same time I'm adding `ignore_errors: yes` to the Mosh Iptables task, to be aligned to the previous tasks for Mosh and others where errors where ignored if the packages are not installed. This was particularly failing in a vanilla RHEL 8.1 deployment.

##### ISSUE TYPE
Role not working properly

##### COMPONENT NAME
The `set-repositories` role

### DIFF
--- a/ansible/roles/set-repositories/tasks/rhn-repos.yml
+++ b/ansible/roles/set-repositories/tasks/rhn-repos.yml
@@ -1,41 +1,31 @@
 # vim: set ft=ansible:
 ---
-- name: Force unregister before register
-  redhat_subscription:
-    state: absent
-  register: task_result
-  until: task_result | succeeded
-  retries: 10
-  delay: 5
 
-- name: register node with subscription-manager
+- name: Force unregister before register
+  command: "{{ item }}"
+  args:
+    warn: False
+  with_items:
+  - 'subscription-manager clean'
+  - 'subscription-manager remove --all'
+  - 'yum remove -y "katello-ca-consumer-*"'
+
+- name: 'Register system using Red Hat Subscription Manager'
   redhat_subscription:
     state: present
     username: "{{ rhel_subscription_user }}"
     password: "{{ rhel_subscription_pass }}"
-    autosubscribe: false
-  register: task_result
-  until: task_result | succeeded
-  retries: 10
-  delay: 5
+    pool_ids: "{{ rhsm_pool_ids }}"
+    auto_attach: false
 
-# TODO: should take pool ids from a var
-- name: get product pool id
-  shell: /usr/bin/subscription-manager list --all --available --matches="*{{rhn_pool_id_string}}*" | awk '/Pool ID/ {print $3}' | head -1
-  # command: subscription-manager list --all --available --matches="OpenShift Container Platform" | awk '/Pool ID/ {print $3}' | head -1
-  register: pool_id
-  until: pool_id | succeeded
-  retries: 10
-  delay: 5
-
-- name: attach host to subscription pool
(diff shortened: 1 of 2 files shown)


ansible/roles/set-repositories/tasks/rhn-repos.yml
```diff
@@ -1,41 +1,31 @@
...
-  register: task_result
-  until: task_result.rc == 0
-  retries: 10
-  delay: 5
-
-- name: enable repos for rhel
-  command: subscription-manager repos --enable "{{ item }}"
+- name: "Build command line for repos to enable"
+  set_fact:
+    repos_params: "{{ repos_params|default('') }} --enable={{ item }}"
```
fridim: You should not use the same variable `repos_params`.

`set_fact` won't append the `--enable={{ item }}` if the `repos_params` is set via extra-vars.
When set as extra-vars, it takes precedence over `set_fact`.

makentenza: `repos_params` is only used during run-time for this playbook, is not meant to be used in extra-vars while using the role, the repositories related variable provided by the user is `rhel_repos`. Does it make sense?

fridim: OK!
I'll give a try and let you know!

makentenza: Great, as an example for the actual config I'm developing these are the RHSM related variables:

```repo_method: rhn
use_subscription_manager: true
rhel_subscription_user: "xxxx@redhat.com"
rhel_subscription_pass: "xxxxxxxx"
rhsm_pool_ids:
  - "xxxxxxxxxxxxxxxxxxxxxxxxxxxx"
rhel_repos:
   - "rhel-8-for-x86_64-baseos-rpms"
   - "rhel-8-for-x86_64-appstream-rpms"
   - "ansible-2-for-rhel-8-x86_64-rpms"```

File: ansible/roles/set-repositories/tasks/rhn-repos.yml [hunk 2]
```diff
@@ -1,41 +1,31 @@
...
   with_items:
-    - '{{ rhel_repos }}'
+  - "{{ rhel_repos }}"
+
+- name: "Run 'subscription-manager to disable/enable repos"
+  command: "subscription-manager repos {{ repos_params }}"
```
fridim: you could probably just do:

```yaml
- name: "Run 'subscription-manager to disable/enable repos"
   command: >-
    subscription-manager repos {% for repo in rhel_repos %}
    --enable={{ repo }}
    {% endfor %}
  when: rhel_repos | default([]) | length > 0
```

?

makentenza: Yes, indeed we can for this purpose as we are assuming this is a new deployment and actually unregister the system in case the image used was already registered. Just changed that part of the code.

I think this is ok for this purpose, but if you think this role is going to be used for more than registering systems during deployment phase we could extend it to cover more use cases, but maybe this is too much here.

Check these examples (as I'm using both as a base for this) to see what I'm referring to:

- https://github.com/redhat-sap/sap-rhsm/blob/master/tasks/main.yml
- https://github.com/redhat-cop/infra-ansible/blob/master/roles/rhsm/tasks/main.yml

fridim:
I'll try to test this today and if successful will merge it

PR #931, opened 2019-12-12
Modify Satellite role to get it working with the latest release

makentenza:
##### SUMMARY
Latest Satellite release is expecting different parameters from the one configured at the moment. Also the order on how the playbooks are executed needs to change as Satellite will prevent to install `firewalld` package if Satellite is installed before the package

##### ISSUE TYPE
The actual role is failing with 2 issues:

- `foreman-admin-username` and `foreman-admin-password` have been deprecated in favour of `foreman-initial-admin-username` and `foreman-initial-admin-password
- `firewalld` won't be installed due to packages been excluded by `foreman-protector`

##### COMPONENT NAME
Role `satellite-installation`

### DIFF
--- a/ansible/roles/satellite-installation/tasks/main.yml
+++ b/ansible/roles/satellite-installation/tasks/main.yml
@@ -1,3 +1,4 @@
 ---
-- import_tasks: satellite_installation.yml
 - import_tasks: firewalld.yml
+- import_tasks: satellite_installation.yml
+
--- a/ansible/roles/satellite-installation/tasks/satellite_installation.yml
+++ b/ansible/roles/satellite-installation/tasks/satellite_installation.yml
@@ -1,4 +1,9 @@
 ---
+
+- name: Ensure server hostname is set as FQDN hostname
+  hostname: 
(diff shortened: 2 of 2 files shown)


makentenza:
@miteshrh please have a look

miteshrh:
@makentenza Thank you for update, I have one doubt, Is this --foreman-initial-admin-username, --foreman-initial-admin-password option available for satellite version 6.4 ? if yes then we are good to go. 

FYI, This role was originally designed for 6.4 only.

makentenza:
@miteshrh right now the role will always install the latest Satellite version available since there is no version "locker" in the role, so for anyone using the role right now will fail.

https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/satellite-installation/tasks/satellite_installation.yml#L11-L26

miteshrh:
@makentenza, I updated existing satellite installation role with extendable capabilities. Let my PR get merge then you can check if it works else update the role and create new PR. Thanks

makentenza:
Closing as this is covered already by #946

PR #932, opened 2019-12-12
Fix issues with NFS and RHEL 8

makentenza:
##### SUMMARY
There are several changes happening in RHEL 8 regarding NFS Server service. As an example the following ones that prevent using this role as-is:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#the-nobody-user-replaces-nfsnobody_shells-and-command-line-tools

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#the-etc-sysconfig-nfs-file-and-legacy-nfs-service-names-are-no-longer-available_file-systems-and-storage

In order to evaluate the RHEL major version for some tasks, some conditionals have been added to configure the system accordingly to the corresponding RHEL release

##### ISSUE TYPE
Role not working in RHEL 8

##### COMPONENT NAME
Role host-ocp-nfs

### DIFF
--- a/ansible/roles/host-ocp-nfs/handlers/main.yml
+++ b/ansible/roles/host-ocp-nfs/handlers/main.yml
@@ -13,7 +13,7 @@
     - /var/named
     - /etc/named.conf
 
-- name: restart nfs services
+- name: restart nfs services 7
   service:
     name: '{{  item }}'
     enabled: true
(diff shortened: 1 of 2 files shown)


makentenza:
@wkulhanek @pranavgaikwad I think you guys maintain this role. Could you please have a look to this quick PR?

PR #940, opened 2019-12-17
Switch security group port specs to something OSP can use

rtweed:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New config Pull Request
- New role Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


ansible/configs/satellite-v64-prod/env_vars.yml
```diff
@@ -184,7 +183,6 @@ capsule_instances:
     count: "{{capsule_instance_count}}"
     security_groups:
       - CapsuleSG
-      - DefaultSG
```
fridim: Why remove the defaultSG ?

rtweed: When deploying to OSP, a duplicate rule configuration causes the deployment to fail - whereas AWS copes with this just fine. The long-term solution is to deduplicate the security group declarations, but as capsules aren't even used in this configuration at present then removing DefaultSG here allows the Satellite deployment to complete in both OSP and AWS.

fridim: OK if it works with both, all good

PR #941, opened 2019-12-17
Three tier app: upgrade ftl-injector version in config's requirements.yml

tonykay:
##### SUMMARY
Upgraded the version of ftl-injector used by config three-tier-app

##### ISSUE TYPE

- Feature Pull Request
- New config Pull Request
- New role Pull Request

##### COMPONENT NAME
three-tier-app is used as the test bed for FTL and changes are required for new FTL functionality in the **external** ftl-injector role

### DIFF
--- a/ansible/configs/three-tier-app/requirements.yml
+++ b/ansible/configs/three-tier-app/requirements.yml
@@ -3,4 +3,4 @@
 
 - src: https://github.com/redhat-gpte-devopsautomation/ftl-injector
   name: ftl-injector
-  version: v0.11
+  version: 0.14.0


ansible/configs/three-tier-app/requirements.yml
```diff
@@ -3,4 +3,4 @@
 
 - src: https://github.com/redhat-gpte-devopsautomation/ftl-injector
   name: ftl-injector
-  version: v0.11
+  version: 0.14.0
```
fridim: scheme changed ?
`v0.14.0` vs `0.14.0` ?

tonykay: Yes, trying to be consistent across our portfolio. So since it isn't in production yet now was a good time to change.

That said if we, as a team, want to prefix with v happy to go that way. Thoughts re how semantic versioning should be applied?

PR #962, opened 2020-01-06
SAP HANA initial AgnosticD configuration

makentenza:
##### SUMMARY
Add new configuration to AgnosticD. This configuration will be used for 'Techincal Enablement on how to insall SAP HANA on an standalone server using Ansible and/or Ansible Tower'

##### COMPONENT NAME
Config 'sap-hana-auto'

##### ADDITIONAL INFORMATION
This has been tested in the RED Cluster. Additional modifications may be required to accommodate BLUE configuration once existing blockers (RHEL 8 images, RHEL 8 Subscriptions) are removed, but this is ready to go.


ansible/configs/sap-hana-auto/README.adoc
```diff
@@ -0,0 +1,52 @@
+== Overview
+
+*SAP HANA Auto* _config_ will cover the Technical Enablement for 'RHEL for SAP Solutions' related trainning. This is the Techincal Enablement on how to insall SAP HANA on an standalone server using both Ansible and Ansible Tower
+
+It makes an ideal base infrastructure to build on and can easily be extended via it's `env_vars.yml` to less or more machines and also to different operating system images.
```
fridim: We now support `default_vars.yml`. Try to prefer `default_vars.yml` over `env_vars.yml` (same thing, just a different name)

if some variables are specific to a cloud provider, for example osp,  put them in `default_vars_osp.yml`.

`default_vars_ec2.yml` for aws

ansible/configs/sap-hana-auto/software.yml
```diff
@@ -0,0 +1,48 @@
...
+- name: Deploy Ansible Tower
+  hosts: towers
+  gather_facts: False
+  become: yes
+  tasks:
+    # This is a bit hacky but we are facing some issues with Ansible, RHEL8 and python for some
+    # modules and this workaround solved this particular issue
+    - name: Ensure Python3 package is installed and alternatives for python updated
+      shell: >
+        yum install -y python3 && alternatives --set python /usr/bin/python3
```
fridim: can you split this shell task in 2 tasks, 1 to install the package using `yum` or `package` module, and another one to set alternatives ?

ansible/configs/sap-hana-auto/README.adoc
makentenza: We don't expect this config to be used in AWS, only OSP

ansible/configs/sap-hana-auto/software.yml
makentenza: I tried that already and setting alternatives in a different task after python3 has been installed fails as there is no default python in RHEL 8. Will give it another try with the new image you guys pushed and update here

makentenza:
@fridim I have merged both automated and non-automated configurations in just one now, and moved from 'env_vars' to 'default_vars' as requested.

makentenza:
will create new PR as I'm rebasing my branch so this is cleaner

PR #1001, opened 2020-01-16
loop.index0 changed to loop.index in cloud_template_master.j2 template

prakhar1985:
Changed cloud_template_master.j2 
Issue: 
In multi-server OSP config server names were assigned as server0, server1...
Changed loop.index0 to loop.index for names assigned server1,server2,....

### DIFF
--- a/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
+++ b/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
@@ -10,9 +10,13 @@
 
 - when: _instance.count|int > 1
   block:
-    - loop: "{{ range(1, _instance.count | int + 1) | list }}"
+    - loop: "{{ range(_start_at | int, _instance.count | int + 1) | list }}"
       loop_control:
         loop_var: myindex
       vars:
+          # Be very conservative, ensure every records are deleted
+          # regardless of if it was before or after the patch that makes the instances
+          # start at index1 and not index0.
+          _start_at: "{{ 0 if _dns_state == 'absent' else 1 }}"
         _instance_name: "{{ _instance.name }}{{ myindex }}"
       include_tasks: nested_loop.yml
--- a/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
+++ b/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
@@ -3,7 +3,8 @@
   fail:
     msg: "Instance {{ _instance.name }} doesn't have a 'count'"
 
-- when: _instance.count|int == 1
+- name: Create or delete DNS entry (single instance)
+  when: _instance.count|int == 1
   include_tasks: nested_loop.yml
   vars:
     _instance_name: "{{ _instance.name }}"
@@ -15,7 +16,8 @@
       # start at index1 and not index0.
       _start_at: "{{ 0 if _dns_state == 'absent' else 1 }}"
   block:
-    - loop: "{{ range(_start_at | int, _instance.count | int + 1) | list }}"
+    - name: Create or Delete DNS entries (multiple instances)
+      loop: "{{ range(_start_at | int, _instance.count | int + 1) | list }}"
       loop_control:
         loop_var: myindex
       vars:
--- a/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
+++ b/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
@@ -12,7 +12,7 @@
   block:
     - loop: "{{ range(1, _instance.count | int + 1) | list }}"
       loop_control:
-        index_var: myindex
+        loop_var: myindex
       vars:
         _instance_name: "{{ _instance.name }}{{ myindex }}"
       include_tasks: nested_loop.yml
--- a/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
+++ b/ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml
@@ -9,14 +9,15 @@
     _instance_name: "{{ _instance.name }}"
 
 - when: _instance.count|int > 1
+  vars:
+      # Be very conservative, ensure every records are deleted
+      # regardless of if it was before or after the patch that makes the instances
+      # start at index1 and not index0.
+      _start_at: "{{ 0 if _dns_state == 'absent' else 1 }}"
   block:
     - loop: "{{ range(_start_at | int, _instance.count | int + 1) | list }}"
       loop_control:
         loop_var: myindex
       vars:
-          # Be very conservative, ensure every records are deleted
-          # regardless of if it was before or after the patch that makes the instances
-          # start at index1 and not index0.
-          _start_at: "{{ 0 if _dns_state == 'absent' else 1 }}"
         _instance_name: "{{ _instance.name }}{{ myindex }}"
       include_tasks: nested_loop.yml
--- a/ansible/configs/hacluster-pntae/files/cloud_providers/osp_cloud_template_master.j2
+++ b/ansible/configs/hacluster-pntae/files/cloud_providers/osp_cloud_template_master.j2
@@ -104,7 +104,7 @@ resources:
   #############
 {% for instance in instances %}
   {% for myinstanceindex in range(instance.count|int) %}
-    {% set iname = instance.name if instance.count|int == 1 else [instance.name, loop.index0] | join() %}
+    {% set iname = instance.name if instance.count|int == 1 else [instance.name, loop.index] | join() %}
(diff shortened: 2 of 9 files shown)


stencell:
While I don't think there is anything that is using this necessarily, we should be careful changing a common role like this. Is there a problem with it starting with 0? It seems natural to me...OpenShift does the same with it's ordering of new machines.

prakhar1985:
No there is no problem. It’s just we have always been using server names starting with 1. Server0 for us was very awkward as we were use to server1...n. If it’s effecting any other config we can close the PR.

miteshrh:
@stencell one more thing this change will keep similarity between aws and osp instances if want to use same config for multi cloud. By the way, We can use custom template if changes affect other configs.

stencell:
This is new enough that I don't think that it would cause a problem for any live labs TBH, so while I prefer 0 I like consistency even more.

fridim:
I thought of naming it server1 .. serverN  but there is another place to fix if we want to do so, but i can't remember where it was.

Give me some time to remember

fridim:
i think it was when we were using the nested version of heat templates with `%index%` and heat starts at `0`.

It's also there in those templates:

```
lolo:{imap-too-many-connection-error} ~/sync/dev/aad/ansible$ grep %index% * -r
configs/ocp4-cluster/files/cloud_providers/osp_cloud_template_master.j2:          instance_name: {{ instance['name'] }}_%index%
configs/ocp4-disconnected-osp-lab/files/cloud_providers/osp_cloud_template_master.j2:          instance_name: {{ instance['name'] }}_%index%
roles/infra-osp-template-generate/templates/nested_version/cloud_template_master.j2:          instance_name: {{ instance['name'] }}%index%
```

So i guess it's fine to change it in the common template now that we have a flat template and we control every naming with jinja2.

Can you test:

- make sure three-tier-app deployed with https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/three-tier-app/sample_vars_osp.yml  still works

miteshrh:
**@fridim I ran https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/three-tier-app/sample_vars_osp.yml to test and result fails at following point.** 

TASK [infra-osp-dns : debug] ************************************************************************************************************************************************
Tuesday 21 January 2020  21:26:24 +0530 (0:00:00.045)       0:02:11.079 ******* 
ok: [localhost] => {
    "msg": "The floating IP for app0 is "
}

TASK [infra-osp-dns : DNS entry (present)] **********************************************************************************************************************************
Tuesday 21 January 2020  21:26:24 +0530 (0:00:00.058)       0:02:11.137 ******* 
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Invalid/malformed value"}

PLAY RECAP ******************************************************************************************************************************************************************
localhost                  : ok=53   changed=7    unreachable=0    failed=1    skipped=23   rescued=0    ignored=0 


**I thing loop.index0 is not the right place to change may be somewhere else we need to make change to start name number from 1. Please suggest**

fridim (dismissed):
If we want to change this and start instance names at 1 instead of 0, the DNS module `infra-osp-dns` must be changed too:

```
- when: _instance.count > 1
  block:
    - loop: "{{ range(1, _instance.count | int + 1) | list }}"
      loop_control:
        index_var: myindex
      vars:
        _instance_name: "{{ _instance.name }}{{ myindex }}"
      include_tasks: nested_loop.yml
```

If we do so and update that portion of code and since there are environments deployed with instance names starting at 0, we have to make sure that we delete old DNS entries starting with 0.

Could be manual, not necessary in agnosticd, but it's something to keep in mind with this kind of changes.

fridim:
Unless someone has the time to test this and fix what needs to be, make sure it doesn't break anything, i think we can live with index0
Tagging this PR with "help wanted" for now.

stencell:
I honestly don't see the benefit if fiddling with this now. There are more important things to address and as more content has started to use the OSP provider, we don't know what changing naming defaults would cause it to go out of sync with.

miteshget:
> Unless someone has the time to test this and fix what needs to be, make sure it doesn't break anything, i think we can live with index0
> Tagging this PR with "help wanted" for now.

miteshget:
> If we want to change this and start instance names at 1 instead of 0, the DNS module `infra-osp-dns` must be changed too:
> 
> ```
> - when: _instance.count > 1
>   block:
>     - loop: "{{ range(1, _instance.count | int + 1) | list }}"
>       loop_control:
>         index_var: myindex
>       vars:
>         _instance_name: "{{ _instance.name }}{{ myindex }}"
>       include_tasks: nested_loop.yml
> ```
> 
> If we do so and update that portion of code and since there are environments deployed with instance names starting at 0, we have to make sure that we delete old DNS entries starting with 0.
> 
> Could be manual, not necessary in agnosticd, but it's something to keep in mind with this kind of changes.

@fridim  I have tested after doing small change as follows `_instance_name: "{{ _instance.name }}{{ myindex | int + 1 }}"` and got positive result. Now it will help us to deploy similar env on osp as we get on ec2, specially name convention which help us to keep one course content for both cloud providers'. 

FYI 
I haven't made and change yet, looking for your suggestion .

miteshget:
@fridim Please suggest

miteshget:
```
PLAY RECAP ************************************************************************************************************************************************************************************
app1                       : ok=21   changed=10   unreachable=0    failed=0    skipped=9    rescued=0    ignored=0   
app2                       : ok=21   changed=11   unreachable=0    failed=0    skipped=9    rescued=0    ignored=0   
appdb1                     : ok=21   changed=10   unreachable=0    failed=0    skipped=9    rescued=0    ignored=0   
bastion                    : ok=62   changed=42   unreachable=0    failed=0    skipped=20   rescued=0    ignored=0   
frontend1                  : ok=21   changed=10   unreachable=0    failed=0    skipped=9    rescued=0    ignored=0   
localhost                  : ok=80   changed=12   unreachable=0    failed=0    skipped=29   rescued=0    ignored=0   ```

**Comment by fridim** [comment]
I updated the infra-osp-dns role to work with the new index1 naming.

For deletion, i changed it to be very conservative and delete both  index0 and index1, so DNS of instances created before the change of the PR will still be deleted.


To continue with this, we have to change `index0` to `index` in the following templates:

```
configs/sap-rhvh/files/cloud_providers/osp_cloud_template_master.j2
configs/sap-smart/files/cloud_providers/osp_cloud_template_master.j2
configs/multi-network-app-pntae/files/cloud_providers/osp_cloud_template_master.j2
configs/sap-integration/files/cloud_providers/osp_cloud_template_master.j2
configs/hacluster-pntae/files/cloud_providers/osp_cloud_template_master.j2
```

@makentenza @ashish-k-s Can we make that now ?

**Comment by makentenza** [comment]
> I updated the infra-osp-dns role to work with the new index1 naming.
> 
> For deletion, i changed it to be very conservative and delete both index0 and index1, so DNS of instances created before the change of the PR will still be deleted.
> 
> To continue with this, we have to change `index0` to `index` in the following templates:
> 
> ```
> configs/sap-rhvh/files/cloud_providers/osp_cloud_template_master.j2
> configs/sap-smart/files/cloud_providers/osp_cloud_template_master.j2
> configs/multi-network-app-pntae/files/cloud_providers/osp_cloud_template_master.j2
> configs/sap-integration/files/cloud_providers/osp_cloud_template_master.j2
> configs/hacluster-pntae/files/cloud_providers/osp_cloud_template_master.j2
> ```
> 
> @makentenza @ashish-k-s Can we make that now ?

@fridim as discussed this is fine for me even I can't make changes at the moment to solve any issues on my configs coming after these changes. But will do it in the future.

Also from the list you are missing `configs/sap-hana/files/cloud_providers/osp_cloud_template_master.j2` even you have changed the value in the PR already.

**Comment by fridim** [comment]
Configs update.


PR ready for review

File: ansible/roles-infra/infra-osp-dns/tasks/instance_loop.yml [hunk 1]
```diff
@@ -9,10 +9,15 @@
...
 
 - when: _instance.count|int > 1
+  vars:
+      # Be very conservative, ensure every records are deleted
+      # regardless of if it was before or after the patch that makes the instances
+      # start at index1 and not index0.
+      _start_at: "{{ 0 if _dns_state == 'absent' else 1 }}"
   block:
-    - loop: "{{ range(1, _instance.count | int + 1) | list }}"
+    - loop: "{{ range(_start_at | int, _instance.count | int + 1) | list }}"
```
stencell, line 18: Can we name these tasks? Really all of the tasks in this task list should be updated with names to adhere to our standards. Non-named tasks make JK sad.

stencell (approved):
Other than naming comment that should be fixed, lgtm.

PR #1006, opened 2020-01-16
Revert "ssh_config template update for bastion and set_env_authorized_key roles"

fridim:
Reverts redhat-cop/agnosticd#1000

@miteshrh 
I think we should iterate over the hosts and just use the private ip address, instead of a wildcard 192.168.*. Something in the tone of

```
Host {{ host.name }}
Hostname {{ host.private_ip_address }}
...
```

### DIFF
--- a/ansible/roles/bastion/files/bastion_ssh_config.j2
+++ b/ansible/roles/bastion/files/bastion_ssh_config.j2
@@ -1,4 +1,3 @@
-{% if cloud_provider == 'ec2' %}
 Host ec2* *.internal *.example.com
   User {{remote_user}}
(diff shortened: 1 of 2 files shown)


fridim:
From Nate:

> In this case, https://github.com/redhat-cop/agnosticd/pull/1000 should not have been merged. It sets conditionals with the ec2 and osp providers. Unfortunately, what we would need for the osp provider, which is the *.example.com domain is now only under ec2. This will break our main OpenShift ILT lab if we pick it up the next time we tag.


So also, we just keep the *.example.com  too.

PR #1015, opened 2020-01-17
update config:satellite-vm

miteshrh:
SUMMARY

Config: satellite-vm has extended to support multi cloud like aws and osp

ISSUE TYPE
- Feature enhanced

### DIFF
--- a/ansible/configs/satellite-vm/default_vars.yml
+++ b/ansible/configs/satellite-vm/default_vars.yml
@@ -1,0 +1,61 @@
+---
+
+env_type: satellite-vm 
+output_dir: /tmp/workdir                # Writable working scratch directory
+email: "{{env_type}}@example.com"
+guid: defaultguid
+
+# for file based repo
+repo_method: file 
+use_own_repos: true
+repo_version: "6.4"
+## For RHN login 
+# repo_method: rhn
+# rhsm_pool_ids:
+#   - 8a85f99b6b498682016b521dfe463949
+# rhel_subscription_user:
+# rhel_subscription_pass: 
+######
+
+deploy_local_ssh_config_location: "{{output_dir}}/"
+key_name: ocpkey                        # Keyname must exist in AWS
+use_own_key: true
+env_authorized_key: "{{guid}}key"
+set_env_authorized_key: true
+default_key_name: ~/.ssh/{{key_name}}.pem
+
+install_bastion: true
+install_common: true
+install_ipa_client: false
+tower_run: false
+update_packages: false
+update_all_packages: false
+install_satellite: True
+configure_satellite: false
+
+project_tag: "{{ env_type }}-{{ guid }}"
+
+rhel_repos:
+  - rhel-7-server-rpms 
+  - rhel-server-rhscl-7-rpms 
+  - rhel-7-server-satellite-6.4-rpms 
+  - rhel-7-server-satellite-maintenance-6-rpms 
+  - rhel-7-server-ansible-2.6-rpms
+  - rhel-7-server-extras-rpms
+
+common_packages:
+  - python
+  - unzip
+  - bash-completion
+  - tmux
+  - wget
+  - git
+  - vim-enhanced
+  - at
+  - python27-python-pip
+  - bind-utils
+
+cf_template_description: "{{ env_type }}-{{ guid }} Ansible Agnostic Deployer "
+
+
+...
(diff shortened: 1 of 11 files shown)


ansible/configs/satellite-vm/default_vars_ocp.yml
```diff
@@ -0,0 +1,100 @@
...
+use_dynamic_dns: true
+osp_project_create: true
+student_name: student
+admin_user: opentlc-mgr
+
+
+bastion_instance_type: 2c2g30d
+satellite_instance_type: 8c32g100d
+
+___image: rhel-server-7.7-update-2
```
stencell: why so many _____ for this var? Makes it hard to determine if there is a mistake at a glance.

ansible/configs/satellite-vm/default_vars.yml
```diff
@@ -0,0 +1,61 @@
...
+
+env_type: satellite-vm 
+output_dir: /tmp/workdir                # Writable working scratch directory
+email: "{{env_type}}@example.com"
+guid: defaultguid
+
+# for file based repo
+repo_method: file 
+use_own_repos: true
+repo_version: "6.4"
```
stencell: Why such an old repo? Any reason not to update so you get updated packages with security and other fixes?

ansible/configs/satellite-vm/default_vars_ec2.yml
```diff
@@ -1,17 +1,14 @@
...
+HostedZoneId: Z3IHLWJZOU9SRT
+aws_region: ap-southeast-2
 
-default_key_name: ~/.ssh/{{key_name}}.pem
-tower_run: false
 
 # How many do you want for each instance type
-
-
 bastion_instance_type: "t2.medium"
```
stencell: Consider moving to t3 for more future proof if you are already making and testing changes.

ansible/configs/satellite-vm/default_vars_ec2.yml
```diff
@@ -1,17 +1,14 @@
...
 bastion_instance_image: RHEL75
```
stencell: Why such an old version of RHEL?

ansible/configs/satellite-vm/default_vars.yml
```diff
@@ -0,0 +1,61 @@
...
+common_packages:
+  - python
+  - unzip
+  - bash-completion
+  - tmux
+  - wget
+  - git
+  - vim-enhanced
+  - at
+  - python27-python-pip
```
stencell: Any chance you could look at moving to python3 with these changes to be more future-proof? Several of the common roles have already been adapted to be compatible with either and we are exclusively python3 with the latest OCP on OSP config.

ansible/configs/satellite-vm/default_vars_ocp.yml
miteshrh: changed to one _  thanks.

ansible/configs/satellite-vm/default_vars_ec2.yml
miteshrh: Changed to t3.medium.

ansible/configs/satellite-vm/default_vars.yml
miteshrh: Yeah, This is repo_version=6.4 is for satellite v6.4 and as of now we are doing only satellite 6.4. Hence no need to update, we need to keep it as is.

ansible/configs/satellite-vm/default_vars_ec2.yml
miteshrh: Satellite 6.4 supports only rhel 7 hence at time of writing config I used RHEL75, Now if I want to change then it requires more time to test bit by bit of satellite installation and configuration. So I will keep this version, may be later we will change.

ansible/configs/satellite-vm/default_vars.yml
miteshrh: Few of our modules which we use daily does not like python3 i.e. os_auth is still using python2.7 based openstacksdk so we like to keep as it is.

stencell: Your call...we use python3 for ocp4-disco environment and that includes openstacksdk as well.

miteshrh:
@stencell I have made changes as asked and made some changes as per suggestion also. Please review and merge.

PR #1037, opened 2020-01-23
Wk cluster

wkulhanek:
This config adds ocp4-cluster. And makes a few changes to related roles.

The config works except for Registry storage. Will need to test on Blue.

But would be good to review and merge this bit.


ansible/configs/ocp4-cluster/files/requirements_k8s.txt
```diff
@@ -0,0 +1,36 @@
+adal
```
stencell: Can we consider pinning these to a version with pip3 freeze?

ansible/configs/ocp4-cluster/files/requirements_k8s.txt
```diff
@@ -0,0 +1,40 @@
+adal==1.2.2
```
wkulhanek: @stencell Done. :-)

ansible/configs/ocp4-cluster/default_vars.yml
```diff
@@ -0,0 +1,195 @@
...
+student_name: lab-user
+
+# Enable this if you want to use IPA for user authentication.
+# Mutually exclusive with {{ install_student_user }}
+install_ipa_client: false
+
+# TODO: What does this really do besides run the role?
+set_env_authorized_key: true
+env_authorized_key: "{{guid}}key"
+key_name: "default_key_name"
```
fridim: key_name should not be needed anymore when `cloud_provider == osp` thanks to this recent commit: https://github.com/redhat-cop/agnosticd/commit/926a092587159ebcc816cbdbe39db71b1f5b702b#diff-beb8a8b8bdf73e75e38e21e91cb5ef28R2

PR #1044, opened 2020-01-23
Added RHEL81 ami region mappings

bkoz:
##### SUMMARY

This change allows RHEL8.1 AMI's to be deployed.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Adds a RHEL81 ami to roles/infra-ec2-template-generate/defaults/main.yml
##### ADDITIONAL INFORMATION

```paste below

```


stencell:
@fridim Is this a good opportunity to look at doing mappings to the non-hourly AMIs and take advantage of the cloud access images to save cost?

fridim:
I think this PR is simple enough to be merged

The problem with the cloud access images (let's call them GOLD images) is that the target aws account should have access to them.

Not all of our accounts have access.

So we should come up with a plan to address that.

PR #1050, opened 2020-01-24
Add agnosticd_user_info module

jkupferer:
##### SUMMARY

Add agnostid_user_info module implemented as action plugin.

This module is proposed as a replacement for using debug statements.
Switching to this module allows us to more easily programatically
capture output by putting user info directly into the output dir.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

agnosticd user info

##### ADDITIONAL INFORMATION

This PR only adds the `agnosticd_user_info` module. Configs and workloads will need to be adapted to use it. In order to convert to `agnosticd_user_info`:

Go from this:

```
- name: Show generated password
  debug:
    msg: "user.info: generated password {{ generated_password }}"
```

To this:

```
- name: Show generated password
  agnosticd_user_info:
    msg: "generated password {{ generated_password }}"
```

This module also supports outputing structured data, so the following may be preferred to allow the value to be picked up by automation more easily:

```
- name: Show generated password
  agnosticd_user_info:
    msg:
      generated_password: "{{ generated_password }}"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ ansible/configs/*/roles
 my_*_vars.yml
 
 .vscode/settings.json
+__pycache__
--- a/ansible/action_plugins/agnosticd_user_info.py
+++ b/ansible/action_plugins/agnosticd_user_info.py
@@ -52,7 +52,7 @@ class ActionModule(ActionBase):
                 )
             )
             fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
-            fh.write('- ' + json.dumps(result['msg']) + "\n")
+            fh.write('- ' + json.dumps(self._task.args['msg']) + "\n")
             fh.close()
             result['failed'] = False
         except Exception as e:
--- a/ansible/action_plugins/agnosticd_user_info.py
+++ b/ansible/action_plugins/agnosticd_user_info.py
@@ -22,7 +22,7 @@ __metaclass__ = type
 import json
 import os
 
-from ansible.errors import AnsibleUndefinedVariable
+from ansible.errors import AnsibleError, AnsibleUndefinedVariable
 from ansible.module_utils.six import string_types
 from ansible.module_utils._text import to_text
 from ansible.plugins.action import ActionBase
@@ -31,6 +31,7 @@ class ActionModule(ActionBase):
     '''Print statements during execution and save user info to file'''
 
     TRANSFERS_FILES = False
+    _VALID_ARGS = frozenset(('msg',))
 
     def run(self, tmp=None, task_vars=None):
         self._supports_check_mode = True
@@ -45,7 +46,11 @@ class ActionModule(ActionBase):
         result['_ansible_verbose_always'] = True
 
         try:
-            output_dir = task_vars.get('output_dir', task_vars.get('playbook_dir', '.'))
+            output_dir = task_vars.get('output_dir',
+                task_vars['hostvars'].get('localhost',{}).get('output_dir',
+                    task_vars.get('playbook_dir', '.')
+                )
+            )
             fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
             fh.write('- ' + json.dumps(result['msg']) + "\n")
             fh.close()
--- a/ansible/action_plugins/agnosticd_user_info.py
+++ b/ansible/action_plugins/agnosticd_user_info.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python
+# 
+# Copyright: (c) 2020, Johnathan Kupferer <jkupfere@redhat.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+#
+# This plugin is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import json
+import os
+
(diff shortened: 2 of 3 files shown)


tonykay:
So the `user.info` is key to the email generation and existing processes. I would suggest maintaining the ability to prefix the information with `user_info:`. Making this behavior default but adding the ability to. Turn it off. Also I think there are advantages into each developer being  able to output whatever it is they care about e.g. `user_info: s3 bucket is http://...`

jkupferer:
@tonykay The `agnosticd_user_info` module always prepends `user-info: ` to all messages displayed in the output log:
https://github.com/redhat-cop/agnosticd/blob/agnosticd-user-info/ansible/action_plugins/agnosticd_user_info.py#L44

There isn't currently an ability to turn this off. I thought about making it optional... that would mean user info could be set in the output file that wouldn't be displayed in the old way, but why implement a feature we probably don't need?

fridim:
To echo what Tony said:

Current process is: cloudforms parses the logs that way:

```ruby
    userInfoResult ||= searchLog('\"user.info:', logfile, sshHost, sshUser, sshKey)
    unless userInfoResult[0].nil? || userInfoResult[0] == ""
      userInfoResult[0] = userInfoResult[0].gsub(/.*"user.info: /,'')
      userInfoResult[0] = userInfoResult[0].gsub(/",/,'')
      userInfoResult[0] = userInfoResult[0].gsub(/"}/,'')
      userInfoResult[0] = userInfoResult[0].gsub(/"$/,'')
      c = 0
      userInfoResult[0].split(/\n+/).each do |line|
        $evm.root['service_template_provision_task'].destination.custom_set("Environment Info #{c.to_s}", line)
        c = c + 1
      end
      body +=<<EOT
```

Keep in mind that it's kind of fragile.

For example:
- if we change ansible.cfg and set `stdout_callback = yaml`, it doesn't work, cloudforms doesn't catch it.
- It has to work when debug tasks are skipped. The string is printed, but differently:

```
TASK [Print Student aws access as user.info] ***********************************
Friday 24 January 2020  13:04:40 -0500 (0:00:01.684)       0:36:04.563 ******** 
skipping: [localhost] => (item=user.info: Top level domain: .example.opentlc.com) 
skipping: [localhost] => (item=user.info: ) 
skipping: [localhost] => (item=user.info: WARNING: with great power comes great responsibility. We monitor usage.) 
skipping: [localhost] => (item=user.info: Your AWS programmatic access:) 
skipping: [localhost] => (item=user.info: aws_access_key_id = ...) 
skipping: [localhost] => (item=user.info: aws_secret_access_key = ...) 
```

In your action module you just write to a file, right ?

What is the difference with:

```yaml
- copy:
    content: ...
    dest: user-info.yml
```
?

fridim:
We could improve Cloudforms' code

- look for a `user-info.yaml` file instead of parsing the logs,  grab it and convert to structured data in ruby
- print the email using that data
- if no `user-info.yaml` found, fallback to old method, parse the logs

But it should be changed everywhere in all cloudforms email methods (all instances)

Let's discuss with @rut31337

jkupferer:
@fridim lots of differences using `agnosticd_user_info` from using the `copy` module.

* appends to a `user-info.yaml` file in the `output_dir`, so it can be called repeatedly
* is an action plugin, and so always executes on localhost
* uses `_ansible_verbose_always` to always print output like debug

I used the `debug` action plugin as reference when implementing `agnosticd_user_info` to get the same behavior as much as possible.

Using this playbook...

```
- hosts: localhost
  gather_facts: false
  tasks: 
  - debug:
      msg: "user.info: hello, world"
  - agnosticd_user_info:
      msg: "hello, world"
```

This is the log output:

```
PLAY [localhost] ***********************************************************************

TASK [debug] **************************************************************************
Sunday 26 January 2020  09:16:21 -0600 (0:00:00.023)       0:00:00.023 ******** 
ok: [localhost] => {
    "msg": "user.info: hello, world"
}

TASK [agnosticd_user_info] **********************************************************
Sunday 26 January 2020  09:16:21 -0600 (0:00:00.016)       0:00:00.039 ******** 
ok: [localhost] => {
    "changed": false,
    "msg": "user.info: hello, world"
}

PLAY RECAP ***********************************************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

jkupferer:
I don't understand the bit about the skipped tasks though... this seems bizarre...

So something like this...

https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/ocp4-ha-lab/software.yml#L225-L238
```
    - name: Print Student aws access as user.info
      debug:
        msg: "{{ item }}"
      with_items:
        - "user.info: Top level domain: {{ subdomain_base_suffix }}"
        - "user.info: "
        - "user.info: WARNING: with great power comes great responsibility. We monitor usage."
        - "user.info: Your AWS programmatic access:"
        - "user.info: aws_access_key_id = {{ student_access_key_id }}"
        - "user.info: aws_secret_access_key = {{ student_secret_access_key }}"
      when:
        - not install_ocp4 | bool
        - student_access_key_id is defined
        - student_secret_access_key is defined
```

Are you saying that the the user info is picked up from these even if the tasks are skipped? So the `when` clause is functionally ignored?

fridim:
i'm saying that when the task is skipped, the message is printed, but differently and the regex in cloudforms is smart enough to not pick it up.

jkupferer:
Cool, so same deal here. The message being printed when it is skipped is just core ansible, nothing special there. One difference would be that this...

```
- hosts: localhost
  gather_facts: false
  tasks:
  - agnosticd_user_info:
      msg: "{{ item }}"
    with_items:
    - foo
    - bar
    when: false
```

Prints:

```
PLAY [localhost] **************************************************************

TASK [agnosticd_user_info] ****************************************************
Monday 27 January 2020  10:21:04 -0600 (0:00:00.036)       0:00:00.036 ********
skipping: [localhost] => (item=foo) 
skipping: [localhost] => (item=bar) 

PLAY RECAP ********************************************************************
localhost : ok=0    changed=0    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
```

No "user.info: " prefix. I don't think it matters.

fridim:
OK, that's fine.

Now the question is, should we search and replace in the whole code base  at once ?

jkupferer:
"should we search and replace in the whole code base at once?"

I am confident, but not that confident. I'd rather get this in, then switch over a single config or workload to verify, and then later switch over the rest.

fridim:
i know python2 support dropped, but did you test it though?

PR #1053, opened 2020-01-27
empty-config: Add aws_region as it is needed by destroy.yml playbook

fridim:
This commit fixes the following error:

```
TASK [Set aws_region_final as provided with aws_region] ***************************************************************************************************************************************
Monday 27 January 2020  13:05:08 +0000 (0:00:00.028)       0:00:02.343 ********                                                                                                                
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'aws_region' is undefined\n\nThe error appears to be in '/ansible/cloud_provider
s/ec2_detect_region_tasks.yml': line 26, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# else just set as the prov
ided aws_region\n- name: Set aws_region_final as provided with aws_region\n  ^ here\n"}  
```

That happens when we use `ansible/destroy.yml`.

(The error is not present when using `destroy_env.yml` directly from the config)


fridim:
Yes, i just want to have a default when run from laptop.
When i will create the agnosticv vars, i will pick up whatever is passed from Cloudforms, eventually, if we want to test different region endpoints

PR #1059, opened 2020-01-28
Remove install_dependencies var for ftl-injector task in bastion role

stencell:
##### SUMMARY
The `bastion` role passes `install_dependencies: false` to the `ftl-injector` role. This causes the conditional check to fail and no virtual envs are set up for FTL. The default in the `ftl_injector` role is true for this.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- bastion

### DIFF
--- a/ansible/roles/bastion/tasks/main.yml
+++ b/ansible/roles/bastion/tasks/main.yml
@@ -117,7 +117,6 @@
       name: ftl-injector
     vars:
       student_login: "{{ student_name }}"
-      install_dependencies: false
 
 - name: Install jq on the bastion
   get_url:


wkulhanek:
@stencell I added that because it broke otherwise (quite a while ago). Does this now work?

stencell:
Well it definitely breaks without it 😆 

Do you remember what it broke before?

Here is what is installed when this is set to true:
```
yum_dependencies:
  - ansible-tower-cli
  - autoconf
  - automake 
  - git
  - libtool 
  - python
  - python-devel  
  - python-pip

pip_dependencies:
  - virtualenv
  - ansible-tower-cli
```

stencell:
Tested local deploys of ocp-ha-lab and clientvm. Lets go ahead and merge this and we'll increment the tag for the installation lab, which is all that requires this right now.

PR #1063, opened 2020-01-29
Clientvm osp

wkulhanek:
Redo ClientVM Configurations:

- Add support for OpenStack
- Generate repo lists dynamically from rhel_repos variable in default_vars.
- Unify ocp-client-vm and ocp4-client-vm roles ==> ocp-client-vm
- Add more meaningful variables to pin software versions (while keeping default support for osrelease)


tonykay:
Oh that prompt :(

PR #1077, opened 2020-01-31
config: Satellite-vm

miteshget:
Readme.adoc file updated


stencell:
This change looks like it ended up merging just a readme change? I mean, not a big deal, but in the one sentence there are like 4 typos. We should try to be better and more professional with the documentation we write.

How about...

"With {config}, we can deploy bastion and Satellite on Openstack and AWS cloud providers.

PR #1085, opened 2020-02-01
Added a sample vars file for a private three-tier-app on OSP

tonykay:
##### SUMMARY
Added a sample vars file for three-tier-app using a new pattern of config/<config-name>/sample_vars/ directory as a form of blueprints to allow users to start a particular customized config quickly.

This one allows three-tier-app to deploy onto an OSP cluster with all but the bastion private and not exposed.

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME
three-tier-app enhancement
 
##### ADDITIONAL INFORMATION

No change to the code base, just an optional vars files

### DIFF
--- a/ansible/configs/three-tier-app/sample_vars/private-dns-osp.yml
+++ b/ansible/configs/three-tier-app/sample_vars/private-dns-osp.yml
@@ -1,0 +1,161 @@
+# sample vars configuration file to deploy three-tier-app to OSP
+#
+# Only the bastion is exposed and the other hosts do not have public DNS or
+# Floating IPs
+# Consider exposing Frontends via public_dns: true in the instance dictionary
+#
+# Usage: ansible-playbook main.yml -e @configs/three-tier-app/sample_vars/private-dns-osp.yml
+# 
+# Ideally make and keep a copy OUTSIDE your repo, especially if using Cloud Credentials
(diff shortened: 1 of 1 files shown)


ansible/configs/three-tier-app/sample_vars/private-dns-osp.yml
```diff
@@ -0,0 +1,161 @@
...
+# Instance type - you can list these with `openstack image list`
+
+bastion_instance_type: 2c2g30d
+app_instance_type: 2c2g30d
+appdb_instance_type: 2c2g30d
+frontend_instance_type: 2c2g30d
+support_instance_type: 2c2g30d
+
+#___image: rhel-guest-7.7u2    # blue
+___image: rhel-server-7.7-update-2  # red
```
stencell: Just a suggestion...make this var with less _. I don't get the reason for such a long prefix before image since it would be easy to miss one and cause failures when other vars rely on this.

fridim: that's me.

The purpose of this var was just not to repeat myself in the following lines. The scope of this var was just 3-4 lines and should not be used anywhere else, i just added quickly `_`s because 'image' was too generic and i could have risk of collision.

Feel free to rename to anything that makes sense, or just duplicate the string maybe and remove the var.

PR #1088, opened 2020-02-03
Small change to support Python2

agonzalezrh:
##### SUMMARY

Using Python2 generates the strings from an array with module template with u"TEST" instead TEST

Playbook:

> - hosts: localhost
>   gather_facts: false
>   vars:
>     servers: 
>      - "8.8.8.8"
>      - "1.1.1.1"
>   tasks:
>   - template:
>      src: templates/test.j2
>      dest: /tmp/test

Template:

> Original: {{ servers }}
> After PR: [{{ servers | list | join(",") }}]

File generated using Python 2:

> Original: [u'8.8.8.8', u'1.1.1.1']
> After PR: [8.8.8.8,1.1.1.1]


fridim:
We dropped the nested version, you probably want to change the other template https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/infra-osp-template-generate/templates/cloud_template_master.j2

tonykay:
(we need to get off python2 in the future)

PR #1096, opened 2020-02-04
Add completion callback

jkupferer:
##### SUMMARY

Add completion callback and userinfo to agnosticd for babylon integration.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Main entry point playbooks

##### ADDITIONAL INFORMATION

When initially adopted this will result in duplicate completion callbacks from ansible tower and agnosticd. These will not cause any issues with babylon processing.

### DIFF
--- a/ansible/configs/test-empty-config/post_software.yml
+++ b/ansible/configs/test-empty-config/post_software.yml
@@ -22,9 +22,9 @@
       pause:
         seconds: 30
 
-    - debug:
+    - agnosticd_user_info:
         msg: >-
-          user.info: Some random password
+          Some random string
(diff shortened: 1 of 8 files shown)


ansible/configs/test-empty-config/post_software.yml
```diff
@@ -22,9 +22,9 @@
       pause:
         seconds: 30
 
-    - debug:
+    - agnosticd_user_info:
         msg: >-
-          user.info: Some random password
+          Some random string
```
fridim: Note to myself: update https://github.com/redhat-cop/babylon/blob/master/test/babylon-empty-config.groovy#L132

PR #1105, opened 2020-02-06
Add data support to agnosticd_user_info

jkupferer:
##### SUMMARY

Add ability to collect data in the output directory and pass the data in job completion callback.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

agnosticd_user_info

##### ADDITIONAL INFORMATION

Usage example:
```
    - name: Return user info data:
      agnosticd_user_info:
        msg: Generated password is {{ generated_password }}
        data:
          generated_password: "{{ generated_password }}"
```

### DIFF
--- a/ansible/action_plugins/agnosticd_user_info.py
+++ b/ansible/action_plugins/agnosticd_user_info.py
@@ -31,7 +31,7 @@ class ActionModule(ActionBase):
     '''Print statements during execution and save user info to file'''
 
     TRANSFERS_FILES = False
-    _VALID_ARGS = frozenset(('msg',))
+    _VALID_ARGS = frozenset(('msg','data'))
 
     def run(self, tmp=None, task_vars=None):
         self._supports_check_mode = True
@@ -42,7 +42,19 @@ class ActionModule(ActionBase):
         result = super(ActionModule, self).run(tmp, task_vars)
         del tmp # tmp no longer has any effect
 
-        result['msg'] = 'user.info: ' + self._task.args['msg']
+        msg = self._task.args.get('msg')
+        data = self._task.args.get('data')
+
+        if msg:
+            result['msg'] = 'user.info: ' + msg
+        if data:
+            result['data'] = data
+            if not isinstance(data, dict):
+                result['failed'] = True
+                result['error'] = 'data must be a dictionary of name/value pairs'
+                return result
+
+        # Force display of result like debug
         result['_ansible_verbose_always'] = True
 
         try:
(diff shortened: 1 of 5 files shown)


ansible/action_plugins/agnosticd_user_info.py
```diff
@@ -53,9 +65,15 @@ def run(self, tmp=None, task_vars=None):
...
-            fh.write('- ' + json.dumps(self._task.args['msg']) + "\n")
-            fh.close()
+            if msg:
+                fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
+                fh.write('- ' + json.dumps(msg) + "\n")
+                fh.close()
+            if data:
+                fh = open(os.path.join(output_dir, 'user-data.yaml'), 'a')
+                for k, v in data.items():
+                    fh.write("{0}: {1}\n".format(k, json.dumps(v)))
```
fridim: I'm wondering  what would happen if the same key is written 2 times ?
There would be duplicates in the file and that would just be a warning ?
Can't we just use `json.dump(data)` directly ?

```python
import json 
import os
try:
    fh = open(os.path.join(output_dir, 'user-data.yaml'), 'r')
    cur_data=json.loads(fh.read())
    cur_data.update(data)
    fh = open(os.path.join(output_dir, 'user-data.yaml'), 'w')
    json.dump(cur_data, fh)
except:
    fh = open(os.path.join(output_dir, 'user-data.yaml'), 'w')
    json.dump(data, fh)
```

I used `try` here and you would need to handle errors better than that, but it's just to illustrate. Wdty?

ansible/action_plugins/agnosticd_user_info.py
```diff
@@ -53,9 +65,15 @@ def run(self, tmp=None, task_vars=None):
                     )
                 )
             )
-            fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
-            fh.write('- ' + json.dumps(self._task.args['msg']) + "\n")
-            fh.close()
+            if msg:
+                fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
```
fridim: What happens to `user-info.yaml` if ansible-playbook command is run several time ? Do we have the same messages stacked ? Or is the file reset between calls of `ansible-playbook` ?

ansible/action_plugins/agnosticd_user_info.py
jkupferer: It seemed simpler to me to not have to read the file in, just taking advantage of JSON documents being valid inside of YAML documents. It's not a strong opinion, but it is more lines of code and more cpu/memory usage to read it in, update, and write it back out.

If someone set the same data twice then... well, that would be a silly thing to do. Undefined behavior. Probably the YAML parser would override the first setting with the second, but no promises.

ansible/action_plugins/agnosticd_user_info.py
jkupferer: The output directory should start fresh for each run of `ansible-playbook`. The initialization also overwrites the file with `---`.

https://github.com/redhat-cop/agnosticd/blob/babylon-callback-data/ansible/setup_runtime.yml#L33-L40

fridim: i was asking for  when working from laptop, ansible-playbook is run several times.
in that case, the output_dir is not cleaned up

jkupferer: The file is still reset, so should be fine

ansible/action_plugins/agnosticd_user_info.py
fridim: I was thinking when developing from laptop, could that fail because of consecutive runs ?

fridim: OK i see the initialization now. All good

PR #1111, opened 2020-02-08
Certbot breaking in python3

stencell:
##### SUMMARY
Another python2 vs python3 problem.

This [task](https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml#L47-L50) was added in #1089 and unfortunately doesn't work in python3 environments since it is not specifying the executable so when we have no pip and no executable defined, it fails.

This PR is a workaround for now. We need a real effort to fix these python2/3 issues, but I need this to unblock for now. Fortunately not all configs have been tagged otherwise all of the OCP4-disconnected deploys would have failed all week.

The PR also has has a small change to add the port option for the DNS tasks similar to what was added to the role.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-disconnected-ha-lab
host-lets-encrypt-certs-certbot

### DIFF
--- a/ansible/configs/ocp4-disconnected-osp-lab/destroy_env.yml
+++ b/ansible/configs/ocp4-disconnected-osp-lab/destroy_env.yml
@@ -19,6 +19,7 @@
         zone: "{{ osp_cluster_dns_zone }}"
         record: "{{ item }}.{{ guid }}"
         type: A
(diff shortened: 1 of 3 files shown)


stencell:
@wkulhanek This works around your change, so please review.

PR #1136, opened 2020-02-12
Enable service catalog and template service broker for odo lab

jankleinert:
##### SUMMARY
Enable service catalog and template service broker for odo lab

- Feature Pull Request

##### COMPONENT NAME
ocp-workload-odo-workshop

### DIFF
--- a/ansible/roles/ocp-workload-odo-workshop/files/tsb-instance.yml
+++ b/ansible/roles/ocp-workload-odo-workshop/files/tsb-instance.yml
@@ -1,0 +1,6 @@
+apiVersion: osb.openshift.io/v1
+kind: TemplateServiceBroker
+metadata:
+  name: openshifttemplateservicebroker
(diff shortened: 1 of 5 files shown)


stencell:
Can we rename this to ocp4-workload-* since this is not really associated with OCP3?

jankleinert:
@stencell That's fine with me - is that something you'd like me to submit a PR for? I'd go with ocp4-workload-homeroomlab-odo-workshop for consistency. Also, what's the process for making sure than any agnosticv configs also get updated to reflect new naming? I want to make sure nothing gets broken as a result of changing the name.

stencell:
Fair point. I'd probably just duplicate these to ocp4* and then open a ticket (rhpds-help@redhat.com) to get the agnosticv configs updated. Once that is done, you can PR the removal of the ocp* version.

Obviously not super critical to do this...just a best effort for consistency that will battle for your time like everything else.

PR #1140, opened 2020-02-12
Fix for ansible 2.7, make sure the callback key is present

fridim:
This commit if applied fixes the following error with ansible 2.7

```
fatal: [localhost]: FAILED! => {"msg": "The conditional check
'__meta__.callback.url
| default('') != ''' failed. The error was: error while evaluating conditional
(__meta__.callback.url | default('') != ''): '__meta__' is undefined
```

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/ansible/completion_callback.yml
+++ b/ansible/completion_callback.yml
@@ -5,7 +5,7 @@
     - name: Attempt completion callback
       when:
       - __meta__ is defined
-      - 'callback' in __meta__
+      - "'callback' in __meta__"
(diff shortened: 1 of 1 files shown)


fridim:
@jkupferer when __meta__ is defined, will always be a callback key ?

can we double check with what i put in that PR ? It doesn't cost anything to have an extra test.

jkupferer:
Closing this PR as it should not be necessary...

Issue #1027, opened 2020-01-21
Parameterize wait on stop instances


jimrigsbee:
**Is your feature request related to a problem? Please describe.**
Some processes need to ensure that we wait until all instances of an OpenShift cluster have stopped completely before ending the AgnosticD job.

**Describe the solution you'd like**
The job author can set a parameter to wait for stop.

**Describe alternatives you've considered**
Hard coding to true.  This may not work for everyone, so a parameter is requested.

**Additional context**
None.

jimrigsbee:
I will be merging code soon that will fix this issue.

PR #1172, opened 2020-02-20
Parameterize the waiting of AWS power state.

### LINKED ISSUE #1026: Avoid trying to shutdown or start up terminated AWS instances
**Describe the bug**
The lifecycle.yml playbook will attempt to shutdown or start all instances that match a guid. We need to avoid trying to start or stop terminated instances.

**To Reproduce**
Steps to reproduce the behavior, ex:
1. Install an OpenShift cluster, cluster bootstrap machine will be terminated at the end of install.
2. Try to stop the cluster with the ocp4-workshop/lifecycle.yml and AWS will report that it cannot stop the terminated instance causing the AgnosticD job to fail.

**Expected behavior**
No AWS terminated instances will be included in the start/stop logic.

**Screenshots / logs**
None.

**Versions (please complete the following information):**
 - OS - any
 - AAD version (`git log --pretty=oneline -2`)
 - Ansible (`ansible --version`) - any
 - cloud provider CLI (aws-cli,  azure, ..) - aws-cli 

**Additional context**
So far this is noted in the ocp4-workshop config...could be a bug in others that include a lifecycle play.

rmahroua:
##### SUMMARY
Add a 'aws_instance_wait_for_stop' variable to control the AWS
instance state return.

By default, `lifecycle.yml` does not wait.

Fixes #1026, #1027

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`ocp4-workshop`

##### ADDITIONAL INFORMATION
These changes allow the user to parameterize the instance state for performance enhancements -- if `aws_instance_wait_for_stop` is set to `false`, the life-cycle playbook does not wait for the full stop of the instances to continue its execution.

### DIFF
--- a/ansible/configs/ocp4-workshop/lifecycle.yml
+++ b/ansible/configs/ocp4-workshop/lifecycle.yml
@@ -64,10 +64,11 @@
       when: ACTION == 'stop'
       ec2_instance:
         state: stopped
-        wait: no
+        wait: "{{ aws_instance_wait_for_stop }}"
         filters:
           "tag:guid": "{{ guid }}"
           "tag:env_type": "{{ env_type }}"
+          instance-state-name: running
 
     - name: Start instances by (guid, env_type) tags
       when: ACTION == 'start'
@@ -77,7 +78,7 @@
         filters:
           "tag:guid": "{{ guid }}"
           "tag:env_type": "{{ env_type }}"
-
+          instance-state-name: stopped
     - when: ACTION == 'status'
       block:
         - name: Get EC2 facts using (guid, env_type) tag
--- a/ansible/configs/ocp4-workshop/README.adoc
+++ b/ansible/configs/ocp4-workshop/README.adoc
@@ -22,7 +22,7 @@ just replace in the command used to deploy, the `main.yml` with `destroy.yml`
 
 * `ansible-playbook ansible/destroy.yaml -e @/home/myself/myenvironment-variables.yml -e@/home/myself/my_secret.yml`
 
-== Sample Environmet Variables
+== Sample Environment Variables
 
 [source,yaml]
 ----
@@ -32,6 +32,7 @@ guid: myguid
 cloud_provider: ec2
 cloud_tags: {'Purpose': 'development'}
 aws_region: us-east-2
+aws_instance_wait_for_stop: true
 cloudformation_retries: 0
 key_name: ocpkey
 software_to_deploy: none
--- a/ansible/configs/ocp4-workshop/env_vars.yml
+++ b/ansible/configs/ocp4-workshop/env_vars.yml
@@ -87,6 +87,8 @@ HostedZoneId: Z3IHLWJZOU9SRT
 aws_region: us-east-1
 # The key that is used to
 key_name: "default_key_name"
+# Whether to wait for AWS to return that the instance has stopped
+aws_instance_wait_for_stop: false
 
 ## Networking (AWS)
 subdomain_base_short: "{{ guid }}"


rmahroua:
@jimrigsbee  @joelbirchler AWS power state PR

joelbirchler:
Travis fails on Python 2, but our change appears unrelated.

joelbirchler:
The Travis fail was due to missing quotes around {{ }} chars.

ansible/configs/ocp4-workshop/lifecycle.yml
```diff
@@ -64,10 +64,11 @@
       when: ACTION == 'stop'
       ec2_instance:
         state: stopped
-        wait: no
+        wait: "{{ aws_instance_wait_for_stop }}"
         filters:
           "tag:guid": "{{ guid }}"
           "tag:env_type": "{{ env_type }}"
+          instance-state-name: running
```
fridim: Why do we have to specify the state here ?

is there any other type that should be included ?  "starting"  "pending"  ... ?

I'd rather not touch this part, especially since i think this should be taken care of by the ansible module:

https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/amazon/ec2_instance.py#L1532-L1542

ansible/configs/ocp4-workshop/lifecycle.yml
```diff
@@ -77,7 +78,7 @@
         filters:
           "tag:guid": "{{ guid }}"
           "tag:env_type": "{{ env_type }}"
-
+          instance-state-name: stopped
```
fridim: same here

ansible/configs/ocp4-workshop/lifecycle.yml
jimrigsbee: We were running into problems with trying to stop things that were not running like terminated instances.

ansible/configs/ocp4-workshop/lifecycle.yml
jimrigsbee: same comment....if we try to start things that are terminated we run into problems

jimrigsbee:
The state should be defaulted to no:
`wait: "{{ aws_instance_wait_for_stop }} | d(no)"`

ansible/configs/ocp4-workshop/lifecycle.yml
fridim: Did you open an issue to ansible/ansible project ?

it seems to me that this should be taken care by the module:

https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/amazon/ec2_instance.py#L1533

It doesn't try to stop an already stopped instance, so it should not try to stop an instance marked  as Terminated.

The problem with the filter approach is that it's not exhaustive.

We can add more filters to reduce the set, but we can't do, as far i know, inclusive OR.

```
filters: running | pending | starting | ...
state: stopped
```

ansible/configs/ocp4-workshop/lifecycle.yml
fridim: i guess this happens only when developing and reusing the same guid

ansible/configs/ocp4-workshop/lifecycle.yml
```diff
@@ -64,10 +64,11 @@
       when: ACTION == 'stop'
       ec2_instance:
         state: stopped
-        wait: no
+        wait: "{{ aws_instance_wait_for_stop }}"
```
fridim: can we have a default here ?

```yaml
wait: "{{ aws_instance_wait_for_stop | default(false) }}"
```

fridim:
I'm closing this Pull request.

Feel free to reopen and address the comments made.

PR #1035, opened 2020-01-23
disable gitea

pb82:
This PR disables the gitea installation for current Integreatly workshops. This is because the image that the gitea operator is deployment no longer exists. A fixed version of Gitea will be included in the next release again.

PR #1222, opened 2020-03-03
Clean up intly vars

grdryn:
##### SUMMARY

In the integreatly ansible installation repo, we've just changed the default for Gitea to be "false", so there shouldn't be a need to override that here.

Additionally, the default for amq_streams is already "true" in the group vars for the "PDS" flavour of integreatly (which is used here), so that also shouldn't need to be overridden.

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME

integreatly

### DIFF
--- a/ansible/roles/ocp-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp-workload-integreatly/defaults/main.yml
@@ -8,7 +8,6 @@ inventory_hosts_file: inventories/pds.template
 release_tag: "release-{{ ig_version }}"
 webapp_namespace: webapp
 self_signed_certs_enabled: True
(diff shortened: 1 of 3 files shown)


grdryn:
@pb82 @pmccarthy in https://github.com/redhat-cop/agnosticd/pull/1035 you say:

> A fixed version of Gitea will be included in the next release again.

Could you confirm that this won't be the case?

pb82:
@grdryn Gitea has been fixed in the installer (https://github.com/integr8ly/installation/pull/1243) which means we can include it again, but we don't have to. As far as i know it is not used anywhere so it's safe to remove it.

PR #1228, opened 2020-03-04
Support custom compute flavor

tkummer33:
##### SUMMARY

created new variable "compute_flavor" in config for ocp4-cluster, which is used in install-config.yaml.j2. Currently there is hardcoded value in the install-config.yaml.j2 template.

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME
ocp4-cluster -> default_vars_osp.yml
ocp4-cluster -> install-config.yaml.j2
```

### CODE REVIEW

File: ansible/configs/ocp4-cluster/default_vars_osp.yml [hunk 1]
```diff
@@ -185,3 +188,6 @@ master_instance_type: 4c16g30d
 
 # Worker Instance Type
 worker_instance_type: 4c16g30d
+
+# Compute Flavor
+compute_flavor: 4c16g30d
```
wkulhanek: I'm not sure I get it. What's the difference between compute_flavor and worker_instance_type? Why is worker_instance_type not enough?

PR #1244, opened 2020-03-05
Adding a Google Cloud Provider and the ocp4-cluster as gcp enabled

vincepower:
##### SUMMARY
Adds a cloud_provider for Google Cloud

And since every cloud provider needs a test case, I added support to ocp4-cluster to use the Google Cloud provider

##### ISSUE TYPE
- Feature Pull Request
- New role Pull Request

##### COMPONENT NAME
cloud_provider components:
ansible/cloud_providers/gcp_destroy_env.yml
ansible/cloud_providers/gcp_infrastructure_deployment.yml
ansible/cloud_providers/gcp_ssh_config_setup.yml

Updated config/ocp4-cluster

Added two new roles
roles/infra-gcp-create-inventory
roles/infra-gcp-template-destroy

##### ADDITIONAL INFORMATION
It works for me (famous last words)


ansible/cloud_providers/gcp_infrastructure_deployment.yml
```diff
@@ -0,0 +1,193 @@
...
+
+    - name: Fail if gcloud not available
+      fail:
+        msg: you need Google Cloud SDK installed
+      when: gcloud_result is failed
+
+    - name: Get SSH public key
+      set_fact:
+        ssh_key: "~/.ssh/{{key_name}}.pem"
+        ssh_key_data: "{{lookup('file', '~/.ssh/{{key_name}}.pub')}}"
```
fridim: Do we have to use a local key ?

We do that for AWS because it's not possible to create a key along with a cloudformation stack.

But for OSP for example, we create an infra ssh key inside the HEAT template and use it for the deployment.

In other words, is it possible to create the infra key inside the gcp template ?
if yes, then i'd like to get rid of any stateful code like this one

```yaml
ssh_key: "~/.ssh/{{key_name}}.pem"
```

And use the key from the template.

Logic is:

- create the stack
- get the key from the output of the stack
- save it to `output_dir`
- use that key instead, and not something from `~/.ssh/`

ansible/cloud_providers/gcp_infrastructure_deployment.yml
```diff
@@ -0,0 +1,193 @@
...
+        gcloud auth
+        activate-service-account "{{ gcp_account }}"
+        --key-file="{{ gcp_credentials_file }}"
+        --project="{{ gcp_project_id }}"
+      environment:
+        PATH: /usr/bin
+      tags:
+        - validate_gcp_template
+        - create_inventory
+        - must
```
fridim: How does this gcloud auth exactly work ?
it seems stateful.

How would that work with concurrent deployments to different gcp accounts in ansible tower ?

Is there a way to use environment variables instead ?

To give an example, here is how we authenticate with OSP:

```yaml
- environment:
    OS_USERNAME: "{{ osp_auth_username }}"
    OS_PASSWORD: "{{ osp_auth_password }}"
  ...
```

And most ansible modules understand those.
That makes the playbooks stateless.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
```diff
@@ -0,0 +1,193 @@
...
+        auth_kind: "{{ gcp_auth_type }}"
+        service_account_file: "{{ gcp_credentials_file }}"
+      register: gcp_managed_zone
+      tags:
+        - gcp_infrastructure_deployment
+      when:
+        - HostedZoneId != "none"
+
+    - name: Add delegation for NS to the main DNSZone
+      gcp_dns_resource_record_set:
```
fridim: This can't be part of the cloud template ?

ansible/cloud_providers/gcp_infrastructure_deployment.yml
```diff
@@ -0,0 +1,193 @@
...
+          - "{{ gcp_managed_zone.resources[0].nameServers[3] }}"
+        state: present
+      tags:
+        - gcp_infrastructure_deployment
+      when:
+        - env_type != "ocp4-cluster"
+        - HostedZoneId != "none"
+
+    - name: Add bastion entry to the main DNSZone
+      gcp_dns_resource_record_set:
```
fridim: same question

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: Templates aren't that smart in Google, but I will look at the code in OSP to see if I can find a middle ground.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: There is an environment variable, I haven't used it with service accounts so I will test it out and convert after.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: I modeled after the Azure cloud provider so I didn't consider that, I believe I can move this into the template.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: same answer

vincepower: There is no clean way to include DNS records in the deployment manager template. I can include a third party script but then we would also need to branch that script and track changes to it. At least the way it is now it uses Ansible so its stateless.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: There is no clean way to include DNS records in the deployment manager template. I can include a third party script but then we would also need to branch that script and track changes to it. At least the way it is now it uses Ansible so its stateless.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: This has been completed, it now loads the authentication from environment variables and uses a config directory with guid in the name.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: I see how you are doing it with your heat templates and I'm not sure how'd I'd make that work with Google.
Google does create a private set of ssh keys which can be used to access the hosts but you need to connect via "gcloud compute ssh -z zone1b instance1" instead of just ssh from the command line. That would be a lot of modification to a lot of roles when we can just use this pattern which matches how Azure does it.
Luckily Alibaba works like AWS with stored keys. so I won't need to use this pattern for the next cloud I will likely enable.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: After more research, Google doesn't update DNS in templates anywhere its its sample library without using a custom python script which is something else we'd need to watch and version.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
fridim: Can we instead of having the key as a requirement, generate it for each deployment ?

IMO, the "infra" key for the environment should be per-deployment.

> Google does create a private set of ssh keys which can be used to access the hosts but you need to connect via "gcloud compute ssh -z zone1b instance1" instead of just ssh from the command line
I see.
We can generate the key in `output_dir` using basic `ssh-keygen` commands, see https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/infra-local-create-ssh_key/tasks/main.yml  

Regarding extra access to the environment, i would prefer if we just use regular authorized_keys like this:

https://github.com/redhat-cop/agnosticd/blob/85463267058cc276631aacbffb0b89935cae0427/ansible/roles/infra-osp-template-generate/defaults/main.yaml#L4-L6


That way:
1. we have an infra ssh key generated per-deployment
2. if we need to, we can add more access using the pub keys of admin

wdyt ?

ansible/configs/ocp4-cluster/files/cloud_providers/gcp_cloud_template.j2
```diff
@@ -0,0 +1,83 @@
...
+
+- name: {{ project_tag + '-vm' }}
+  type: compute.v1.instance
+  properties:
+    zone: {{ gcp_zone }}
+    machineType: zones/{{ gcp_zone }}/machineTypes/n1-standard-2
+    metadata:
+      items:
+        - key: ssh-keys
+          value: "gcpuser:{{ ssh_key_data }} userkey"
```
fridim: The extra ssh keys would go here i guess

ansible/cloud_providers/gcp_infrastructure_deployment.yml
fridim: OK, it's OK for now.

ansible/cloud_providers/gcp_infrastructure_deployment.yml
vincepower: We can add keys to the authorized_keys file, that's not a problem.

And I will build some logic towards having the infra_key always generated and also apply it to Azure over the next couple days. It can be done in agnosticD, just not in the Google template like you do with Heat.

vincepower: its a user:key parameter in the template file, so we can make a user account for for every admin key or add all the keys to a single user like opentlc-mgr.

vincepower: I have the admin keys all being added to the {{ remote_user }} authorized_keys file for Azure and GCP along with an auto generated ssh_key

vincepower: Added the public ssh keys to default_vars.yml so you can easily override it

vincepower:
> Can you create (and test) a requirements.txt file to work with google cloud ?
> 
> We save them in: https://github.com/redhat-cop/agnosticd/tree/development/tools/virtualenvs

I added one for Azure and another for GCP, and tested both a couple times with "python3 -m venv"

fridim:
You asked for Wolfgang's review so i'm waiting on him.
Let me know when ready to merge.

wkulhanek:
@fridim if you're happy with it go ahead and merge

PR #1251, opened 2020-03-06
Added role ocp4-workload-summit-api-cicd for API CICD Lab Env Set up

satyaj:
##### SUMMARY

This is an OCP4 workload role to be run on bastion as opentlc-mgr to install the 3scale, tekton and camel-k operators and required objects for the API Lifecycle CICD lab.

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME

ocp4-workload-summit-api-cicd
##### ADDITIONAL INFORMATION

```paste below

```


ansible/roles/ocp4-workload-summit-api-cicd/tasks/remove_workload.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: virtualenv path to use for pip
+  set_fact:
+    pip_path: "{{ ansible_python_interpreter | dirname }}"
+
+- name: "Uninstall lxml module from python"
+  command: "{{ pip_path }}/pip uninstall lxml -y"
+  become: True
+
+- name: "Remove {{ API_MANAGER_NS }} Project"
+  shell: "oc delete project {{API_MANAGER_NS}}"
```
wkulhanek: Would be better to do a k8s loop:

```
k8s:
  state: absent
  api_version: project.openshift.io/v1
  kind: Project
  name: "{{ item }}"
loop:
- "{{ DEV_GW_PROJECT }}"
- ....
```

ansible/roles/ocp4-workload-summit-api-cicd/templates/global-operators.j2
```diff
@@ -0,0 +1,8 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  annotations:
+    olm.providedAPIs: Build.v1.camel.apache.org,CamelCatalog.v1.camel.apache.org,Config.v1alpha1.operator.tekton.dev,Integration.v1.camel.apache.org,IntegrationKit.v1.camel.apache.org,IntegrationPlatform.v1.camel.apache.org
+  name: global-operators
+  namespace: openshift-operators
```
wkulhanek: Hm. Will this work if another global-operators operator group is already there?

ansible/roles/ocp4-workload-summit-api-cicd/tasks/remove_workload.yml
satyaj: Thanks @wkulhanek . I changed this to use k8s. Will send another pull request to address it.

ansible/roles/ocp4-workload-summit-api-cicd/templates/global-operators.j2
satyaj: Yes, this assumes no other operators exist. Is taking an approach of using CRD and oc create a better approach?

wkulhanek: I think as long as you're the only workload on that cluster... it may be problematic on a shared cluster (like RHPDS)

PR #1270, opened 2020-03-09
Fix workloads

wkulhanek:
Updated workloads:
ocp4-workload-infra-nodes:
- Added wait step to make sure the nodes are up before continuing
- Enabled Taints and Tolerations
- Enabled creation of dedicated Elasticsearch Nodes
- Actually move workloads to infranodes
- Add workaround for bug where some DaemonSets don't have universal tolerations

ocp4-workload-logging:
- Added check if Elasticsearch is already running
- Added logic to retrieve the channel for Elasticsearch and Logging
- Fixed logic to wait for operators to be alive
- Added support for Taints and Tolerations
- Added support to deploy the logging stack to Infra Nodes or dedicate Elasticsearch Nodes


ansible/roles/ocp4-workload-infra-nodes/defaults/main.yml
```diff
@@ -3,5 +3,11 @@ become_override: False
 ocp_username: opentlc-mgr
 silent: False
 
-_infra_node_replicas: 3
-_infra_node_instance_type: m4.4xlarge
\ No newline at end of file
+_infra_node_replicas: 1
```
stencell: Is this going to cause a problem with the number of replicas needed for some of the infra components?

wkulhanek: You can always make it bigger... Also we can move logging to dedicated nodes now.

wkulhanek: You can also make the instance types bigger. These are just defaults.

ansible/roles/ocp4-workload-logging/defaults/main.yml
```diff
@@ -3,8 +3,14 @@ become_override: False
 ocp_username: opentlc-mgr
 silent: False
 
-_logging_elasticsearch_replicas: 3
-_logging_elasticsearch_memory_request: "4Gi"
+_logging_elasticsearch_replicas: 1
+_logging_elasticsearch_memory_request: "8Gi"
 _logging_elasticsearch_storage_request: "200Gi"
```
stencell: Can and should we make this smaller...and get curator configured to purge after like 3 or 5 days? Assuming curator works.

ansible/roles/ocp4-workload-logging/files/eo_namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-operators-redhat
```
stencell: I know the docs say this, but I think we have been putting this in openshift-operators (where it lands using UI to install as well), which is default namespace for cluster wide. I don't know if that is a docs bug or what, but maybe we could standardize on the "right" namespace.

ansible/roles/ocp4-workload-logging/defaults/main.yml
wkulhanek: Curator is configured quite conservatively. Like 2 days. There's a curator config map

ansible/roles/ocp4-workload-logging/files/eo_namespace.yaml
wkulhanek: I'm pretty sure that would make managing the operator group a nightmare.

stencell: I don't think so...there is already an operatorgroup in place for the openshift-operators namespace that handles this for global/cluster wide operators

```
[nstephan@bastion ~]$ oc get operatorgroup -A
NAMESPACE                              NAME                           AGE
openshift-monitoring                   openshift-cluster-monitoring   133m
openshift-operator-lifecycle-manager   olm-operators                  133m
openshift-operators                    global-operators               133m
[nstephan@bastion ~]$ oc get operatorgroup global-operators -n openshift-operators
NAME               AGE
global-operators   133m
[nstephan@bastion ~]$ oc get operatorgroup global-operators -n openshift-operators -oyaml
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  creationTimestamp: "2020-03-09T18:56:52Z"
  generation: 1
  name: global-operators
  namespace: openshift-operators
  resourceVersion: "4965"
  selfLink: /apis/operators.coreos.com/v1/namespaces/openshift-operators/operatorgroups/global-operators
  uid: bc9f3eb5-b6c6-4244-a845-5e23f91b7fff
status:
  lastUpdated: "2020-03-09T19:02:01Z"
  namespaces:
  - ""
```

wkulhanek: I'll double check with the engineering lead. Depending on what he says I'll update the workload once again - or not. :-)

PR #1274, opened 2020-03-10
OCP4 NFS workload role

stencell:
##### SUMMARY
This role can be used to deploy an NFS server pod inside an OpenShift cluster. It is helpful when you need more than RWO that is provided with AWS & OpenStack.

A PV will be provisioned for the NFS server pod
A new storage class will be created
A project and associated objects will be created to run the NFS server pod
A statefulset will be created to deploy the NFS server pod
The script that runs will set up a finite number of exports (to be variablized)
PVs will be created

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
ocp4-workload-nfs-server

### DIFF
--- a/ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
@@ -6,7 +6,7 @@
       apiVersion: storage.k8s.io/v1
       kind: StorageClass
       metadata:
-        name: manual
+        name: "{{ _nfs_provided_storage_class }}"
       provisioner: kubernetes.io/no-provisioner
       reclaimPolicy: Retain
       volumeBindingMode: WaitForFirstConsumer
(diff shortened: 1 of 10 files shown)


ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
```diff
@@ -0,0 +1,202 @@
...
+    state: present
+    definition:
+      apiVersion: v1
+      kind: ServiceAccount
+      metadata:
+        name: "{{ _nfs_app }}"
+        namespace: "{{ _nfs_project }}"
+
+- name: Add anyuid SCC to SA
+  shell: oc adm policy add-scc-to-user anyuid system:serviceaccount:{{ _nfs_project }}:{{ _nfs_app }}
```
fridim: `s/shell/command`

ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
```diff
@@ -0,0 +1,202 @@
...
+      kind: ImageStream
+      metadata:
+        labels:
+          app: "{{ _nfs_app }}"
+        name: "{{ _nfs_app }}"
+        namespace: "{{ _nfs_project }}"
+      spec: {}
+
+- name: Import image to ImageStream
+  shell: oc import-image --from={{ _nfs_server_image }} --scheduled {{ _nfs_app }} -n {{ _nfs_project }}
```
fridim: `s/shell/command`

ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
```diff
@@ -0,0 +1,202 @@
...
+      kind: ServiceAccount
+      metadata:
+        name: "{{ _nfs_app }}"
+        namespace: "{{ _nfs_project }}"
+
+- name: Add anyuid SCC to SA
+  shell: oc adm policy add-scc-to-user anyuid system:serviceaccount:{{ _nfs_project }}:{{ _nfs_app }}
+
+- name: Add privileged SCC to SA
+  shell: oc adm policy add-scc-to-user privileged system:serviceaccount:{{ _nfs_project }}:{{ _nfs_app }}
```
fridim: `s/shell/command`

ansible/roles/ocp4-workload-nfs-server/tasks/workload.yml
stencell: Why do you want shell instead of command? I have never heard a firm reason one way or the other.

fridim: Because sometimes you just don't need a shell to run something.

you usually use `shell`  when you need features from the shell like:

- pipes
- variables
- `&`, `if/else`
- stream operation `>` `<`
- some variables like $HOME
- ...

If you just run a simple command like here, i'd prefer command.
I assume it's faster, and probably safer: shell is another potentially "tainted" environment that could go wrong.

wkulhanek: You could also just use k8s to create the role binding

PR #1289, opened 2020-03-10
Satellite role - create organization through foreman-ansible-modules

ezr-ondrej:
##### SUMMARY
This is switching the organization creation for satellite to be done through 'foreman-ansible-modules'

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
satellite-manage-organization

##### ADDITIONAL INFORMATION
The satellite team provides sets of ansible modules. It only makes sense to use them. This is a first PR with this change, others will follow after we set course in this one.

The PR needs to have the collection of foreman modules installed locally with it's dependencies.

I've opened PR for them to get taken care of automatically
- [ ] Dependencies - #1288
- [ ] Collection installation - #1282

###### To test manually in meantime

```bash
pip install apypie
ansible-galaxy collection install theforeman.foreman
```

### DIFF
--- a/tests/static/syntax-check.sh
+++ b/tests/static/syntax-check.sh
@@ -38,17 +38,18 @@ for i in \
         )
     fi
 
-    if [ "${env_type}" = satellite-vm ] || [ "${env_type}" = satellite-multi-region ]; then
-        pip install apypie
-        ansible-galaxy collection install theforeman.foreman
-    fi
-
-
     if [ -e "${ansible_path}/configs/${env_type}/hosts" ]; then
         inventory=(-i "${ansible_path}/configs/${env_type}/hosts")
     else
         inventory=(-i "${static}/tox-inventory.txt")
     fi
+
+    # Setup galaxy roles and collections, make sure it works
+    ansible-playbook --tags galaxy_roles \
+                     "${inventory[@]}" \
+                     ${ansible_path}/main.yml \
+                     ${extra_args[@]} \
+                     -e @${i}
 
     for playbook in \
         ${ansible_path}/main.yml \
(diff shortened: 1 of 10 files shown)


wkulhanek:
@ezr-ondrej Travis checks failed. Please fix.

tests/static/syntax-check.sh
```diff
@@ -38,6 +38,11 @@ for i in \
         )
     fi
 
+    if [ "${env_type}" = satellite-vm ] || [ "${env_type}" = satellite-multi-region ]; then
+        pip install apypie
+        ansible-galaxy collection install theforeman.foreman
+    fi
```
ezr-ondrej: @wkulhanek is this correct way to install dependencies?
I've got #1282 and #1288 to fix it for users, but it's not going to help with tests anyway.

wkulhanek: @fridim ?

fridim: I think it's ok to add it to the requirement https://github.com/redhat-cop/agnosticd/blob/development/tests/static/requirements.txt 

It sounds more maintainable than introducing `if/else` in the test script.

i approved and merged #1288

fridim: that was for the python dependencies.

For the ansible role dependencies, we need to find a more generic way.

Maybe we run agnosticd with `--tags step0000` so the galaxy roles are installed.

Let me try something

fridim: @ezr-ondrej i created a PR for you to update this one.

Let me know what you think

ezr-ondrej:
Thanks @fridim, we are :green_heart: now. I've added your changes and added `apipy` to test `requirements.txt` to note it'd be necessary.

tonykay:
After your update happy to merge, thanks

PR #1292, opened 2020-03-11
Enable Infra Nodes (and Machine Autoscaling) for OpenStack

wkulhanek:
Much simpler logic because GPTE OpenStack has no regions or availability zones...


ansible/configs/ocp4-cluster/destroy_env.yml
```diff
@@ -17,7 +17,7 @@
     - name: Remove DNS entry for OpenShift API and ingress
       nsupdate:
         server: "{{ osp_cluster_dns_server }}"
-        zone: "{{ cluster_dns_zone }}"
+        zone: "{{ osp_cluster_dns_zone }}"
```
stencell: where are you setting this? I don't see it in default_vars?

wkulhanek: Secret vars

wkulhanek: Since it depends on the cluster

PR #1293, opened 2020-03-11
Satellite subscriptions management

ezr-ondrej:
##### SUMMARY
Improve roles that takes care of subscriptions and repositories in satellite.

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
Updates `satellite-manage-manifest`
Roles `satellite-manage-subscriptions` and `satellite-manage-sync` got merged into `satellite-manage-repositories`

##### ADDITIONAL INFORMATION
1. Add ability to get manifest from url
1. Merge satellite roles to manage subscriptions and synchronizations as both roles take care of repositories, it makes no sense to split them into two especially once they only consists from three tasks.


ansible/roles/satellite-manage-manifest/tasks/version_6.x.yml
```diff
@@ -0,0 +1,36 @@
...
+  register: manifest_file_existance_result
+
+- name: Copy manifest
+  copy:
+    src: "{{ manifest_file }}"
+    dest: /tmp
+  when:
+    - not manifest_file_existance_result.exists
+    - "not manifest_file.startswith('http')"
+    - "not manifest_file.startswith('ftp')"
```
ezr-ondrej: Very nasty check whether the file is local or not :thinking:
Any suggestions how to improve it? :upside_down_face:

.gitignore
```diff
@@ -22,6 +22,7 @@ virtualenv/
 *.rc
 .tox
 ansible/configs/*/roles
+ansible/configs/*/collections
```
fridim: i believe it is `ansible/collections`

ansible/roles/satellite-manage-manifest/tasks/main.yml
```diff
@@ -1,8 +1,8 @@
 ---
 
 ## Import for version satellite 6.4 ##
-- import_tasks: version_6.4.yml
-  when: satellite_version == 6.4
+- import_tasks: version_6.x.yml
+  when: satellite_version >= 6.4
```
fridim: This won't work. Ansible will convert to float. For example:

`6.10 >= 6.4` is false

There is a special jinja2 filter for that.

```
when: satellite_version is version_compare('6.4', '>=')
```
You test work

ansible/roles/satellite-manage-manifest/tasks/version_6.x.yml
fridim: That test is not necessary.

ansible module `copy` will not copy if the files already exist and is the same checksum or content.

ansible/roles/satellite-manage-manifest/tasks/version_6.x.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name:
+  get_url:
+    url: "{{ manifest_file }}"
+    dest: /tmp
+  when:
+    - not manifest_file_existance_result.stat.exists
+    - "manifest_file.startswith('http') or manifest_file.startswith('ftp')"
```
fridim: `manifest_file`  suggests it's a filename or file content.

I recommend you add another variable, `manifest_url` or something.
Using the same for both file and url is errorprone.

You can decide what method to use depending on the existence of the variables, or simply add a `manifest_method`.

You can also use a dictionary

```yaml
manifest_info:
  type: file
  content: "{{ ... }}"

manifest_info:
  type: url
  url: http://...
```

And do things accordingly, depending of the value of `manifest_info.type`

ansible/roles/satellite-manage-repositories/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+## Import for version satellite 6.4 ##
+- import_tasks: version_6.x.yml
+  when: satellite_version >= 6.4
```
fridim: same as above, use `version_compare()`

ansible/roles/satellite-manage-repositories/tasks/version_6.x.yml
```diff
@@ -0,0 +1,51 @@
...
+
+- name: Wait until all Syncs have finished
+  async_status:
+    jid: "{{ repo_sync_sleeper_item.ansible_job_id }}"
+  loop: "{{ repo_sync_sleeper.results }}"
+  loop_control:
+    loop_var: repo_sync_sleeper_item
+  when: repo_sync_sleeper_item.ansible_job_id is defined  # Skip items that were skipped in the previous task
+  register: async_job_result
+  until: async_job_result.finished
```
fridim: Why do you detach ?

Usually in ansible we detach if we plan to do other tasks in between and join later to wait for the background tasks.

Here you can just run with async (because it's a very long task, so it's required) but without the extra wait by not specifying `poll: 0`, or specify `poll: 60` or whatever reasonable value.


Please use a readable format for async timeout:

```
          async: "{{ 2 * 60 * 60 }}"

```

ezr-ondrej: Here I detach, because I want to sync in parallel and then have a semaphore to wait.
If I cycle the async, I'll sync the repositories in serial, right?

ansible/roles/satellite-manage-repositories/tasks/main.yml
ezr-ondrej: Awesome, thanks! :+1:

ansible/roles/satellite-manage-repositories/tasks/version_6.x.yml
fridim: ah i missed that, good!

.gitignore
ezr-ondrej: for me, it clones them to the config directory :thinking:

ansible/roles/satellite-manage-manifest/tasks/version_6.x.yml
ezr-ondrej: Yeah, that's a nice idea :+1:

.gitignore
fridim: i created a PR to fix what we did:
https://github.com/redhat-cop/agnosticd/pull/1295

See especially:
https://github.com/redhat-cop/agnosticd/pull/1295/commits/2962a931ea54e9136399746f6591d69a65e8b802

ezr-ondrej: Ok, droping :+1:

ezr-ondrej:
I've updated the manifest variable and the manifest role accordingly. I've also updated docs that I forgotten before.

ezr-ondrej:
@fridim should I split it up into two PRs?

PR #1296, opened 2020-03-11
Separate infra roles create new role dir 'roles-infra'

fridim:
Prefer 'roles-infra' over the more natural 'infra-roles'  so it appears when
using tab completion when developing.

In the future, we can have other separations like for example:

- ocp workloads, 'roles-ocp-workloads'

##### SUMMARY
- Feature Pull Request


ezr-ondrej:
I'm lovin' it @fridim! :+1:  It bugs me all the time, we have them together.

stencell:
This seems like it was a pretty big set of changes to shotgun merge. Was there some cursory testing done on a handful of these to make sure there wasn't any unexpected results?

tonykay:
So I read through the initial comment and did not see a WIP. We need a pattern for WIPs so this is extremely clear. Do we actually need a real development branch and a real master?

fridim:
We reverted the PR, something was wrong.
I'll create another one later.

stencell:
@tonykay Sorry, but the lack of a WIP as any justification for this is pretty weak. Yes, it should have probably had WIP. Just because it does not have that doesn't mean it should be merged this quickly, though. This still happens too often in this repo (and others) by people that don't really know what they are merging IMO. We can use common sense and courtesy when deciding whether or not to merge. No one should get a free pass or less of a look than others. Ask @wkulhanek how annoying it is to him that I read every one of his PRs before approving it...usually to find someone else came in and robo-approved it while I was reading and commenting. I may not be the fastest, but no one reviews code as fast as I see some of these going in.

Also merging and approval are different. If you approve of this PR, go ahead and stamp it with your approval. Doesn't mean you have to merge it. Let the person who opened the PR own that responsibility unless you are eminently comfortable with the change and there has been sufficient time for others to weigh in.

In this case, there were 4 reviewers requested. If someone actual requests reviewers, they should be given a chance to...review. Quadruply so for a change like this that touches over 100 files over quite a few products that AgnosticD touches and manages.

If a PR comes in from someone for Ansible, I will sometimes approve if I understand the changes being made and they seem minor, but I won't typically merge it unless very specifically asked. I have no business doing that. Same with most Middleware unless it is a very seemingly simple change.

Just because we all have permissions on this mono-repo, doesn't mean we have the duty or capability to approve and merge anything without a bit more thought and testing.

fridim:
Thanks @stencell for taking the time. I think Tony's aligned, we even discussed that very specific issue ealier yesterday.

i created a PR, feel free to jump in there: https://github.com/redhat-cop/agnosticd/pull/1306

PR #1298, opened 2020-03-11
include_role for infra roles

agonzalezrh:

### DIFF
--- a/ansible/configs/osp-migration/infra.yml
+++ b/ansible/configs/osp-migration/infra.yml
@@ -185,13 +185,13 @@
     # Use  server.metadata.hostname  instead
     - name: Run infra-osp-create-inventory Role
       include_role:
-        name: infra-osp-create-inventory
+        name: "{{ANSIBLE_REPO_PATH}}/roles-infra/infra-osp-create-inventory"
       vars:
         _name_selector: 'metadata.hostname'
 
     - name: Run Common SSH Config Generator Role
-      import_role:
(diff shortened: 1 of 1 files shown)


ansible/configs/osp-migration/infra.yml
```diff
@@ -185,13 +185,13 @@
     # Use  server.metadata.hostname  instead
     - name: Run infra-osp-create-inventory Role
       include_role:
-        name: infra-osp-create-inventory
+        name: "{{ANSIBLE_REPO_PATH}}/roles-infra/infra-osp-create-inventory"
```
wkulhanek: Do we actually need the ANSIBLE_REPO_PATH ? Isn't that in the search path anyway?

agonzalezrh: I copied from here:

./configs/linklight-foundations/post_infra.yml:      name: "{{ANSIBLE_REPO_PATH}}/roles-infra/infra-common-ssh-config-generate"

Maybe you are right and we don't needed it

wkulhanek: @fridim  ?

agonzalezrh: I modified it now, we can try without the path

ezr-ondrej:
Hi, just wondering, wouldn't this be enough: #1299

wkulhanek:
Right. That was my point. Should be in the "search path"

PR #1309, opened 2020-03-11
[WIP] New way of doing role inputs

wkulhanek:
New Role: ocp4-workload-machinesets
New way of specifying input to roles:
- Role has a dictionary: <role>_defaults
- Additional inputs from AgnosticV (or elsewhere) comes as a dictionary: <role>_input
- Secret variables can come from a dictionary: <role>_secret (so far used only in authentication role for LDAP bind password).

### DIFF
--- a/ansible/roles/ocp4-workload-authentication/OWNERS
+++ b/ansible/roles/ocp4-workload-authentication/OWNERS
@@ -1,4 +1,0 @@
-Red Hat GPTE
-
-Author(s):
-- Wolfgang Kulhanek  (wkulhane@redhat.com)
--- a/ansible/roles/ocp4-workload-authentication/OWNERS
+++ b/ansible/roles/ocp4-workload-authentication/OWNERS
@@ -1,0 +1,4 @@
+Red Hat GPTE
+
+Author(s):
+- Wolfgang Kulhanek  (wkulhane@redhat.com)
--- a/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
@@ -4,10 +4,11 @@
     ocp4_workload_authentication: >-
       {{ ocp4_workload_authentication_defaults
        | combine(ocp4_workload_authentication_input  | default( {} ),
-                 ocp4_workload_authentication_secret | default( {} ))
+                 ocp4_workload_authentication_secret | default( {} ), recursive=true)
       }}
 - name: Print combined role variables
   debug: var=ocp4_workload_authentication
+  verbosity: 2
 
 - name: Check that ocp4_workload_authentication.idm_type is defined
   when:
--- a/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
@@ -10,11 +10,12 @@
   debug: var=ocp4_workload_authentication
   verbosity: 2
 
-- name: Check that ocp4_workload_authentication.idm_type is defined
-  when:
-  - ocp4_workload_authentication.idm_type is not defined
-  fail:
-    msg: "ocp4_workload_authentication.idm_type is not defined"
+- name: Check that ocp4_workload_authentication.idm_type is defined and valid
+  assert:
+    that:
+    - ocp4_workload_authentication.idm_type is defined
+    - ocp4_workload_authentication.idm_type == "none" or ocp4_workload_authentication.idm_type == "htpasswd" or ocp4_workload_authentication.idm_type == "ldap"
+    fail_msg: "ocp4_workload_authentication.idm_type is not defined or not in ('none', 'htpasswd', 'ldap')."
 
 - name: Setup HTPasswd Authentication
   when: ocp4_workload_authentication.idm_type == "htpasswd"
@@ -61,7 +62,6 @@
     when:
     - r_htpasswd_line is succeeded
 
-  # Generate htpasswd file
   - name: Generate htpasswd file
     template:
       src: "htpasswd.j2"
@@ -83,9 +83,6 @@
   - name: Update OAuth Configuration
     k8s:
       state: present
-      merge_type:
-      - strategic-merge
-      - merge
       definition: "{{ lookup('file', item ) | from_yaml }}"
     loop:
     - ./files/oauth-htpasswd.yaml
--- a/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-authentication/tasks/workload.yml
@@ -1,48 +1,64 @@
 ---
-# Implement your Workload deployment tasks here
+- name: Set up ocp4_workload_authentication combined dictionary
+  set_fact:
+    ocp4_workload_authentication: >-
+      {{ ocp4_workload_authentication_defaults
+       | combine(ocp4_workload_authentication_input  | default( {} ),
+                 ocp4_workload_authentication_secret | default( {} ))
+      }}
+- name: Print combined role variables
+  debug: var=ocp4_workload_authentication
 
-- name: Check that ocp4_idm_install is defined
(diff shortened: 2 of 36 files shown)


wkulhanek:
I consider this finished. Only WIP because I want to run a final test end to end on AWS and OpenStack.

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
```diff
@@ -1,48 +1,64 @@
...
+    msg: "ocp4_workload_authentication.idm_type is not defined"
 
 - name: Setup HTPasswd Authentication
-  when: ocp4_idm_install == "htpasswd"
+  when: ocp4_workload_authentication.idm_type == "htpasswd"
   block:
-  # Generate Passwords if no passwords specified
   - name: Generate cluster admin password
-    when: ocp4_idm_htpasswd_admin_password | d('') | length == 0
+    when: ocp4_workload_authentication.htpasswd_admin_password | d('') | length == 0
```
stencell: In your defaults for the role, this is set to `openshift_admin`, so wouldn't this always evaluate false and never generate?

stencell: Maybe I'll read ahead next time.

ansible/roles/ocp4-workload-authentication/OWNERS
```diff
@@ -0,0 +1,4 @@
+Red Hat GPTE
+
+Author(s):
+- Wolfgang Kulhanek  (wkulhane@redhat.com)
```
fridim: That's a good idea.

Can we use a format like yaml ?

What about the standard ansible `meta/main.yml` ? https://galaxy.ansible.com/docs/contributing/creating_role.html#role-metadata

We usually don't target galaxy, but we could take inspiration from it ?

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
```diff
@@ -1,48 +1,64 @@
 ---
-# Implement your Workload deployment tasks here
-
-- name: Check that ocp4_idm_install is defined
+- name: Set up ocp4_workload_authentication combined dictionary
+  set_fact:
+    ocp4_workload_authentication: >-
+      {{ ocp4_workload_authentication_defaults
+       | combine(ocp4_workload_authentication_input  | default( {} ),
+                 ocp4_workload_authentication_secret | default( {} ))
```
fridim: Don't you want the combine to be recursive: `combine( A, B, recursive=True)` ?

> The filter also accepts an optional recursive=True parameter to not only override keys in the first hash, but also recurse into nested hashes and merge their keys too

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
```diff
@@ -1,48 +1,64 @@
...
+      }}
+- name: Print combined role variables
+  debug: var=ocp4_workload_authentication
```
fridim: i would remove that debug or add `verbosity: 2` to it, since the secrets are merged into that variable too.

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
```diff
@@ -1,48 +1,64 @@
...
   when:
-  - ocp4_idm_install is not defined
+  - ocp4_workload_authentication.idm_type is not defined
   fail:
-    msg: "ocp4_idm_install is not defined"
+    msg: "ocp4_workload_authentication.idm_type is not defined"
 
 - name: Setup HTPasswd Authentication
-  when: ocp4_idm_install == "htpasswd"
+  when: ocp4_workload_authentication.idm_type == "htpasswd"
```
fridim: FYI i find this way convenient for managing assertions / checks: https://github.com/redhat-gpte-devopsautomation/babylon_aws_sandbox/blob/master/tasks/pre_checks.yaml

ansible/roles/ocp4-workload-logging/tasks/workload.yml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Set up ocp4_workload_logging combined dictionary
+  set_fact:
+    ocp4_workload_logging: >-
+      {{ ocp4_workload_logging_defaults
+       | combine(ocp4_workload_logging_input  | default( {} ),
+                 ocp4_workload_logging_secret | default( {} ))
```
fridim: `recursive=True` ?

ansible/roles/ocp4-workload-logging/tasks/workload.yml
```diff
@@ -1,4 +1,14 @@
...
+      }}
+- name: Print combined role variables
+  debug: var=ocp4_workload_logging
```
fridim: same, don't print passwords or add `verbosity`

ansible/roles/ocp4-workload-machinesets/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+# Do not modify this file
+
+- name: Running Pre Workload Tasks
+  include_tasks:
+    file: ./pre_workload.yml
```
fridim: you don't need the `./`

ansible/roles/ocp4-workload-machinesets/tasks/workload.yml
```diff
@@ -0,0 +1,71 @@
+---
+- name: Set up ocp4_workload_machinesets combined dictionary
+  set_fact:
+    ocp4_workload_machinesets: >-
+      {{ ocp4_workload_machinesets_defaults
+       | combine(ocp4_workload_machinesets_input  | default( {} ),
+                 ocp4_workload_machinesets_secret | default( {} ))
```
fridim: same question, `recursive=True` ?

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
tonykay: Good point, to me this should be the default behavior but isn't. I agree `recursive=True`

ansible/roles/ocp4-workload-authentication/OWNERS
wkulhanek: Yeah. I'll just create a meta/main in these roles instead of OWNER.

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
wkulhanek: Good point. Done.

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
wkulhanek: Makes sense. Done.

ansible/roles/ocp4-workload-logging/tasks/workload.yml
wkulhanek: Done.

ansible/roles/ocp4-workload-logging/tasks/workload.yml
wkulhanek: Done

ansible/roles/ocp4-workload-machinesets/tasks/main.yml
wkulhanek: @fridim Those are in every single workload....

ansible/roles/ocp4-workload-machinesets/tasks/workload.yml
wkulhanek: Done

ansible/roles/ocp4-workload-machinesets/tasks/main.yml
fridim: OK we can mass-fix later then.

ansible/roles/ocp4-workload-authentication/tasks/workload.yml
wkulhanek: Changed it to assert.

PR #1322, opened 2020-03-13
[WIP] RHEL8 clientvm

stencell:
##### SUMMARY
It is a good divergence point between RHEL 7 and RHEL 8 to start with a new config for the RHEL 8 client VM. Nothing changes here in the infra phases, all is software phases and relatively minor in this pass. Since @wkulhanek cleaned some of this up somewhat recently, it is not as messy as it might have otherwise been.

There are a few common roles that I touched that I would like another set of eyes on @fridim .

TODO: 
Update the readme
A bit more cleanup of unused tasks and files
Reorder a few tasks
Cleanup default vars files

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
clientvm-rhel8
common
set-repositories
ocp-client-vm
bastion-lite


ansible/configs/ocp-clientvm-rhel8/default_vars_ec2.yml
```diff
@@ -0,0 +1,91 @@
+### AWS EC2 Environment settings
```
stencell: I'm not sure why we don't have `cloud_provider` set in here. I thought it was supposed to be set and I've seen it set to both `ec2` and `aws`. Should it be set and which is the correct one?

fridim: Because `cloud_provider`, along with `env_type` must always be passed as extra vars. They are the top-level / mandatory / static variables.

If you look at `ansible/main.yml` you'll see things like

```yaml
- import_playbook: "./configs/{{ env_type }}/pre_infra.yml"
```

And since `import_playbook` is static and imported and "compiled" in the early stages of runtime you cannot define `env_type` in a file that is dynamically included with an `include_vars` task.

Also, `cloud_provider` must be defined before the file you're mentioning is included: `default_vars_`**`ec2`**`.yml`.  That's dynamically included via `include_vars` and `default_vars_{{ cloud_provider }}.yml`.

ansible/configs/ocp-clientvm-rhel8/files/cloud_providers/ec2_cloud_template.j2
```diff
@@ -0,0 +1,356 @@
+#jinja2: lstrip_blocks: True
```
fridim: for something as simple as clientvm, do we really need this template file ?
Can we use the default template instead ?

ansible/configs/ocp-clientvm-rhel8/software.yml
```diff
@@ -0,0 +1,28 @@
...
+        msg: "Software tasks started"
+
+- name: Set up ClientVM
+  hosts: bastions
+  gather_facts: false
+  become: true
+  tasks:
+  - name: Set up Client VM for OCP 3
+    include_role:
+      name: "ocp-client-vm"
```
fridim: If you don't specify a condition, it means it will always be included, so you can simply use `import_role` here.
That way it's included in the syntax-check tests too.

ansible/roles/common/tasks/packages_el8.yml
```diff
@@ -0,0 +1,14 @@
+#vim: set ft=ansible:
+---
+######################### Install Basic Packages
+- name: install common packages for RHEL 8
+  yum:
```
fridim: Does `yum` work ?

I thought we would have to switch to `package` and let ansible discover the package manager.

ansible/roles/set-repositories/tasks/main.yml
```diff
@@ -18,3 +18,7 @@
     - name: Configure Repository File
       import_tasks: ./file-repos.yml
```
fridim: prefer `include_tasks` here so the tasks are not imported.

`import_tasks` + a false `when` ==  all tasks are imported and the condition is copied to all the imported tasks == logs are full of `skipped` tasks.

`include_tasks` + a false `when` ==  the include is skipped, nothing is included.

ansible/roles/set-repositories/tasks/main.yml
```diff
@@ -18,3 +18,7 @@
...
       when: repo_method == "file"
+
+    - name: Configure Repository File
+      import_tasks: ./satellite-cert.yml
```
fridim: same, `include_tasks`

fridim: Why don't you update the satellite-repo.yml  and add a condition with `satellite_cert` there ?
`satellite-cert.yml` and `satellite-repo.yml` look very similar

ansible/configs/ocp-clientvm-rhel8/default_vars_ec2.yml
stencell: So my question was...why don't we have `cloud_provider` defined in this `default_vars_ec2` file? And how does this work without it? Because it does work. And why is it `ec2` instead of `aws`?

stencell: Nevermind me on why does it work...it was in my vars file that I pass in.

Back to my original question, should we have `cloud_provider: ec2` set as a default in this file ?

fridim: `cloud_provider` and `env_type` are always passed as extra vars.

So to reply to your question:
> Back to my original question, should we have cloud_provider: ec2 set as a default in this file ?

No, it would not have any effect and would create confusion i think.

ansible/configs/ocp-clientvm-rhel8/files/cloud_providers/ec2_cloud_template.j2
stencell: I agree. Change pushed.

ansible/configs/ocp-clientvm-rhel8/software.yml
stencell: Is there a benefit of include vs import ?

ansible/roles/common/tasks/packages_el8.yml
stencell: Yum works fine, both ansible module and command line even though it is dnf behind the scenese. We really don't do anything other than RHEL so I'm inclined to leave it unless you think the `package` module is better than it used to be.

ansible/configs/ocp-clientvm-rhel8/default_vars_ec2.yml
stencell: It would create confusion to have the `cloud_provider` set in the default vars file for the...cloud provider? I would think that makes it more user friendly so it is one less thing a user has to feed in as an extra var.

ansible/roles/set-repositories/tasks/main.yml
stencell: These have been merged into `satellite-repo.yml`

ansible/roles/set-repositories/tasks/main.yml
stencell: These have been switched...thanks for the tip.

ansible/roles/common/tasks/packages_el8.yml
stencell: Since this is a common role, literally, I am leaving this since all of the new functionality has been folded into the existing config. I don't want to risk breaking something right now.

ansible/configs/ocp-clientvm-rhel8/software.yml
stencell: changed.

ansible/configs/ocp-clientvm-rhel8/README.adoc
```diff
@@ -0,0 +1,119 @@
...
+----
+# Generic Vars
+
+GUID=sborenstest5
+REGION=ap-southeast-2
+KEYNAME=ocpkey
+ENVTYPE="ocp-clientvm"
+CLOUDPROVIDER=ec2
+HOSTZONEID='Z3IHLWJZOU9SRT'
+BASESUFFIX='.example.opentlc.com'
```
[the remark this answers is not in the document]
stencell: I agree it should be changed...I am just not going to tackle this update now.

stencell:
Quick update comment:

All of this was merged into the existing clientvm role rather than have a separate.

PR #1334, opened 2020-03-13
Fix scc cleanup (Quay Operator Bug)

wkulhanek:
Work around Quay Operator Bug


wkulhanek:
@tonykay lol. Unfortunately not. :-( And really the operator needs to clean this up. AND there seems to be a bug in the product..... Argh.

PR #1366, opened 2020-03-18
Update pipelines workload

wkulhanek:


ansible/roles/ocp4-workload-pipelines/tasks/remove_workload.yml
```diff
@@ -0,0 +1,69 @@
...
+  - ./templates/subscription.j2
+
+- name: Remove openshift-pipelines project
+  k8s:
+    state: absent
+    api_version: project.openshift.io/v1
+    kind: Project
+    name: openshift-pipelines
+
+- name: Remove tkn and tkn bash completion
```
stencell: Not that it is hurting, but is this necessary to remove?

wkulhanek: Well. The workload installs it. So I feel removing the workload should remove it.

wkulhanek: Hm. Maybe CRDs need cleaning up, too...

ansible/roles/ocp4-workload-pipelines/tasks/remove_workload.yml
```diff
@@ -0,0 +1,69 @@
...
+    api_version: operators.coreos.com/v1alpha1
+    kind: Subscription
+    name: openshift-pipelines-operator
+    namespace: openshift-operators
+  register: r_subscription
+
+- name: Print CSV
+  debug:
+    msg: "CSV: {{ r_subscription }}"
+- name: Remove CSV
```
stencell: formatting nit: add line above

Also, do you need to remove the CSV or will removing the subscription take care of this also?

wkulhanek: Need to. Removing just the subscription leaves the CSV.

PR #1403, opened 2020-03-25
infra-ec2-template-generate: upgrades the strong, downgrades the weak

newgoliath:
##### SUMMARY

* save $
* reduce deployment failures 
* reduce support calls
* reduce complexity in agnosticV settings
* tested against ocp-clientvm

- test aws account access to images, upgrade and downgrade as necessary.

###### Process

- for all instances and the aws account
- validate images that begin with RHEL against aws_ami_reigon_mapping
- validate against AWS bumping configs that use RHEL<VERSION> images to RHEL<VERSION>GOLD
- change successful bumps
- degrade RHEL<VERIONS>GOLD failures to RHEL<VERSION>
- validate all images
- continue CF template generation

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
infra-ec2-template-generate

###### To test

Emit a CF tempalte only - no need to deploy env

1. create launch script:
```
cd ~/agnosticv/
GUID=test_clientvm
./merge.sh gpte OCP_CLIENTVM dev > $HOME/ocp_clientvm.yaml                                                                                                                                                                            
cd ~/agnosticd/
mkdir ~/output/
ansible-playbook -vvvv ./ansible/main.yml  -e @${HOME}/ocp_clientvm.yaml -e @${HOME}/secrets/secret.yaml -e guid=$GUID -e output_dir=${HOME}/output/$GUID/ -e debug_template=true
```
1. get the GPTE aws config/credentails files with accounts `default` and `infra`
1. run config as `default` aws account (full access to GOLD images). `export AWS_PROFILE=default`
1. change agnosticv `gpte` `OCP_CLIENTVM` dev.yaml: `clientvm_instance_image: RHEL77GOLD`
1. run script
1. observe CF template is OK with RHEL77GOLD
1. change agnosticv `gpte` `OCP_CLIENTVM` dev.yaml: `clientvm_instance_image: RHEL77`
1. run script
1. observe CF template is bumped up to RHEL77GOLD
1. observe ansible logs showing AWS tests
1. change to "infra" user who does not have GOLD access  `export AWS_PROFILE=infra`
1. run script
1. observe CF template remains RHEL77
1. observe ansible logs checking access
1. change agnosticv `gpte` `OCP_CLIENTVM` dev.yaml: `clientvm_instance_image: RHEL77GOLD`
1. run script
1. observe CF template is degraded to RHEL77
1. observe ansible logs checking access

### DIFF
--- a/ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
+++ b/ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
@@ -1,5 +1,7 @@
-  - name: test 1
-    debug: msg="{{ instances }}"
+  - name: what instances am I dealing with?
+    debug:
+      verbosity: 2
+      var: instances
 
   - name: prep facts
     set_fact:
@@ -31,8 +33,8 @@
     when: 
       - item.value['image'] is defined
       - item.value['image'] is regex("^RHEL.*$")
-      - item.value['image'] is not regex("^.*GOLD$")      
-      - aws_ami_region_mapping[aws_region][item.value['image']|regex_replace("^(.*)$", "\\1GOLD")] is defined
+      - item.value['image'] is not regex("^.*GOLD$")
+      - item.value['image'] |regex_replace("^(.*)$", "\\1GOLD")] in aws_ami_region_mapping[aws_region]
     ec2_ami_facts:
       # profile: "{{ profile }}"
       region: "{{ aws_region }}"
@@ -44,7 +46,8 @@
 
   - name: debug ami_presence
     debug:
-      msg: "{{ ami_presence.results }}"
+      verbosity: 2
+      var: ami_presence.results
 
   - name: rename successful bumps from RHEL<VERSION> to RHEL<VERSION>GOLD in instances_DoD
     # items that returned images array not empty
@@ -113,4 +116,5 @@
 
   - name: debug instances
     debug:
-      msg: "{{ instances }}"
+      verbosity: 2
+      var: instances
--- a/ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
+++ b/ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
@@ -1,0 +1,116 @@
+  - name: test 1
+    debug: msg="{{ instances }}"
+
+  - name: prep facts
+    set_fact:
+      instances_DoD: "{{ [] }}"  # DoD
+  
+  - name: fail if I can't map an image to aws_ami_region_mapping
+    loop: "{{ instances }}"
+    when:
+    - item.image is defined
+    - aws_ami_region_mapping[aws_region][item.image] is not defined
+    fail:
+      msg: >- 
+        instance.image "{{ item.image }}" is not defined in aws_ami_region_mapping.
+
+  - name: create instances_DoD from instances LoD
+    loop: "{{ instances }}"
+    loop_control:
+      extended: yes
+    set_fact: 
+      instances_DoD: >-
+        {{ 
+          instances_DoD| combine(
+            { ansible_loop.index0: instances[ansible_loop.index0]  }
+          )
+        }}
+
+  - name: test if bump RHEL<VERSION> images to GOLD, if available in aws_ami_region_mapping
+    loop: "{{ instances_DoD|dict2items }}"
+    when: 
+      - item.value['image'] is defined
+      - item.value['image'] is regex("^RHEL.*$")
+      - item.value['image'] is not regex("^.*GOLD$")      
+      - aws_ami_region_mapping[aws_region][item.value['image']|regex_replace("^(.*)$", "\\1GOLD")] is defined
+    ec2_ami_facts:
+      # profile: "{{ profile }}"
+      region: "{{ aws_region }}"
+      image_ids: >-
+        {{ aws_ami_region_mapping[aws_region][
+            item.value['image']|regex_replace("^(.*)$", "\1GOLD")
+          ] }}
+    register: ami_presence
+
+  - name: debug ami_presence
+    debug:
+      msg: "{{ ami_presence.results }}"
+
+  - name: rename successful bumps from RHEL<VERSION> to RHEL<VERSION>GOLD in instances_DoD
+    # items that returned images array not empty
+    loop: "{{ ami_presence.results|json_query(' [? not_null(invocation) && not_null(images)].item.key' ) }}"
+    set_fact:
+      instances_DoD: >-
+        {{ instances_DoD| combine(
+          { item:
+            { "image": instances_DoD[item]['image']|regex_replace("^(.*)$", "\1GOLD")}
+            }, 
+          recursive=True
+        )}}
+
+  - name: test all images.  Some have <ANYTHYING>, RHEL<VERSION>, RHEL<VERSION>GOLD
+    ignore_errors: True
+    loop: "{{ instances_DoD|dict2items }}"
+    when: 
+      - item.value['image'] is defined
+      - aws_ami_region_mapping[aws_region][item.value['image']] is defined
+    ec2_ami_facts:
+      # profile: "{{ profile }}"
+      region: "{{ aws_region }}"
+      image_ids: "{{ aws_ami_region_mapping[aws_region][item.value['image']] }}"
+    register: ami_presence
+
+  - name: if there was a failure of a GOLD and there's no RHEL<VERSION>, bail out
+    loop: "{{ ami_presence.results|json_query(' [? not_null(invocation) && ! not_null(images)].item.key ') }}"
+    vars:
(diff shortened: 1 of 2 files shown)


ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
...
+        {{ instances_DoD| combine(
+          { item:
+            { "image": instances_DoD[item]['image']|replace("GOLD", "")}
+            }, 
+          recursive=True
+        )}}
+
+  - name: rebuild instances LoD
+    set_fact:
+      instances: "{{ instances_DoD.values() | list }}"
```
fridim: This won't work if `instances` is passed as extra-vars.
The value passed as extra-vars will always takes precedence over `set_fact` and thus stay unchanged.

I don't know if it's an issue, but something to keep in mind.

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
...
+    loop: "{{ instances }}"
+    loop_control:
+      extended: yes
+    set_fact: 
+      instances_DoD: >-
+        {{ 
+          instances_DoD| combine(
+            { ansible_loop.index0: instances[ansible_loop.index0]  }
+          )
+        }}
```
fridim: Isn't this equivalent to:

```yaml
- name: create instances_DoD from instances LoD
  set_fact:
    instances_DoD: "{{ instances }}"
```

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+  - name: test if bump RHEL<VERSION> images to GOLD, if available in aws_ami_region_mapping
+    loop: "{{ instances_DoD|dict2items }}"
```
fridim: And then you wouln'dt need `dict2items`

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
...
+    when: 
+      - item.value['image'] is defined
+      - item.value['image'] is regex("^RHEL.*$")
+      - item.value['image'] is not regex("^.*GOLD$")      
+      - aws_ami_region_mapping[aws_region][item.value['image']|regex_replace("^(.*)$", "\\1GOLD")] is defined
```
fridim: So `item.value['image']` would simply become `item.image`  is you're using a copy of `instances`.

Not 100% sure but this line with key evaluation could cause error in ansible.

A better way to test presence of a key is:


```yaml
item.image | regex_replace("^(.*)$", "\\1GOLD") in  aws_ami_region_mapping[aws_region] | default([])   
```

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
...
+    # items that returned images array not empty
+    loop: "{{ ami_presence.results|json_query(' [? not_null(invocation) && not_null(images)].item.key' ) }}"
+    set_fact:
+      instances_DoD: >-
+        {{ instances_DoD| combine(
+          { item:
+            { "image": instances_DoD[item]['image']|regex_replace("^(.*)$", "\1GOLD")}
+            }, 
+          recursive=True
+        )}}
```
fridim: So the idea is to  update a key in a dict that is part of a list. (`instances`).

Why don't we come up with a new dedicated variable instead ?

```dynamic_instances_image_promotion:
  instance_name:  RHEL77GOLD
```
And then in the template, instead of doing:

```yaml
      ImageId:
        Fn::FindInMap:
        - RegionMapping
        - Ref: AWS::Region
        - {{ instance.image | default(aws_default_image) }}
```

We do

```yaml
      ImageId:
        Fn::FindInMap:
        - RegionMapping
        - Ref: AWS::Region
        - {{ dynamic_instances_image_promotion[instance.name] | default(instance.image) | default(aws_default_image) }}
```

That would require to update all the templates.
But we will have to update all of them since in this code we're using aws_ami_region_mapping  mapping and it's not always included..

```
lolo:{fix-test-bool-when} ~/sync/dev/aad/ansible$ find -name 'ec2_cloud_template.*'| while read i; do if ! grep -q aws_ami_region_mapping $i; then echo $i;fi; done
./configs/ocp-ha-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/ansible-provisioner/files/cloud_providers/ec2_cloud_template.j2
./configs/three-tier-app/files/cloud_providers/ec2_cloud_template.j2
./configs/ansible-cicd-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/ocp-ha-disconnected-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/ocp-gpu-single-node/files/cloud_providers/ec2_cloud_template.j2
./configs/ansible-tower-implementation/files/cloud_providers/ec2_cloud_template.j2
./configs/rhte-oc-cluster-vms/files/cloud_providers/ec2_cloud_template.j2
./configs/ocp-implementation-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/ocp-storage-cns/files/cloud_providers/ec2_cloud_template.j2
./configs/ansible-windows/files/cloud_providers/ec2_cloud_template.j2
./configs/ans-tower-ops-demo/files/cloud_providers/ec2_cloud_template.j2
./configs/ans-network-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/amq-messaging-foundations/files/cloud_providers/ec2_cloud_template.j2
./configs/ocp-workshop-with-clientvms/files/cloud_providers/ec2_cloud_template.j2
./configs/ans-tower-lab/files/cloud_providers/ec2_cloud_template.j2
./configs/ansible-integrations/files/cloud_providers/ec2_cloud_template.j2
./configs/test-empty-config/files/cloud_providers/ec2_cloud_template.j2
```

I expect all those configs to fail with your code, since the mapping doesn't exist in the templates ?

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
```diff
@@ -0,0 +1,116 @@
+  - name: test 1
+    debug: msg="{{ instances }}"
+
+  - name: prep facts
+    set_fact:
+      instances_DoD: "{{ [] }}"  # DoD
```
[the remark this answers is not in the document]
newgoliath: Dict of Dicts

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: Can I access extra-vars to integrate them into this?

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: yup

newgoliath: Nope, it's going from a LIST of DICTS (LoD) to a DICT of DICTS (DoD)

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: I turn the index of the list into a keybut I want to keep order.  Otherwise I can't find the instances in the proper order.  If order even matters.

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: seems legit.  I think I didn't test that.

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: It does work, because all code has access to the `aws_ami_region_mapping` even if not used in the template.

If the template doesn't process instance.image (which many still don't, as above) then nothing is changed.

Template that do not use them, ignore them.

ansible/roles-infra/infra-ec2-template-generate/tasks/validate_instances_amis.yaml
newgoliath: I'm changing to that test format.  If I used `|default([])`, don't I always resolve True?

PR #1407, opened 2020-03-25
Studentvm

wkulhanek:
New studentvm config (for EC2 and OSP)
roles-student directory with roles to customize a studentvm


wkulhanek:
@fridim @stencell For your reading pleasure. :-) And approval.

ansible/configs/studentvm/default_vars_ec2.yml
```diff
@@ -0,0 +1,50 @@
...
+
+subdomain_base_short: "{{ guid }}"
+subdomain_base_suffix: ".example.opentlc.com"
+subdomain_base: "{{subdomain_base_short}}{{subdomain_base_suffix}}"
+
+## Environment Sizing
+
+studentvm_instance_type: "t3a.large"
+studentvm_instance_image: RHEL81GOLD
+studentvm_rootfs_size: 200
```
stencell: I don't actually know what the cost is here, but considering how many clientvm type workloads we configure, this can probably be defaulted to much smaller. If it needs to be large for some reason, it can be overidden in agnosticv

ansible/configs/studentvm/files/repos_template.j2
```diff
@@ -0,0 +1,25 @@
+# RHEL 8 Repos
```
stencell: If you intend on using Satellite with content views, this file is not even needed. It was only needed for bypassing that and accessing directly with certs. Now that we randomize names, there is really no good reason to do that...

ansible/configs/studentvm/sample_vars/studentvm_osp_bare_bones.yml
```diff
@@ -0,0 +1,32 @@
...
+# Settings for role studentvm-user
+# studentvm_user_vars:
+#   # Install a user id "{{ student_name }}"
+#   user_name: "{{ student_name }}"
+#   # If no student_password is provided a randomly generated password will be used
+#   student_password: ""
+
+# Deploy in a Ansible 2.9 Virtualenv
+agnosticv_meta:
+  virtualenv: openstack-ansible-2.9
```
stencell: Does this work? I didn't think we went back and update some of the module names that changed from 2.8 > 2.9. Like os_facts/info.

ansible/configs/studentvm/software.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Run Student VM Role
+      include_role:
+        name: "{{ studentvm_role }}"
+      loop: "{{ studentvm_roles }}"
+      loop_control:
+        loop_var: studentvm_role
+
+  - name: Install FTL
+    when: >-
+      ftl_install | d(False) | bool or
```
stencell: This is kind of in the middle of its evolution. You have `ftl_install` set to false (as well as default here), but you have `ftl_injector_tag` defined, so it will run. The idea was to drop the `ftl_install` completely and let the `ftl_injector` var decide if anything happens or not, but that got sidetracked and then there was more attention on fixing FTL wholistically.

So, a long way of saying, consider dropping one of these or cleaning up the default vars so it is not as confusing where you say `ftl_install=false`, but by defining the injector tag this evaluates true.

ansible/configs/studentvm/sample_vars/studentvm_osp_bare_bones.yml
wkulhanek: Yes it does. And facts gets mapped into info on 2.9. It'll show a warning.

ansible/configs/studentvm/default_vars_ec2.yml
wkulhanek: Yeah. t3a.medium should be fine.

ansible/configs/studentvm/files/repos_template.j2
wkulhanek: Let me try to delete and run one again.

ansible/configs/studentvm/software.yml
wkulhanek: OK. Removing install_ftl

ansible/roles-studentvm/studentvm-bash-prompt/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    var: studentvm_bash_prompt
+    verbosity: 2
+
+- name: Install bash-git-prompt
+  become: true
+  block:
+  - name: Download bash-git-prompt
+    git:
+      repo: https://github.com/magicmonty/bash-git-prompt.git
+      dest: "/home/{{ studentvm_bash_prompt.user_name }}/.bash-git-prompt"
```
stencell: This is a little confusing. The default that is set here is an email address which might error out if it is not a user on the system. Can this reference something like `{{ student_name }}`? Or better, in the defaults for the role.

ansible/configs/studentvm/default_vars_ec2.yml
stencell: i meant for disk sizes...not instance sizes. Although both, I guess.

ansible/roles-studentvm/studentvm-bash-prompt/tasks/main.yml
wkulhanek: Sample vars:
Set studentvm_bash_promt_vars.user_name: {{ student_name }}

stencell: ah, i missed that. this is what you get with so many files.

PR #1410, opened 2020-03-25
Fix ODH CR for CephObjectStore

LaVLaS:
##### SUMMARY
Fixes an error in role `ocp4-workload-rhte-analytics_data_ocp_workshop` when creating a CephObjectStore

##### ISSUE TYPE
Bugfix Pull Request

##### COMPONENT NAME
ocp4-workload-rhte-analytics_data_ocp_workshop


LaVLaS:
@thoraxe Here is the fix for the ODH role

PR #1435, opened 2020-03-27
Osp public hostname

ezr-ondrej:
##### SUMMARY
Set osp `public_dns_hostname` fact for all instances that have `public_dns: true`

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Cloud provider `osp`

### DIFF
--- a/ansible/configs/ansible-tower-implementation/files/cloud_providers/osp_cloud_template_master.j2
+++ b/ansible/configs/ansible-tower-implementation/files/cloud_providers/osp_cloud_template_master.j2
@@ -172,16 +172,18 @@ resources:
       user_data_format: RAW
       networks:
         - port: {get_resource: port_{{ iname }}}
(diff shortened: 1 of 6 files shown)


ezr-ondrej:
Should get review from @fridim!

ezr-ondrej:
After a discussion, it apprears it could be a good idea to have a `public_dns` for every host with a floating IP. That case, we could just drop the `public_dns` instance variable and be done with it.

NonyNo3lle:
@ezr-ondrej @fridim can we close this one? It seems old and no longer needed

ezr-ondrej:
If it wasn't addressed by someone else, it would be nice to have. I can finish it off if I get an opinion. But I'm ok if we close as well, it's not a show stopper

fridim:
closing, as discussed

PR #1457, opened 2020-03-30
Make RSA output multiline

makentenza:
##### SUMMARY
Make the output of the SSH key multiline and remove '/n' characters.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
sap-hana

### DIFF
--- a/ansible/configs/sap-hana/post_software.yml
+++ b/ansible/configs/sap-hana/post_software.yml
@@ -75,41 +75,51 @@
 
       when: tower_instance_count > 0
 
-- name: Print out information for the user to consume the lab
-  hosts: localhost
-  gather_facts: False
-  become: false
-  tasks:
-
-    - name: print out user.info
-      debug:
-        msg: "{{ item }}"
-      loop:
-        - "user.info: You can access your bastion via SSH:"
-        - "user.info: ssh cloud-user@bastion-{{ guid }}.{{ guid }}.{{ osp_cluster_dns_zone }}"
-        - "user.info: "
-        - "user.info: Make sure you use the following RSA key while connecting to the Bastion Host:"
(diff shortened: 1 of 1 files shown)


ansible/configs/sap-hana/post_software.yml
```diff
@@ -108,7 +108,7 @@
         - "user.info: ssh cloud-user@bastion-{{ guid }}.{{ guid }}.{{ osp_cluster_dns_zone }}"
         - "user.info: "
         - "user.info: Make sure you use the following RSA key while connecting to the Bastion Host:"
-        - "user.info: {{ lookup('file', '{{ output_dir }}/{{ guid }}_infra_ssh_key.pem') | trim }}"
+        - "user.info: {{ lookup('file', '{{ output_dir }}/{{ guid }}_infra_ssh_key.pem').split('\n') }}"
```
jkupferer: I don't believe this will be picked up by cloudforms as you might expect... Is this for Red Hat Summit? If so, then we should get you switched over to using `agnosticd_user_info`. I will reach out directly...

jkupferer: Here is an example of how this may look using the new `agnosticd_user_info` module: https://github.com/redhat-cop/agnosticd/blob/sap-hana-user-info/ansible/configs/sap-hana/post_software.yml#L85-L123

makentenza: @jkupferer no, this is not for Summit. Can I still use agnosticd_user_info or do you recommend any other approach?

jkupferer: I still recommend using `agnosticd_user_info`. It is compatible with the older `debug` module approach and eventually we are going to move everything over anyway. This will just put your config ahead of the curve :)

makentenza: Cool, will do that then

makentenza:
@jkupferer just added your suggestions

PR #1468, opened 2020-03-31
Changing key creation to make it simple

### LINKED ISSUE #1031: Simplify instructions for generating SSH keys for AWS
**Describe the bug**

The SSH key generation for AWS are over-complex and could greatly be simplified at https://github.com/redhat-cop/agnosticd/blob/development/docs/Preparing_your_workstation.adoc#aws-existing-resources

**To Reproduce**
Go to the link and read :-)

**Expected behavior**
using `ssh-keygen` instead of openssl would get rid of all the messing around with chmod.
And touching the config isn't necessary. `ssh-keygen -t rsa -b 4096 -a 100` worked for me, but I didn't want to propose a PR without being sure that I don't miss another point.

**Screenshots / logs**
n/a

**Verions (please complete the following information):**
n/a

**Additional context**
n/a

gmontalvoy:
##### SUMMARY
Change the prepare your workstation and agnosticd_mac documents to simplify the key creation process
Fixes #1031 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
agnosticd_mac.adoc
Preparing_your_workstation.adoc


gmontalvoy:
based on aws docs:
       --public-key-material (blob)
          The public key. For API calls, the text must be base64-encoded.  For
          command line tools, base64 encoding is performed for you.

ssh-keygen does not generated a base64 and relaying on aws to perform the encoding results on the following failure:
`➜  ~ aws ec2 import-key-pair --key-name rsaOsx --region=us-east-2 --output=text --public-key-material "`cat ~/.ssh/id_rsa.pub | grep -v PUBLIC`"

Invalid base64: "ssh-rsa aBLABLABLABLABLA+cBcj5xvC/dCcGbXLLAfehqqErNDeA2cZeRIusY91Blgi3EI3tUl3zk5stBv9EVBdHUcGqD+BI5ExIc2MQ+/BSy3e2h2qaD2AW1mo3En2p13K6hU4c+ykuHFCDwrNuHpHozjBacoujN3xl0H5VQQn9B7PxQkQ/ germanmontalvo@singularity"`

But it can be available to import a keypair changing the --public-key-material to fileb://~/.ssh/${KEYNAME}.pub

gmontalvoy:
There are a lot of references along several configs which is pointing to ${{ KEYNAME }}.pem so this PR will break the preparing workstation both linux and osx.

I will revisit this in a few days.

PR #1476, opened 2020-04-01
Ocp4 cluster update

wkulhanek:
Added support for ec2


ansible/configs/ocp4-cluster/default_vars.yml
```diff
@@ -54,19 +54,23 @@ install_ftl: false
 
 # TODO: Decide on whether to use sat or give access to repos directly with key
 # This will tell Agnosticd to use either:
-# sattelite, rhn, or file for repos
+# satellite, rhn, or file for repos
+# Only satellite is supported for RHEL 8
 repo_method: file
```
stencell, line 59: If we're setting RHEL 8 as the default for these providers, we should switch this to Satellite at the default.

ansible/configs/ocp4-cluster/default_vars.yml
```diff
@@ -91,6 +95,22 @@ common_packages:
 - podman
 - tree
 
+# Packages to install for RHEL 8
+common_packages_el8:
```
stencell, line 99: I think these are already defaults in `common` role now. Removing them from here would be one less thing to maintain. Same with the RHEL 7 ones.

wkulhanek: Right. Couldn't do that in my branch - I don't have your changes merged yet. Subsequent PR will remove those two common_packages...

ansible/configs/ocp4-cluster/default_vars.yml
wkulhanek: I set satellite for ec2 and osp. Don't know if we have RHEL 8 images for GCP and Azure...

ansible/configs/ocp4-cluster/pre_infra.yml
```diff
@@ -1,33 +1,49 @@
...
+    when: windows_password is not defined
+    block:
+    - name: Generate windows Administrator password if not already defined
+      command: openssl rand -base64 25
+      register: password_gen_r
+    - name: set_fact generated_windows_password (just generated)
+      set_fact:
+        generated_windows_password: "{{ password_gen_r.stdout }}"
+
+  - name: Set up AWS Student User
```
stencell, line 35: What exactly is this used for?

ansible/configs/ocp4-cluster/default_vars.yml
stencell: I see them set in sample vars. They should be in default vars for those two cloud providers since RHEL 8 will not work with file repo_method at this time.

ansible/roles-infra/infra-osp-template-generate/templates/nested_version/cloud_template_master.j2
```diff
@@ -25,7 +25,9 @@ resources:
       user: { get_resource: {{ guid }}-project_user }
       roles:
         - {project: {{ osp_project_name }}, role: _member_ }
+{% if osp_use_swift | d(true) | bool %}
```
stencell, line 28: I think we can consider getting rid of this file. @fridim left it here when we first transitioned to the single one last year, but I think the disconnected config is the only thing that ever used nested. And that is being converted right now.

wkulhanek: I just added that variable because I need that to not use SWIFT when I deploy 4.4 (on Green). But I'm happy to remove the nested config(s) when the time comes.

ansible/configs/ocp4-cluster/pre_infra.yml
wkulhanek: Azure I believe.

ansible/configs/ocp4-cluster/default_vars.yml
wkulhanek: OK I can restructure after.

ansible/roles/host-ocp4-provisioner/tasks/ec2_prereqs.yml
```diff
@@ -0,0 +1,63 @@
+- name: Get awscli bundle
```
stencell, line 1: Something we can consider installing via pip into venv and pin to a version? Because it never breaks when we use whatever the latest is...

wkulhanek: I was thinking the same.

ansible/roles/host-ocp4-provisioner/tasks/ec2_prereqs.yml
```diff
@@ -0,0 +1,63 @@
...
+    - /tmp/awscli-bundle
+    - /tmp/awscli-bundle.zip
+
+- name: Create .aws directory
+  become: no
+  file:
+    path: ~/.aws
+    state: directory
+
+- name: Add aws credentials for sandboxes
```
stencell, line 32: I may have missed it, but do we get rid of these after the install? There is really no reason to leave credentials on the bastion after the install is done since the cloud-credentials operator has used those and created what it needs inside of OpenShift.

Same for the following task, especially because it is a bigger account.

wkulhanek: We don't. If we do remove then uninstalling openshift will be a problem. But a problem we could potentially solve.

ansible/configs/ocp4-cluster/pre_infra.yml
stencell: `Set up aWS Student User` is used for Azure? That doesn't make any sense.

wkulhanek: Yeah. Actually it happens on EC2 as well if there are Windows VMs... May not be relevant for this config.

ansible/configs/ocp4-cluster/post_infra.yml
```diff
@@ -46,3 +46,59 @@
...
+  - name: AWS configuration for Bastions
+    when: cloud_provider == 'ec2'
+    block:
+    - name: Create .aws for awscli
+      file:
+        path: ~/.aws
+        mode: 0700
+        state: directory
+
+    - name: Write down Route53User credentials
```
stencell, line 93: Do we need this in addition to the credentials that are written out later on? Do these update a different r53 zone that the sandbox or gpte creds?

stencell: Also, just making a note like the others, this should also be removed afterwards.

wkulhanek: Those are for Lets Encrypt. At least they were...

ansible/configs/ocp4-cluster/pre_infra.yml
stencell: Windows VMs? I don't see a conditional for that?

ansible/configs/ocp4-cluster/default_vars.yml
wkulhanek: done

PR #1482, opened 2020-04-01
Updated Student VM Samples and default vars. Removed common packages …

wkulhanek:
…from ocp4-cluster (now default on common role)


wkulhanek:
@stencell This should be easier. :-) And I removed the common packages from ocp4-cluster as well.

stencell:
simple changes, looks good

PR #1526, opened 2020-04-08
Extend tower roles functionality - allow virtualenvs to specify their python executable/version

tonykay:
##### SUMMARY
The tower-custom-venv role automatically used the system python (python 2 on RHEL 7.x). This allows an optional `virtualenv_python` to be specified.

In addition, 2 tower roles were extended to remove the hardcoded tower admin user and allow an optional alternative, but defaulting to admin.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
role/tower-custom-venv


ansible/roles/tower-user-create/tasks/main.yml
```diff
@@ -8,18 +8,20 @@
...
-     email: "{{ item.email | default('rhpds-admins@redhat.com') }}"
-     first_name: "{{ item.firstname | default(item.user) }}"
-     last_name: "{{ item.lastname | default(item.user) }}"
-     superuser: "{{ item.superuser | default('no')}}"
+     username: "{{ user.user }}"
+     password: "{{ user.password | default('change_me') }}"
+     email: "{{ user.email | default('rhpds-admins@redhat.com') }}"
+     first_name: "{{ user.firstname | default(user.user) }}"
+     last_name: "{{ user.lastname | default(user.user) }}"
+     superuser: "{{ user.superuser | default('no')}}"
```
fridim: isn't `user` even more generic than `item` ? what happens if `user` is passed as extra-var ?

PR #1527, opened 2020-04-08
Fix operator versions; add retries to avoid slow operators failing

jamesfalkner:
##### SUMMARY
This fix removes the `startingCSV` values so that the operator subscriptions always get the version appropriate for the OCP release it's being installed on, rather than starting with a specific version and then auto-upgrading (which takes more time and could fail the deployment).

It also adds `retries` to the verification step in case an operator takes a bit longer than expected to asynchronously provision.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`ocp4-workload-quarkus-workshop`

### DIFF
--- a/ansible/roles/ocp4-workload-quarkus-workshop/files/amqstreams_subscription.yaml
+++ b/ansible/roles/ocp4-workload-quarkus-workshop/files/amqstreams_subscription.yaml
@@ -10,4 +10,3 @@ spec:
   name: amq-streams
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: amqstreams.v1.3.0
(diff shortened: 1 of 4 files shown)


stencell:
By removing your `startingCSV`, don't you also run the risk of getting the latest and greatest that may or may not have been tested with your workshop content?

jamesfalkner:
Yes, that is a risk - but there is no other alternative until OCP 4.4. If we _do_ specify a `startingCSV`, we run the risk that the version we specify could be removed from the operator catalog _at any time_ (and this happened a couple of months ago, which is how we got into this chasing-the-operator game in the first place). According to the operators team @ red hat, there's _no solid way_ to 'pin the version' to a specific operator version. So we are taking this risk now, since OCP 4.4 is nearing release.

PR #1533, opened 2020-04-09
Made rhel8lab sampe_vars_osp generic and deployable

tonykay:
##### SUMMARY
sample_vars_osp contained hardcoded and non-generic values breaking deploy if used

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
rhel8lab


tonykay:
> You could remove the ddns as well. Both osp_cluster_dns_server and dns should come from secret

Yes, will do in next

PR #1564, opened 2020-04-15
Fix cronjob PATH + post-hook

fridim:
- fix the path, use absolute
- do not redirect to /dev/null, we should know when something goes wrong
- add post-hook according to documentation

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
hosts-lets-encrypt-certs-certbot role

### DIFF
--- a/ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
+++ b/ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
@@ -198,7 +198,7 @@
       special_time: daily
       job: >-
         {{ _certbot_virtualenv }}/bin/certbot renew
-        {{ _certbot_additional_args|d(_certbot_args)|d('') }}
+        {{ _certbot_additional_args | default(_certbot_args) | default('') }}
         --config-dir={{ _certbot_dir }}/config
         --work-dir={{ _certbot_dir }}/work
         --logs-dir={{ _certbot_dir }}/logs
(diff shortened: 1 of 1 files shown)


wkulhanek:
@fridim Do we need a shell script to run the renew? Like we are doing with original request:
1. Activate the VirtualEnv
2. Call certbot

ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
```diff
@@ -196,4 +196,11 @@
     cron:
       name: "{{ _certbot_cron_job_name }}"
       special_time: daily
-      job: "certbot renew {{ _certbot_additional_args|d(_certbot_args)|d('') }} --config-dir={{ _certbot_dir }}/config --work-dir={{ _certbot_dir }}/work --logs-dir={{ _certbot_dir }}/logs --quiet > /dev/null"
+      job: >-
+        {{ _certbot_virtualenv }}/bin/certbot renew
+        {{ _certbot_additional_args|d(_certbot_args)|d('') }}
```
jkupferer, line 201: Minor comment, preferred style for this line would be:
```
        {{ _certbot_additional_args | default(_certbot_args) | default('') }}
```

fridim:
> @fridim Do we need a shell script to run the renew? Like we are doing with original request:
> 
> 1. Activate the VirtualEnv
> 2. Call certbot

@wkulhanek i tested and it seems to work, so the certbot executable is loading the right libraries from the virtualenv somehow

PR #1568, opened 2020-04-15
Satellite propagate all needed user_info

ezr-ondrej:
##### SUMMARY
The #1567 used incorrect group again, we should run this on localhost.
We have all the information locally, no need to involve remote hosts.

@jkupferer could you take a look?

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Config `satellite-vm`

### DIFF
--- a/ansible/configs/satellite-vm/post_software.yml
+++ b/ansible/configs/satellite-vm/post_software.yml
@@ -26,6 +26,17 @@
   roles:
     - role: satellite-verification
 
+- name: Set user info and data
+  hosts: localhost
+  tasks:
+    - name: Set satellite host name for the user
+      agnosticd_user_info:
+        data:
+          satellite_hostname: "{{ hostvars[groups['satellites'][0]]['publicname'] }}"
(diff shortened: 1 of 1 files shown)


ansible/configs/satellite-vm/post_software.yml
```diff
@@ -26,6 +26,17 @@
...
+- name: Set user info and data
+  hosts: localhost
+  tasks:
+    - name: Set satellite host name for the user
+      agnosticd_user_info:
+        data:
+          satellite_hostname: "{{ hostvars[groups['satellites'][0]]['publicname'] }}"
+          dns_zone: "{{ osp_cluster_dns_zone }}"
+          satellite_username: summit-admin
+          satellite_password: summit2020
```
ezr-ondrej, line 38: This is intentionally static, no point to change AgnosticV at this point and it would be neccesary not to be static, we just want to prepare it for the following usage of the config, so there would not need to be changes to `bookbag`.

PR #1571, opened 2020-04-15
change swiftoperator role default to false

stencell:
##### SUMMARY
In the cloud provider template, the default was to enable the use of the `swiftoperator` role. Since the Novello clusters will not run Swift, this should be defaulted to false. Some of the new clusters don't even have this role defined and deployments fail. If you need to use it, you can still set the `osp_use_swift` var in a config.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
infra-osp-template-generate


stencell:
> As long as the configs (for Blue...) are updated with the right variable...

The only thing that should use Blue is the ocp4-disconnected config...and we haven't used Swift there for a few months now. Since it was UPI, there was no need for ignition. And there are steps in both the solver and lab instructions to patch the image registry to use NFS.

I can't think of another config that is (should) be using this, can you?

wkulhanek:
Yeah. I think we're fine.

PR #1597, opened 2020-04-19
enable sudo and set default packages

stencell:
##### SUMMARY
This PR changes the default for the student VM user workload to enable sudo. You can still override this and set to false if you need, but that should not be as common.

It also adds the buildah and skopeo packages to the common list for EL8.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
common
studentvm-user

### DIFF
--- a/ansible/roles-studentvm/studentvm-user/defaults/main.yml
+++ b/ansible/roles-studentvm/studentvm-user/defaults/main.yml
@@ -5,7 +5,7 @@ studentvm_user_defaults:
   user_name: lab-user
 
   # Enable sudo for the student user
-  enable_sudo: false
+  enable_sudo: true
 
   # Use a password for the student account
   # When false use / generate ssh key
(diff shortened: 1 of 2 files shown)


wkulhanek:
I am against that. If a config needs it it should set the flag. We want to be a bit more restrictive on how many environments actually need root access.

stencell:
> I am against that. If a config needs it it should set the flag. We want to be a bit more restrictive on how many environments actually need root access.

I can go ultimately go either way, but I tend to disagree. Removing sudo access just introduces friction - either on the part of the user or the developer that is creating/maintaining the config. Is there a mandate or reason behind wanting to be more restrictive? These are training environments, not production environments.

Having sudo enabled doesn't mean it has to be used. But if a skilled user runs into a problem, having sudo access allows them to potentially identify or fix it. If they don't have that, they have to do something like open a ticket and they're probably just as likely to stop doing the training at that point.

The onus for disabling sudo should be on the config developer if they are certain it is not needed or beneficial.

PR #1598, opened 2020-04-20
Update cluster name for OCP4 installs

stencell:
##### SUMMARY
When we get an `e` in our GUID, it is sometimes interpreted as an exponent instead of a string. This will change the names of clusters from `$GUID` to `cluster-$GUID`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-disconnected-osp-lab
ocp-infra-osp-fip
ocp4-cluster


stencell:
@wkulhanek I changed the default `install-config.yaml` template to for the ocp4 provisioner role. Since ec2, azure, & gcp are full IPI including DNS I don't think other changes are required. The DNS changes for the osp provider should be in here and I've validated them, but have a look.

PR #1607, opened 2020-04-20
Move KNative/Pipeline at the front of the process

mbogoevici:
Add Tekton resource validation to the Infra role.

### DIFF
--- a/ansible/roles/ocp4-workload-ml-workflows-infra-summit2020/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-ml-workflows-infra-summit2020/tasks/workload.yml
@@ -5,11 +5,6 @@
   become: false
   when: install_storage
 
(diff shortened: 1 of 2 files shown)


ansible/roles/ocp4-workload-ml-workflows-infra-summit2020/tasks/workload.yml
```diff
@@ -20,6 +15,11 @@
   become: false
   when: ACTION == "create" or ACTION == "provision"
 
+- include_tasks: per_user_workload.yml
+  loop: "{{ range(user_count_start | int, (user_count_start | int) + (user_count | int)) | list }}"
+  loop_control:
+    loop_var: user_num
```
jkupferer, line 21: I think we want `user_count` here?

PR #1621, opened 2020-04-21
Fix loop var for user project creation

jkupferer:
##### SUMMARY

The last pull request introduced a bug in the user project creation. This pull request fixes that issue while also introducing safe task level variable naming using `t_` prefix to indicate var is task level.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp4-workload-quarkus-workshop-user-create-project workload role


jkupferer:
Pinging @sshaaf

PR #1635, opened 2020-04-23
Adding new OCP4 role for RHMI 2.x (Integreatly)

pmccarthy:
##### SUMMARY
Adding new base template OCP4 role for managing Red Hat Managed Integration (Integreatly) installations. This role does not do anything yet but will act as the foundation for future development which will be progressed over the coming days/weeks

Equivalent OCP3 role can be found [here](https://github.com/redhat-cop/agnosticd/tree/development/ansible/roles/ocp-workload-integreatly)

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
Red Hat Managed Integration (Integreatly)

### DIFF
--- a/ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
+++ b/ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
@@ -1,0 +1,29 @@
+---
+# Implement your Post Workload deployment tasks here
+# --------------------------------------------------
+
+
+
+
+# Leave these as the last tasks in the playbook
+# ---------------------------------------------
+
+# For deployment onto a dedicated cluster (as part of the
+# cluster deployment) set workload_shared_deployment to False
+# This is the default so it does not have to be set explicitely
+- name: post_workload tasks complete
+  debug:
+    msg: "Post-Workload tasks completed successfully."
+  when:
+  - not silent|bool
+  - not workload_shared_deployment|d(False)
+
+# For RHPDS deployment (onto a shared cluster) set
+# workload_shared_deployment to True
+# (in the deploy script or AgnosticV configuration)
+- name: post_workload tasks complete
+  debug:
+    msg: "Post-Software checks completed successfully"
+  when:
+  - not silent|bool
+  - workload_shared_deployment|d(False)
--- a/ansible/roles/ocp4-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-integreatly/defaults/main.yml
@@ -1,0 +1,27 @@
+---
+become_override: False
+ocp_username: opentlc-mgr
+silent: False
+
+# If your workload needs customization options provide variables
+# in a dictionary. The dictionary should be named "<role name>_defaults".
+# So because this example workload is called "ocp-workload-example" the dictionary
+# should be named "ocp_workload_example_defaults". Note that the dictionary can not
+# include "-" characters, so replace all "-" with "_" instead.
+#
+# You can override the defaults as parameters to the Ansible run that runs
+# your workload. The overrides should be in a dictionary called
+# "<role name>_vars".
+#
+# If there are any secrets that need to be passed (data that should not be in a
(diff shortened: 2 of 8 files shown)


pmccarthy:
@NonyNo3lle @fridim Hey guys, perhaps you could have a quick look at this PR please? Or if not, direct me to someone who might be able to take a look? It's just a basic role that's not tied to any catalog item right now so should be safe to merge back. Thanks!

wkulhanek:
@pmccarthy I would suggest checking out the example workload that we recently updated: ocp-workload-example and starting by making a copy of that one.

We introduced a much cleaner way of passing variables into a workload using Ansible dictionaries. Now since I don't really know Integreatly I don't know if this is necessary - or if Integreatly always deploys the same. But it would make things a bit more encapsulated and less likely to collide variables with other parts of AgnosticD.

We also started using the meta/main.yml to put information about the roles including owners. So usually we don't have a n OWNERS file. Not that it will hurt.

Feel free to ping me anytime if you want more explanation than what's in the readme etc. of that role.

pmccarthy:
@wkulhanek thanks for the update. I'll go ahead and make the appropriate changes based on the example role you've referenced.

ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
```diff
@@ -0,0 +1,29 @@
...
+
+
+
+# Leave these as the last tasks in the playbook
+# ---------------------------------------------
+
+# For deployment onto a dedicated cluster (as part of the
+# cluster deployment) set workload_shared_deployment to False
+# This is the default so it does not have to be set explicitely
+- name: pre_workload tasks complete
```
austincunningham, line 14: Shouldn't this be
```suggestion
- name: post_workload tasks complete
```

ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
```diff
@@ -0,0 +1,29 @@
...
+  debug:
+    msg: "Post-Workload tasks completed successfully."
+  when:
+  - not silent|bool
+  - not workload_shared_deployment|d(False)
+
+# For RHPDS deployment (onto a shared cluster) set
+# workload_shared_deployment to True
+# (in the deploy script or AgnosticV configuration)
+- name: pre_workload tasks complete
```
austincunningham, line 24: Same here ^^

pmccarthy:
@wkulhanek I've just made updates to reflect the latest role configurations if you wouldn't mind taking a look please?

Also, couple of questions:
1) I wasn't sure what information is mandatory in the `meta/main.yml` file. I left out the `license` section as I'm not sure it's applicable?
2) In the README, is the agnosticv deployment section required? I'm not sure this is something that will be used in this role?

Thanks

ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
pmccarthy: @austincunningham - good spot, thanks

ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
wkulhanek: Oh. Yeah. I'll fix that in the example. The important part (the message) is correct. It's just the name of the task. Thanks for spotting. Duh.

wkulhanek:
@pmccarthy Thanks. I'll take a look. As for AgnosticV: I'm 99.9% sure sure it will be required. All new configs use AgnosticV to set the variables rather than our legacy shell scripts.

Once again it will depend if the workload can be customized or if it always uses the same settings.

Some of mine set defaults (e.g. for cli tool versions) that then can be overwritten using AgnosticV. That way we don't have to go and change AgnosticD and add new tags. We just update the config.

The meta/main is just whatever you put there. :-) It's just a standard place (and ansible-galaxy creates a template for it).

PR #1644, opened 2020-04-24
OCP4 Integreatly workload - evals user generation logic

ciaranRoche:
##### SUMMARY

This change generates 50 eval users to be used by integreatly workshop

JIRA: https://issues.redhat.com/browse/INTLY-7046

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Red Hat Managed Integration (Integreatly)

##### ADDITIONAL INFORMATION

```paste below
Wednesday 29 April 2020  12:02:49 +0100 (0:00:00.023)       0:01:17.094 ******* 
=============================================================================== 
ocp4-workload-integreatly-minio : Create objects for Minio operator ------------------------------------------------------------------------------------------------------------------------------------------------------------------ 17.31s
ocp4-workload-integreatly : Generate dedicated admin password ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 11.23s
ocp4-workload-authentication : Retrieve API server configuration (for API endpoint) --------------------------------------------------------------------------------------------------------------------------------------------------- 5.33s
ocp4-workload-authentication : Print User Information for each User ------------------------------------------------------------------------------------------------------------------------------------------------------------------- 5.26s
ocp4-workload-authentication : Generate htpasswd file --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 3.86s
ocp4-workload-integreatly : Delete current htpasswd secret ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.38s
ocp4-workload-authentication : Ensure htpasswd Secret is absent ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.38s
ocp4-workload-authentication : Set up Cluster Admin User ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 2.35s
ocp4-workload-authentication : Remove kubeadmin user secret --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.29s
ocp4-workload-integreatly : Add dedicated admin user to dedicated admin group --------------------------------------------------------------------------------------------------------------------------------------------------------- 2.28s
ocp4-workload-authentication : Update OAuth Configuration ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.25s
ocp4-workload-integreatly : Get current htpasswd secret ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.25s
ocp4-workload-integreatly : Create dedicated admin user ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.16s
Check if desired virtualenv is available on the host ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.03s
ocp4-workload-integreatly : Create new htpasswd secret -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 2.02s
ocp4-workload-authentication : Create htpasswd secret from htpasswd file -------------------------------------------------------------------------------------------------------------------------------------------------------------- 1.99s
ocp4-workload-integreatly : Backup htpasswd ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 1.90s
ocp4-workload-authentication : Generate htpasswd hash for admin user ------------------------------------------------------------------------------------------------------------------------------------------------------------------ 1.90s
ocp4-workload-authentication : Generate htpasswd hash for user passwords -------------------------------------------------------------------------------------------------------------------------------------------------------------- 1.81s
ocp4-workload-integreatly : Check for dedicated admins -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 1.78s


```

### DIFF
--- a/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
+++ b/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
@@ -3,3 +3,11 @@
 ocp_workloads:
   - ocp4-workload-integreatly-minio
   - ocp4-workload-integreatly
+  - ocp4-workload-authentication
+
+ocp4_workload_authentication_vars:
+  idm_type: htpasswd
+  htpasswd_user_base: evals
+  htpasswd_user_count: 50
+
+  admin_user: dedicated-admin
--- a/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
+++ b/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
@@ -2,12 +2,14 @@
 # override ocp workloads to include prerequisite roles
 ocp_workloads:
   - ocp4-workload-integreatly-minio
-  - ocp4-workload-integreatly
   - ocp4-workload-authentication
+  - ocp4-workload-integreatly
 
 ocp4_workload_authentication_vars:
   idm_type: htpasswd
+  
   htpasswd_user_base: evals
   htpasswd_user_count: 50
 
-  admin_user: dedicated-admin
+  # This user is not available in RHMI production/poc clusters. Only RHMI operations teams have cluster-admin access
+  admin_user: admin
--- a/ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
+++ b/ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
@@ -23,6 +23,52 @@
   with_sequence: count="{{ ocp4_workload_integreatly_defaults.user_count }}"
   when: ocp4_workload_integreatly_defaults.action == "install"
 
+- name: Get current htpasswd secret
+  shell: oc get secret htpasswd-secret -n openshift-config -o json | jq -j '.data.htpasswd' | base64 --decode > /tmp/htpasswd
+
+- name: Backup htpasswd
+  shell: cp /tmp/htpasswd /tmp/htpasswd.backup
+
+- name: Generate dedicated admin password
+  shell: openssl rand -base64 12
+  register: _gen_password
+
+- name: Set dedicated admin password
+  set_fact:
+    _dedicated_admin_password: "{{ _gen_password.stdout }}"
+
+- name: Create dedicated admin user
+  shell: htpasswd -B -b /tmp/htpasswd {{ ocp4_workload_integreatly.dedicated_admin_username }} {{ _dedicated_admin_password }}
+
+- name: Ensure htpasswd Secret is absent
+  k8s:
+    state: absent
+    api_version: v1
+    kind: Secret
+    name: htpasswd-secret
+    namespace: openshift-config
+
+- name: Create new htpasswd secret
+  shell: oc create secret generic htpasswd-secret --from-file=htpasswd=/tmp/htpasswd -n openshift-config
+
+- name: Check for dedicated admins
+  shell: oc get groups dedicated-admins
+  register: group_output
+  ignore_errors: True
+
+- name: Create dedicated admin group
+  shell: oc adm groups new dedicated-admins {{ ocp4_workload_integreatly.dedicated_admin_username }}
+  when: group_output.rc == 1
+
+- name: Add dedicated admin user to dedicated admin group
+  shell: oc adm groups add-users dedicated-admins {{ ocp4_workload_integreatly.dedicated_admin_username }}
+  when: group_output.rc != 1
+
+- name: Print Overview
+  debug:
+    msg: "{{ item }}"
+  with_items:
+    - "user.info: Dedicated Admin User: {{ ocp4_workload_integreatly.dedicated_admin_username }} / {{ _dedicated_admin_password }} (This is the most privileged user account available to consultants and customers in RHMI production/poc environments.)"
 
 
 # Leave these as the last tasks in the playbook
--- a/ansible/roles/ocp4-workload-integreatly/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-integreatly/defaults/main.yml
@@ -22,6 +22,7 @@ silent: False
 #
 # The logic in the workload should only use the dictionary "<role name>".
 ocp4_workload_integreatly_defaults:
+    dedicated_admin_username: "dedicated-admin"
     variable_1: "value 1"
     variable_2: "value 2"
     variable_secret: ""


ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
```diff
@@ -33,6 +33,72 @@
   - "Variable 2:      {{ ocp4_workload_integreatly.variable_2 }}"
   - "Variable Secret: {{ ocp4_workload_integreatly.variable_secret }}"
 
+- name: Get cluster token
```
pmccarthy, line 36: @ciaranRoche Wondering should this logic be in the post_workload task instead? 

In my head at a very high level the process would be as follows:
**pre_workload**: _Install and configure Minio_
**workload**: _Install integreatly-operator and an instance of RHMI. Wait for install to complete_
**post_workload**: _Setup users, create user specific resources for fuse/3scale etc._

Happy to hear your thoughts on the above though ^^ 

Pinging @aidenkeating @austincunningham @dimitraz for visibility

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
```diff
@@ -33,6 +33,72 @@
...
+  shell: oc create secret generic htpasswd-secret --from-file=htpasswd=/tmp/htpasswd -n openshift-config
+
+- name: Backup oauth/cluster from openshift-config namespace to tmp/cluster-oauth.backup.json
+  shell: oc get oauth/cluster -n openshift-config -o json > /tmp/cluster-oauth.backup.json
+
+- name: Patch cluster OAuth with htpasswd
+  shell: oc patch oauth/cluster --type json -n openshift-config -p '[{"op":"add","path":"/spec/identityProviders/-","value":{"mappingMethod":"claim","name":"htpasswd_provider","type":"HTPasswd","htpasswd":{"fileData":{"name":"htpass-secret"}}}}]'
+  when: result.stat.exists == false
+
+- name: Give cluster OAuth 30 seconds to adapt to changes
```
pmccarthy, line 82: Is there any resource we could poll instead? I'm always cautious of sleep statements

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
```diff
@@ -33,6 +33,72 @@
...
+  shell: sleep 30
+
+- name: Intialise evals user accounts
```
pmccarthy, line 85: Is this required to login to the cluster? Seems a bit strange that we would need to initialise beforehand

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
wkulhanek: One thing I'd recommend is to use `k8s`everywhere possible instead of oc shell commands.

Technically you could poll the cluster operator and make sure it's done rolling out.

Do you need to *do* anything as any of these users? If not it should be safe to just continue without waiting.

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
```diff
@@ -33,6 +33,72 @@
...
+  shell: oc login -u {{ evals_user_name_format|format(item|int) }} -p {{ evals_password }}
```
wkulhanek, line 86: Why is this bit necessary? Also there is a `k8s_auth` module that does that.

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
```diff
@@ -33,6 +33,72 @@
...
+  with_sequence: count={{ user_count }}
+  register: evals_login_result
+  retries: 20
+  until: evals_login_result.rc == 0
+  delay: 5
+
+- name: Restore kube:admin login session
```
wkulhanek, line 93: Do you mean *system:admin* ? Again not quite sure why this is necessary.

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
ciaranRoche: Yeah I was not 100% sure where this was to go, I initially was leaning towards `post_workload`. Happy to move.

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
ciaranRoche: I will remove the intialise eval users, we can then remove this wait as it will not be needed.

ansible/roles/ocp4-workload-integreatly/tasks/workload.yml
ciaranRoche: Only necessary due to how I initialized the evals user, will be removing

ciaranRoche:
@pmccarthy @wkulhanek applied your feed back. If you get a chance to take another look. Thank you.

wkulhanek:
@ciaranRoche I think I have a more fundamental question.... why do you need to create / update the users yourselves?

We have the capability in AgnosticD to create htpasswd authentication (see role `ocp4-workload-authentication`). In the config that deploys the OCP 4 Cluster you'd just set the right variables and then use that workload to set things up.

Something like:
```
ocp4_workload_authentication_vars:
  idm_type: htpasswd
  htpasswd_user_base: evals
  htpasswd_user_count: 50

  admin_user: admin
```

and optionally passwords for users/admin - otherwise they're generated.

All that would go into the AgnosticV config for the Catalog Item.

ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
```diff
@@ -2,4 +2,12 @@
 # override ocp workloads to include prerequisite roles
```
dimitraz, line 2: Does this directory have to be named `sample_vars`? Maybe just `vars` would be better?

aidenkeating: @dimitraz As far as I'm aware, `vars/` is a "standard" or "expected" directory name in a role for Ansible. I'd be tempted to keep it this way unless there's a good reason not to, to make it clear this is just a local dev example

ciaranRoche:
@aidenkeating @pmccarthy @wkulhanek Updated the PR to use `ocp4-workload-authentication` to create both `evals` users and an `admin` user.

As we require a `dedicated-admin` user and group, I have manually added this logic to the post workload on our intly workload.

ansible/roles/ocp4-workload-integreatly/tasks/post_workload.yml
```diff
@@ -2,7 +2,47 @@
...
+  register: _gen_password
+
+- name: Set dedicated admin password
+  set_fact:
+    _dedicated_admin_password: "{{ _gen_password.stdout }}"
+
+- name: Create dedicated admin user
+  shell: htpasswd -B -b /tmp/htpasswd {{ ocp4_workload_integreatly.dedicated_admin_username }} {{ _dedicated_admin_password }}
+
+- name: Delete current htpasswd secret
```
pmccarthy, line 22: @ciaranRoche Rather than deleting/creating the secret, would it be possible to run `oc replace` or something similar? Reason I ask is that in the RHMI 1.x ansible installer we've run delete/create commands and they can cause problems. For example, failed create commands after removing the resource can leave the cluster in a bad state.

pmccarthy:
@ciaranRoche Left one comment but overall this looks good. Would be interested in getting @wkulhanek feedback though, if he's happy then we can try get this merged.

ciaranRoche:
@pmccarthy I updated to follow the same pattern as the `ocp4-workload-authentication`

ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
dimitraz: Ok that makes sense. 
How are these workloads run sequentially then in an actual Integreatly PDS workshop? Is there some tower job that we have to create? If so where does that have to be created?
Sorry for all the questions, just not too familiar with this workflow

wkulhanek: When we set up a config we provide an ordered list of workloads to be applied after the OpenShift installation has finished. So in your example it would call the authentication workload first, followed by minio, followed by whatever else you need in whatever order.

That is a setting in the configuration for RHPDS.

wkulhanek:
@ciaranRoche @pmccarthy @aidenkeating Can you guys please take a look at the merge conflict. I am not confident I'll pick the right one...

aidenkeating:
@wkulhanek Thank you for taking a look. Rebased

PR #1645, opened 2020-04-24
Migration Workloads : align migration workloads with latest z-stream

pranavgaikwad:
##### SUMMARY
- removes `startingCSV` from OCP 4 workload to accept latest z-stream

- updates SHAs on OCP 3 side to align with latest z-stream 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp-workload-migration
ocp4-workload-migration

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/ocp4-workload-migration/templates/mig-operator-subscription-downstream.yml.j2
+++ b/ansible/roles/ocp4-workload-migration/templates/mig-operator-subscription-downstream.yml.j2
@@ -9,5 +9,3 @@ spec:
   name: cam-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: cam-operator.{{ mig_operator_release }}
-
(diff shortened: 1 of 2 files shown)


ansible/roles/ocp4-workload-migration/templates/mig-operator-subscription-downstream.yml.j2
```diff
@@ -9,5 +9,3 @@ spec:
   name: cam-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: cam-operator.{{ mig_operator_release }}
```
wkulhanek, line 12: Will that introduce the possibility of future breaks? Whenever a new operator version is released?

jwmatthews: We have that situation now.

We were told to use "startingCSV" to avoid this and pin to a specific operator, but that isn't working.

Asking @shawn-hurley (Operator Framework group lead) for help on figuring out what else is needed to pin this down for future, perhaps we are missing something in our rel-eng workflows for the operator.

In meantime, the workload is broken until this is removed.

jwmatthews:
@wkulhanek @newgoliath 

FYI, we learned that we need to change some of the metadata inside of our operator to allow pinning to a specific version in the OLM Subscription. 

We are intending to make the change for the next release of CAM (1.2.0 ~end of May).
Tracked in:  https://issues.redhat.com/browse/MIG-210

We'll share specifics of what is required in that JIRA once we've verified the changes work.

After that version is out we will support pinning and will update the migration workloads in agnosticd to be pinned to a specific version.

cc @jmontleon @pranavgaikwad

PR #1647, opened 2020-04-24
[WIP] Use package to install satellite ca certs

marcosmamorim:
##### SUMMARY
Replace download Satellite certificate by package module to install satellite CA certificates

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
set-repositories

### DIFF
--- a/ansible/roles/set-repositories/tasks/satellite-repos.yml
+++ b/ansible/roles/set-repositories/tasks/satellite-repos.yml
@@ -22,7 +22,7 @@
 - name: Install cert
   tags: packer
   package:
-    name: "http://{{ set_repositories_satellite_url | default(satellite_url) }}/pub/katello-ca-consumer-latest.noarch.rpm"
+    name: "https://{{ set_repositories_satellite_url | default(satellite_url) }}/pub/katello-ca-consumer-latest.noarch.rpm"
     state: present
   register: r_install_cert
   until: not r_install_cert.failed
--- a/ansible/roles/set-repositories/tasks/satellite-repos.yml
+++ b/ansible/roles/set-repositories/tasks/satellite-repos.yml
@@ -19,26 +19,15 @@
     name: katello-ca-consumer-*.noarch
     state: absent
 
-- name: Download Cert from Satellite
-  get_url:
-    url: "https://{{ set_repositories_satellite_url | default(satellite_url) }}/pub/katello-ca-consumer-latest.noarch.rpm"
-    dest: /root/katello-ca-consumer-latest.noarch.rpm
-    mode: 0664
-    validate_certs: no
-  register: r_get_cert
-  until: not r_get_cert.failed
+- name: Install cert
+  tags: packer
+  package:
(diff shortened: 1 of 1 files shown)


wkulhanek:
As long as we have https with proper certs on that server...

ansible/roles/set-repositories/tasks/satellite-repos.yml
```diff
@@ -19,27 +19,16 @@
...
-    validate_certs: no
-  register: r_get_cert
-  until: not r_get_cert.failed
+- name: Install cert
+  tags: packer
+  package:
+    name: "https://{{ set_repositories_satellite_url | default(satellite_url) }}/pub/katello-ca-consumer-latest.noarch.rpm"
+    state: present
+  register: r_install_cert
+  until: not r_install_cert.failed
```
jkupferer, line 28: When run from a server that does not have `/etc/pki/ca-trust/source/anchors/katello-server-ca.pem` this task fails with the error:
```
Failure downloading https://labsat.opentlc.com/pub/katello-ca-consumer-latest.noarch.rpm, Request failed: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:897)>
```

The ca-trust anchor is provided by this package, so it seems the only way this works is if it is run on a host that has already previously installed this package.

The ca-trust file will not show as owned by this package because it is installed during post scripts:

```
# rpm -qf /etc/pki/ca-trust/source/anchors/katello-server-ca.pem
error: file /etc/pki/ca-trust/source/anchors/katello-server-ca.pem: No such file or directory
# rpm -qpl katello-ca-consumer-latest.noarch.rpm
/usr/bin/katello-rhsm-consumer
# rpm -qp --scripts katello-ca-consumer-latest.noarch.rpm
postinstall scriptlet (using /bin/sh):
/bin/bash /usr/bin/katello-rhsm-consumer
postuninstall scriptlet (using /bin/sh):
test -f /etc/rhsm/rhsm.conf.kat-backup && command cp /etc/rhsm/rhsm.conf.kat-backup /etc/rhsm/rhsm.conf
```

jkupferer:
The basic problem seems to be that when downloading the package we are trying to validate the `labsat.opentlc.com` certificate using a certificate that is installed by this same package.

We could disable TLS checks on download, or download without TLS, but then we must check that the package hasn't been tampered with by an attacker. This could be done with GPG, but this package isn't signed with a GPG key.

If we ignore the threat, then we are downloading an RPM without any security checks and then executing any scripts in that RPM as root on the host. The attack vector is trivial.

The solution, I think, is to either have the client provide the CA cert through their agnosticv variables and/or to add the CA cert to this role in agnosticd for any satellite infrastructure we manage. We already have secure mechanisms for communicating with GitHub to download the cert and variables, so this should be fairly simple.

jkupferer:
@marcosmamorim Please see this PR and let me know what you think?
https://github.com/redhat-cop/agnosticd/pull/1649

PR #1650, opened 2020-04-27
Deploy minio operator in integreatly ocp4 workload

dimitraz:
##### SUMMARY
Deploys the minio operator as part of the pre-workload tasks
https://issues.redhat.com/browse/INTLY-7040

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/ansible/configs/ocp-workloads/ocp-workload.yml
+++ b/ansible/configs/ocp-workloads/ocp-workload.yml
@@ -17,7 +17,17 @@
     set_fact:
       ansible_python_interpreter: "/opt/virtualenvs/k8s/bin/python"
 
-  - name: Run workload
+  - name: Run single workload
     include_role:
       name: "{{ ocp_workload }}"
     when: ocp_workload is defined
+
+  - name: Run multiple workloads
+    when:
+      - ocp_workloads is defined
+      - ocp_workloads | length > 0
+    loop: "{{ ocp_workloads }}"
+    loop_control:
+      loop_var: _ocp_workload
+    include_role:
+      name: "{{ _ocp_workload }}"
--- a/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
+++ b/ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
@@ -1,0 +1,5 @@
+---
+# override ocp workloads to include prerequisite roles
+ocp_workloads:
+  - ocp4-workload-integreatly-minio
+  - ocp4-workload-integreatly
--- a/ansible/roles/ocp4-workload-integreatly-minio/defaults/main.yml
+++ b/ansible/roles/ocp4-workload-integreatly-minio/defaults/main.yml
@@ -1,0 +1,44 @@
+---
+become_override: False
+ocp_username: opentlc-mgr
+silent: False
+
+# If your workload needs customization options provide variables
+# in a dictionary. The dictionary should be named "<role name>_defaults".
+# So because this example workload is called "ocp-workload-example" the dictionary
+# should be named "ocp_workload_example_defaults". Note that the dictionary can not
+# include "-" characters, so replace all "-" with "_" instead.
+#
+# You can override the defaults as parameters to the Ansible run that runs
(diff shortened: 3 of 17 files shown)


ansible/roles/ocp4-workload-integreatly/defaults/main.yml
```diff
@@ -21,7 +21,17 @@ silent: False
 # (if defined).
 #
 # The logic in the workload should only use the dictionary "<role name>".
```
aidenkeating, line 23: Based on this comment, should these variables all be under the `ocp4_workload_integreatly_defaults` dict?

dimitraz: I didn't think we'd want those as override params :thinking:

aidenkeating: Yep, ignore

dimitraz:
@wkulhanek @pmccarthy mind taking a look when you get a chance?

ansible/roles/ocp4-workload-integreatly/defaults/main.yml
```diff
@@ -3,6 +3,19 @@ become_override: False
 ocp_username: opentlc-mgr
 silent: False
 
+# variables for the minio operator deployment
+minio_operator_ns: "redhat-rhmi-minio-operator"
```
wkulhanek, line 7: If these are part of the Integreatly workload would you mind sticking these variables into the dictionary?
```
ocp4_workload_integreatly_defaults:
  minio_operator_ns
  ...
```

ansible/roles/ocp4-workload-integreatly/templates/clusterrole.j2
```diff
@@ -0,0 +1,52 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: {{ minio_operator_role }}
```
wkulhanek, line 4: And of course here you'd use `{{ ocp4-workload-integreatly.minio_operator_role }}`

ansible/roles/ocp4-workload-integreatly/defaults/main.yml
wkulhanek: But really I'd stick that into its own workload. Something like ocp4-workload-integreatly-minio. And then run that before running the integreatly workload.

dimitraz: Sure, thanks for the feedback. We've updated this now to a separate workload

ansible/configs/ocp-workloads/ocp-workload.yml
```diff
@@ -17,7 +17,17 @@
...
+
+  - name: Run multiple workloads
+    when:
+      - ocp_workloads is defined
+      - ocp_workloads | length > 0
+    loop: "{{ ocp_workloads }}"
+    loop_control:
+      loop_var: _ocp_workload
+    include_role:
+      name: "{{ _ocp_workload }}"
```
aidenkeating, line 33: @wkulhanek Do you see an issue with having something like this in here? It allows us to test overriding `ocp_workloads` as well as a single `ocp_workload`. Which is useful as for local dev we want to run Minio before the main `integreatly` role

ansible/roles/ocp4-workload-integreatly/sample_vars/ocp4_workload_integreatly_vars.yml
```diff
@@ -0,0 +1,5 @@
+---
+# override ocp workloads to include prerequisite roles
+ocp_workloads:
+  - ocp4-workload-integreatly-minio
+  - ocp4-workload-integreatly
```
aidenkeating, line 5: Wasn't sure where to put this, but it's useful for local dev and it's the order we'll need the roles run in in the actual workshop on RHPDS.

ansible/configs/ocp-workloads/ocp-workload.yml
wkulhanek: @aidenkeating That's a good idea. Had written up a whole thing.... until I realized that the variable is different( workloads vs workload) (plural vs singular). I think this is good.

PR #1662, opened 2020-04-30
Fixed bug in studentvm_amq_client role. Renamed all student roles to …

wkulhanek:
…use underscores. Change role directory to use underscore. Changed all roles to use flat variables rather than dictionary.


ansible.cfg
```diff
@@ -1,7 +1,7 @@
 [defaults]
 nocows                  = 1
 # Don't forget to update other ansible.cfg files like:  ansible/configs/linklight/ansible.cfg
-roles_path              = dynamic-roles:ansible/dynamic-roles:roles-infra:ansible/roles-infra:roles:ansible/roles:ansible/roles-studentvm:roles-studentvm
+roles_path              = dynamic-roles:ansible/dynamic-roles:roles-infra:ansible/roles-infra:roles:ansible/roles:ansible/roles_studentvm:roles_studentvm
```
stencell, line 4: Did we want to drop the `ansible/` as @fridim suggested or leaving it for now?

ansible/roles_studentvm/studentvm_amq_clientvm/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# Websockify Python Package Version
+studentvm_amq_clientvm_websockify_version: "0.9.0"
```
stencell, line 3: Nit: why have `clientvm` name in the var...it makes it unnecessarily long.

ansible.cfg
wkulhanek: I'll let @fridim do that. :-)

PR #1677, opened 2020-05-01
Add support to also deploy KNative Eventing (Tech Preview). Disabled …

wkulhanek:
…by default.


ansible/roles_ocp_workloads/ocp4_workload_serverless/tasks/remove_workload.yml
```diff
@@ -8,6 +8,15 @@
     name: knative-serving
     namespace: knative-serving
 
+- name: Remove Knative Eventing Installation
+  when: ocp4_workload_serverless_install_eventing | bool
```
stencell, line 12: Shouldn't this be `not`?

wkulhanek: It's the remove workload. So when this is true that means that eventing was installed.

PR #1679, opened 2020-05-02
Refactoring of the ocp-workload-fuse-ignite to target ocp4 and the latest Fuse Online operator

btison:
Workload for the Fuse Foundations Part I course.


ansible/roles/ocp-workload-fuse-ignite/tasks/workload.yml
```diff
@@ -1,178 +1,159 @@
...
+- name: create limitrange in namespace
+  k8s:
+    state: present
+    namespace:  "{{ ocp_project }}"
+    definition: "{{ lookup('template', role_path ~ '/templates/limitrange.j2' ) | from_yaml }}"
+
+- name: create CRD for Fuse Online
+  k8s:
+    state: present
+    definition: "{{ lookup('file', role_path ~ '/files/syndesis-crd.yml' ) | from_yaml }}"
```
wkulhanek, line 25: As long as they're all files or templates you can simplify quite a bit by using a `loop`. Like here (templates example): https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_nexus_operator/tasks/workload.yml#L10

PR #1680, opened 2020-05-03
StudentVM Fixes

stencell:
##### SUMMARY
Add `istioctl` to ocp4 tools.
Fix var names in `studentvm_user` role.
Add default `purpose` var to `studentvm` config.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
studentvm
studentvm_ocp4
studentvm_user

### DIFF
--- a/ansible/roles_studentvm/studentvm_ocp4/defaults/main.yml
+++ b/ansible/roles_studentvm/studentvm_ocp4/defaults/main.yml
@@ -16,6 +16,8 @@ studentvm_ocp4_helm_version: '3.1.1'
 studentvm_ocp4_tkn_version: '0.8.0'
 # https://github.com/knative/client/releases
 studentvm_ocp4_kn_version: 'v0.12.0'
+# https://github.com/istio/istio/releases
+studentvm_ocp4_istioctl_version: '1.5.2'
 # https://github.com/operator-framework/operator-registry/releases
 studentvm_ocp4_opm_version: 'v1.11.1'
 # Source to Image URL (can't use just version because of random hash in the file name)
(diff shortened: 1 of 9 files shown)


ansible/roles_studentvm/studentvm_ocp4/defaults/main.yml
```diff
@@ -16,6 +16,8 @@ studentvm_ocp4_helm_version: '3.1.1'
 studentvm_ocp4_tkn_version: '0.8.0'
 # https://github.com/knative/client/releases
 studentvm_ocp4_kn_version: 'v0.12.0'
+# https://github.com/istio/istio/releases
+studentvm_ocp4_istioctl_version: '1.5.2'
```
wkulhanek, line 20: I had left that out on purpose. According to istio Chat istioctl is not a part of OpenShift.... But honestly don't feel strongly about it.

ansible/roles_studentvm/studentvm_user/defaults/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # Account name for the student user
-studentvm_user_name: lab-user
+studentvm_user_user_name: lab-user
```
wkulhanek, line 3: That change was intentional. _user_user_ just doesn't read right.

ansible/roles_studentvm/studentvm_ocp4/defaults/main.yml
stencell: Mmmm, I would prefer we add it. Tekton, knative, etc. are not a "core" part of the platform either since you have to install them. OSSM comes with OpenShift, so that is close enough for me.

Plus it would be silly to have a role just to install that one package.

ansible/roles_studentvm/studentvm_user/defaults/main.yml
stencell: Ok, this was broken because of the defaults referenced other out of date vars. I'll change these back and fix the studentvm config defaults.

stencell: fixes pushed.

PR #1696, opened 2020-05-06
add software_to_deploy

stencell:
##### SUMMARY
@wkulhanek I think this belongs in the `default_vars` for this config. I don't see us ever deploying the `ocp4-cluster` config without...OCP4. It is in the samples already, but there is no reason not have it as a default for everything.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-cluster

### DIFF
--- a/ansible/configs/ocp4-cluster/default_vars.yml
+++ b/ansible/configs/ocp4-cluster/default_vars.yml
@@ -50,6 +50,9 @@ all_use_python3: true
 # This might be enabled when we have solvers to run or graders for ILT
 install_ftl: false
 
+# This will use the software_playbook phase to install OpenShift 4 in the new environment
+software_to_deploy: openshift4
+
 # TODO: Decide on whether to use sat or give access to repos directly with key
 # This will tell Agnosticd to use either:
 # satellite, rhn, or file for repos


fridim:
`software_to_deploy` is one of those special variables that need to be set as extra-vars.

- `env_type`
- `cloud_provider`
- `software_to_deploy`

The reason is because they are used with `import_playbook` and that is static. As far as i know there is no possible way in ansible to use a dynamic variable (defined using `include_vars`) in `import_playbook`.

Can we use this PR to document that instead ?

stencell:
@fridim I don't understand. We do set those other 2 in the default_vars for a config. Are you saying we _cannot_ set `software_to_deploy` there?

wkulhanek:
Correct. I actually had those in the default vars and it didn’t work....

stencell:
Hmm, ok...so we have to define `software_to_deploy` in every agnosticV config that will leverage this functionality rather than being able to include it with the base config?

If that is the case, I'll close this PR and leave the var set in agnosticV

wkulhanek:
Yeah. That's why all the ocp4-cluster (and studentvm) sample_vars have a "mandatory vars" section at the top.

PR #1702, opened 2020-05-07
Create security groups only if they are used in 'instances'

fridim:
##### SUMMARY
If a config has security groups and use those, the default security groups are still create, for nothing.

The commits in this PR addresses that and make sure only used security groups are created in the templates.

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request


stencell:
Tested with disconnected config that has unused security groups, and LGTM.

stencell:
@fridim Do you want to merge latest changes from dev in?

PR #1706, opened 2020-05-07
[WIP] Add instance tags

marcosmamorim:
##### SUMMARY
this PR adds new metadata for instances installed on openstack

* Including new defaults tags course_name, platform, purpose, catalog_item

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
infra-osp-template-generate

### DIFF
--- a/ansible/roles-infra/infra-osp-template-generate/defaults/main.yaml
+++ b/ansible/roles-infra/infra-osp-template-generate/defaults/main.yaml
@@ -74,3 +74,7 @@ default_metadata:
   Project: "{{ project_tag }}"
   guid: "{{ guid }}"
   env_type: "{{ env_type }}"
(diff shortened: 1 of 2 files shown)


stencell:
Answering my own question. So the additional combine happens in the template: https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2#L178-L191

We're combining `instance.metadata`, `instance.tags`, and `cloud_tags`. That seems like a lot of different places to allow people to add tags. Surely we can look at standardizing a little bit for going forward?

marcosmamorim:
Yes, you're right, I don't need to merge the metadata.
I saw a problem because we are able to add the default metadata only if `intance.tag` is defined. I think the default metadata must exist.

PR #1720, opened 2020-05-12
Send documentation link and fix issue with port order

agonzalezrh:
1) Send a mail if documentation_link variable in agnosticV exists

2) The previous commit was not working properly because is was replacing first 443

### DIFF
--- a/ansible/configs/osp-migration/infra.yml
+++ b/ansible/configs/osp-migration/infra.yml
@@ -109,8 +109,8 @@
         msg: >-
           user.info: - {{ item.1
           | replace(80,"http")
-          | replace(443,"https")
-          | replace(8443,"https") }}://{{ item.0.key
+          | replace(8443,"https")
+          | replace(443,"https") }}://{{ item.0.key
           | regex_replace('REPL', guid) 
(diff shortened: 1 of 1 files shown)


ansible/configs/osp-migration/infra.yml
```diff
@@ -101,6 +101,10 @@
         openstack_project_services: {}
       when: openstack_project_services == ""
 
+    - debug:
+        msg: "Access to the documentation: {{ document_link }}"
+      when: document_link | default('') != ''
```
fridim, line 106: Didn't you mean to use `agnosticd_user_info` module here ?

ansible/configs/osp-migration/infra.yml
```diff
@@ -101,6 +101,10 @@
...
+      
     - debug:
          msg: "user.info: you can connect to this environment from the following hostnames and services:"
```
fridim, line 109: same here, `agnosticd_user_info` is to be prefered

ansible/configs/osp-migration/infra.yml
agonzalezrh: updated, could you check?

PR #1725, opened 2020-05-13
Clean-up related to Azure environment variables

vincepower:
##### SUMMARY
Cleaned up environment variables for roles and cloud_provider related to provisioning and destroying on Azure

Also found a change to another role caused an error in ocp4-cluster config when used on Azure or GCP since they use the RHEL8 image from their respective marketplace and not a custom gold image.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
config/aro
cloud_provider/azure
roles-infra/*azure*
roles/set-repositories
roles/satellite

##### ADDITIONAL INFORMATION
Tested with aro3, aro4, ocp4-cluster, and ocp-workshop using Azure

### DIFF
--- a/ansible/cloud_providers/azure_infrastructure_deployment.yml
+++ b/ansible/cloud_providers/azure_infrastructure_deployment.yml
@@ -8,16 +8,17 @@
     - step001
     - deploy_infrastructure
   environment:
+    # These first four are only used by Ansible modules
     AZURE_CLIENT_ID: "{{azure_service_principal}}"
     AZURE_TENANT: "{{azure_tenant}}"
     AZURE_SECRET: "{{azure_password}}"
     AZURE_SUBSCRIPTION_ID: "{{azure_subscription_id}}"
     # AZURE_CONFIG_DIR: create a specific config dir for this stack to allow concurrent access
+    # This variable is also used by the CLI
     AZURE_CONFIG_DIR: "/tmp/.azure-{{project_tag}}"
+    PATH: "/tmp/.azure-{{project_tag}}:/usr/local/bin:/usr/bin"
   tasks:
     - name: Ensure az is installed
-      environment:
-        PATH: /usr/bin
       command: which az
       register: az_result
 
@@ -50,8 +51,6 @@
         -u "{{azure_service_principal}}"
         -p {{azure_password}}
         --tenant {{azure_tenant}}
-      environment:
-        PATH: /usr/bin
       tags:
         - validate_azure_template
         - create_inventory
@@ -119,8 +118,6 @@
         - validate_azure_template
 
     - name: Validate arm template
-      environment:
-        PATH: /usr/bin
       command: >-
         az group deployment validate
         --template-file {{t_dest}}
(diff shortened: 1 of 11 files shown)


ansible/cloud_providers/azure_infrastructure_deployment.yml
```diff
@@ -8,16 +8,17 @@
...
   environment:
+    # These first four are only used by Ansible modules
     AZURE_CLIENT_ID: "{{azure_service_principal}}"
     AZURE_TENANT: "{{azure_tenant}}"
     AZURE_SECRET: "{{azure_password}}"
     AZURE_SUBSCRIPTION_ID: "{{azure_subscription_id}}"
     # AZURE_CONFIG_DIR: create a specific config dir for this stack to allow concurrent access
+    # This variable is also used by the CLI
     AZURE_CONFIG_DIR: "/tmp/.azure-{{project_tag}}"
+    PATH: "/tmp/.azure-{{project_tag}}:/usr/local/bin:/usr/bin"
```
fridim, line 19: Can  you use `output_dir` for that instead ?

With output_dir, our script always make sure everything is cleaned up from the deployer host, whatever happens to the process.

Whereas, if you use something in /tmp/, it's not guaranted.

If it's a secret, you can put it in `{{ output_dir }}/secrets`, it will not be save in the tarball (used for troubleshooting sometimes)

ansible/roles/satellite-register-host/tasks/register.yml
```diff
@@ -19,6 +19,19 @@
   package:
     name: katello-ca-consumer-*.noarch
     state: absent
+  when:
+   - cloud_provider != 'azure'
+   - cloud_provider != 'gcp'
```
fridim, line 24: I don't understand the link between satellite-register-host  and the cloud provider.

Can you explain ?

vincepower: The images used as part of the Azure and GCP marketplace offerings have no Satellite related packages installed currently. 
Since this could change in the future, I've excluded them from the mandatory remove and adds and optional one right below it, so at some point if the image starts to include Satellite it will still get cleaned up.
Not ideal, but if we want to consistently use Satellite I don't know another way without having yet more images to maintain.

ansible/cloud_providers/azure_infrastructure_deployment.yml
vincepower: Sure, I'll fix this so it uses output_dir.

vincepower: Its not the output_dir, output_dir doesn't seem to be defined in the configs I was working with. (aro and ocp4-cluster).

Is output_dir defined every time by default through RHPDS like guid?

I'm ok with making this a variable instead of hardcoding /tmp/, but I want to make sure I'm using something that will exist every time.

fridim: If it's not defined in the config, it should.
`output_dir` should be always defined.
Yes, please don't use `/tmp/` as it could fill up the `/tmp/` of the deploy host.

@rhjcd  FYI. 
Why did this PR was merged before this was resolved ?

ansible/roles/satellite-register-host/tasks/register.yml
fridim: But if the package is not there, then nothing will be done, right ?

IMO those condition are not mandatory and create confusion.

PR #1737, opened 2020-05-14
Update Quay Operator Workload for Red Hat Quay Operator 3.3.0 and Qua…

wkulhanek:
…y 3.3.0


ansible/roles_ocp_workloads/ocp4_workload_quay_operator/defaults/main.yml
```diff
@@ -9,31 +9,39 @@ ocp4_workload_quay_operator_project: quay-enterprise
 ocp4_workload_quay_operator_project_display: Red Hat Quay Registry
 ocp4_workload_quay_operator_name: quay
 
+# Channel to subscribe to
+ocp4_workload_quay_operator_channel: "quay-v3.3"
+
 # Quay starting CSV. Leave empty for current CSV
-# starting_csv: "quay.v1.0.2"
+# starting_csv: "red-hat-quay.v3.3.0"
```
stencell, line 16: I'm torn on this and prefer stability and reliability of the workload above all else. Being on whatever the "latest" is has a chance to cause instability and unexpected differences.

Being on on a specific version runs the risk of it somehow disappearing from OLM, but I hope that this keeps decreasing over time.

Is it maybe a better approach to just leave it latest as you have it here and then pin it to a CSV in agnosticV?

wkulhanek: Theoretically pinning the channel should do the trick as well.

PR #1741, opened 2020-05-15
Update config:smart-management

miteshget:
##### SUMMARY
Config update

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
config: smart-management

### DIFF
--- a/ansible/configs/smart-management/files/repos_template.j2
+++ b/ansible/configs/smart-management/files/repos_template.j2
@@ -36,21 +36,16 @@ baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-server-rhscl
 enabled=1
 gpgcheck=0
 
-[rhel-7-server-ansible-2.6-rpms]
-name=Red Hat Enterprise Ansible 2.6
-baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-7-server-ansible-2.6-rpms
+[rhel-7-server-ansible-2.8-rpms]
+name=Red Hat Enterprise Ansible 2.8
+baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-7-server-ansible-2.8-rpms
 enabled=1
 gpgcheck=0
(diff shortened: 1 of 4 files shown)


ansible/configs/smart-management/files/repos_template.j2
```diff
@@ -36,21 +36,16 @@ baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-server-rhscl
...
-[rhel-7-server-extras-rpms]
-name=Red Hat Enterprise Linux Extra RPMs
-baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-7-server-extras-rpms
-enabled=1
-gpgcheck=0
 
-[rhel-7-server-satellite-6.4-rpms]
-name=Red Hat Enterprise Satellite 6.4
-baseurl={{own_repo_path}}/satellite/{{satellite_repo_version}}/rhel-7-server-satellite-6.4-rpms
+[rhel-7-server-satellite-6.7-rpms]
```
fridim, line 46: version is hardcoded, can you use the `satellite_repo_version` here instead ?

miteshget: This is there for purpose because whole thing is set based on this version for example rhel_repos and I have mentioned  in README.adoc that this is default version. Let me know, I will change if this is not used as general practice.

fridim: Here is why i'm asking:

What would happen to someone  who wants to use this config with 6.4 ?

He/she will have to update:

```yaml
satellite_repo_version: '6.4'
```

Update also, rhel_repos

 **and**  manually the repos_template.j2 ?

miteshget: Understood, I made repository dynamic based on two variables {satellite_repos & rhel_repos} as you wanted. 

This config  installs tower and satellite that is why we need to separate repo list {satellite,rhel}_repos.

PR #1745, opened 2020-05-18
rhmi 3scale service discovery

aidenkeating:
##### SUMMARY

update integreatly workload to include per-user fuse online service discovery in 3scale. even though [1] will still cause an issue in 3scale 2.8. also ensure that pod quotas are not hit on large user count installations.

[1] https://issues.redhat.com/browse/THREESCALE-4851

@ciaranRoche @dimitraz @pmccarthy mind taking a look? i'll give you all write permissions to this branch

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

role: `roles_ocp_workloads/ocp4_workload_integreatly`

##### ADDITIONAL INFORMATION

commit a05891e1d15494b971283d7d3236a800d21b5b31

increase default node pod quota to 500

ensure the default node pod quota is updated for the openshift
cluster, to avoid hitting pod quota limits upon rhmi installation.

verification:
- run the installation
- wait for the installation to get to rhmi cr checks
- once the rhmi cr checks begin, run [1] and ensure each node has
  .status.capacity.pods equal to '500'

[1] 'oc get node --selector=node-role.kubernetes.io/worker='' -o yaml'

commit fdc7531b3c2077b98715e6f94d9c21df9ac8dacc

ensure 3scale service discovery is enabled in per-user fuse

set management url in per-user syndesis custom resources to ensure
that service discovery is enabled for fuse online by default.

verification:
- run the install
- ensure the per-user syndesis crs, format 'evals{N}' have the
  3scale management url feature set and that the url points to the
  per-user 3scale tenant admin url
- create an api provider integration in the per-user fuse online
  instance, [1] can be used as an example openapi spec
- ensure that after the app is published, 3scale integration is set
  to enabled by default in the ui. there should be a disable
  discovery button on the view integration page in the ui
- ensure that after the integration is published, a service exists
  in the per-user fuse online namespace, named after the
  integration, with labels referencing 3scale

[1] https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v3.0/petstore.yaml

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_integreatly/defaults/main.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_integreatly/defaults/main.yml
@@ -5,15 +5,39 @@ silent: False
 
 ocp4_workload_integreatly_documentation_link: https://docs.google.com/document/d/16mFcIjyykwY9oR0K7xRPZ0nAWm6AnhGvEBYISDlakTQ
 
+# Machine Config
+# We need to override the machine pool config to allow for a larger
+# maximum node quota. This is because on clusters with 50 users, 50
+# Fuse Online instances are created, which contain many pods.
+# On these large clusters, resources are not the bottleneck. The default quotas
+# are, as they stop most resources from ever being used.
+ocp4_workload_integreatly_machineconfigpool_name: worker
+ocp4_workload_integreatly_machineconfigpool_label_value: "true"
+
+ocp4_workload_integreatly_kubeletconfig_name: rhmi-max-pod-override
+ocp4_workload_integreatly_kubeletconfig_template: kubeletconfig-pods.yml.j2
+ocp4_workload_integreatly_kubeletconfig_pods_max: 500
+ocp4_workload_integreatly_kubeletconfig_pods_core: 500
+
 # Operator
 ocp4_workload_integreatly_user_count: 50
 ocp4_workload_integreatly_tmp_dir: /tmp
 ocp4_workload_integreatly_prefix: redhat-rhmi-
 ocp4_workload_integreatly_namespace: "{{ ocp4_workload_integreatly_prefix }}operator"
+# The actual release version. This should be treated as the only version input
+# an end user should need to override in normal circumstances.
+# This should be updated on each new release. It should always be the latest release.
 ocp4_workload_integreatly_release: "2.2.0"
-# RC's are used until full release is available
-# ocp4_workload_integreatly_git_ref: "v{{ ocp4_workload_integreatly_release }}"
-ocp4_workload_integreatly_git_ref: v2.2.0-rc1
+# A map for ocp4_workload_integreatly_release to an actual git ref
+# This allows the version itself and the git ref to differ, which is common
+# as the git ref is often prefixed with a 'v' and during release testing will
+# also contain a '-rc{N}' type suffix.
+# This mapping should be added to for each new release.
+ocp4_workload_integreatly_git_ref_map:
+  "2.2.0-rc2": v2.2.0-rc2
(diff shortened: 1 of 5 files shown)


ansible/roles_ocp_workloads/ocp4_workload_integreatly/tasks/workload.yml
```diff
@@ -3,6 +3,40 @@
...
+# worker nodes.
+- name: Check worker MachineConfigPool has enough available nodes
+  k8s_facts:
+    api_version: machineconfiguration.openshift.io/v1
+    kind: MachineConfigPool
+    name: "{{ ocp4_workload_integreatly_machineconfigpool_name }}"
+  register: _get_machinepoolconfig_worker
+  retries: 20
+  delay: 60
+  until: _get_machinepoolconfig_worker.resources[0].status.readyMachineCount in [3, _get_machinepoolconfig_worker.resources[0].status.machineCount]
```
aidenkeating, line 38: we could make this exact, instead of waiting for 3 minimum. it'll just take longer on larger clusters

ansible/roles_ocp_workloads/ocp4_workload_integreatly/defaults/main.yml
```diff
@@ -5,15 +5,39 @@ silent: False
...
 ocp4_workload_integreatly_release: "2.2.0"
-# RC's are used until full release is available
-# ocp4_workload_integreatly_git_ref: "v{{ ocp4_workload_integreatly_release }}"
-ocp4_workload_integreatly_git_ref: v2.2.0-rc1
+# A map for ocp4_workload_integreatly_release to an actual git ref
+# This allows the version itself and the git ref to differ, which is common
+# as the git ref is often prefixed with a 'v' and during release testing will
+# also contain a '-rc{N}' type suffix.
+# This mapping should be added to for each new release.
+ocp4_workload_integreatly_git_ref_map:
```
aidenkeating, line 36: this is just a workaround for release candidates and it gets rid of the assumption that the git ref will always be a release version prefixed with `v`

pmccarthy:
👀

ansible/roles_ocp_workloads/ocp4_workload_integreatly/defaults/main.yml
pmccarthy: We probably shouldn't have any release tags specified in this workload in general as it will be read in from a variable linked to the version dropdown in the workshop submission form. For development purposes we'll just need to ensure that we're specifying a release version using `-e ocp4_workload_integreatly_release=v2.2.0`

pmccarthy: We'll probably just map `ocp4_workload_integreatly_release` to the version dropdown value

aidenkeating: @pmccarthy The reason I'd added this is because the git ref (tag) and the operator release version differ:

- For the git ref `v2.2.0`, `2.2.0`, is the operator release version 
- For the git ref `v2.2.0-rc3`, `2.2.0` is the operator release version
- For the git ref `master`, `2.2.0` is the operator release version

As we really only care about proper releases, I'm happy to let `ocp4_workload_integreatly_release` be the actual git ref and then filter out the `v` for determining the actual operator release version by default though.

So, user selects `v2.0.0` and:
- Role assumes `v2.0.0` is the git ref
- Role assumes `2.0.0` (filter out the `v`) is the operator release version

This would break for RC's, but as you've said, we can just override it externally locally.

austinmartinh:
Verified

PR #1762, opened 2020-05-20
[osp-migration] Create objects with user instead opentlc-mgr

agonzalezrh:
Creating keypair with a opentlc-mgr reachs a quota limit

This reverts: https://github.com/redhat-cop/agnosticd/commit/a26d7d7ca06a82342eb16aab0e4f5efb90a584cf


jkupferer:
Can we fix the `api_pass` variable to use variable naming convention to add role or config name prefix?

PR #1763, opened 2020-05-21
Move user create/delete out of HOT

stencell:
##### SUMMARY
In the current version of the OSP cloud provider, a user and keypair is created as part of instantiating a HOT. This HOT is created as an admin user in the regular user project. But this means that the keypair it creates as part of the HOT is associated with the admin user. This was causing the admin user to have too many keypairs associated with it and exceeding the allowed quota.

This PR moves the user creation and role association out of the HOT closer to the project creation tasks. The keypair generation is left in the HOT so that it can still be consumed via the outputs.

The resource deletion was also updated to make sure the keypair was deleted before the user and finally the project.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
infra-osp-project-create
infra-osp-resources-destroy
infra-osp-template-create
infra-osp-template-generate


stencell:
Moving this to WIP. Need to make the key deletion in the destroy_env more idempotent or it fails on runs where it is not found.

stencell:
@fridim This should be better now and no longer fails on the keypair delete if there is nothing there. It will also pick up any other keypairs for that user so we don't strand them.

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - environment:
     OS_AUTH_URL: "{{ osp_auth_url }}"
-    OS_USERNAME: "{{ osp_auth_username }}"
+    OS_USERNAME: "{{ guid }}-user"
```
fridim, line 4: When destroying, don't we want to use the admin user ?

ansible/roles-infra/infra-osp-resources-destroy/tasks/project.yml
```diff
@@ -2,9 +2,38 @@
...
+      set_fact:
+        osp_user_uuid: "{{ r_osp_user_info | json_query(uuid_query) }}"
+      vars:
+        uuid_query: ansible_facts.openstack_users[].id|[0]
+
+    # These can't use the os_ module since impersonating the user is only
+    # available with the nova client here.
+    - name: Get keypairs for user
+      shell: nova keypair-list --user {{ osp_user_uuid }} | grep ssh | awk '{ print $2 }'
+      register: r_user_keypairs
```
fridim, line 25: i'd prefer if we can `-o json` and get the ssh using the `json_query` filter instead of `grep` + `awk`

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
```diff
@@ -14,7 +14,7 @@ resources:
     properties:
       name: {{ guid }}-infra_key
       save_private_key: true
-
```
fridim, line 17: Can't we just add the user in the `OS::Nova::KeyPair` resource ?


https://docs.openstack.org/heat/pike/template_guide/openstack.html#OS::Nova::KeyPair-prop-user

```yaml
  {{ guid }}-infra_key:
    type: OS::Nova::KeyPair
    properties:
      name: {{ guid }}-infra_key
      user: {{ guid }}-project_user
      save_private_key: true
    depends_on:
      - {{ guid }}-project_user
```

stencell:
> Here my review.
> 
> We need to make sure we don't break this documentation: https://github.com/redhat-cop/agnosticd/blob/development/docs/First_OSP_Env_walkthrough.adoc
> 
> Did you try it with your changes?

I used Student VM as my primary test. I might try to find time to test with this, but I don't see how it will change. None of the vars or resulting objects change...we just create them in a different order with a different context under the covers.

ansible/roles-infra/infra-osp-resources-destroy/tasks/project.yml
stencell: I already went down this path. This was my last option. Since this uses `nova` cli instead of `openstack`, there is no modifying the output into something prettier that I was able to find.

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
stencell: Huh, this is a good catch. I'm guess I missed this and it would potentially make things simpler.  I'll set up a test to make sure it is cleaned up correctly and if so I can simplify this change and keep everything as part of the tempalte.

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
stencell: This was a mistake on my part while experimenting that I didn't revert. Thanks for catching.

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
stencell: Ok, I remember why something like this doesn't work. This whole keypair-associated-to-user thing makes all this more complicated.

If I create the Heat stack as the admin user, I am doing it as the admin user...even though it is in the `{{ guid }}-user` project. Even though I can tell the Heat template to create the keypair AS or FOR another user, when my Heat stack goes to launch instances it does not have access to the keypair. Because I am launching the instances as admin...but I am trying to use a keypair from the user. Remember keypairs are 100% associated to user and not project or anything else.

But why don't we launch the Heat stack as the user as it is today? Chicken and egg...the user gets created with the Heat stack so I still have to create the Heat stack as admin. And then it is back to launching everything in the stack as admin as well.

Moving the user creation out of the Heat stack gives us the ability to create all the things as the user rather than admin. We just have to make sure we do the last bit of cleanup on the user account + keypair and everything else stays the same.

@fridim make sense?

stencell:
@fridim Can we wrap this up? Please take a look at the changes since the last review.

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
fridim: Yes it makes sense.

How urgent is this PR?  Since we're at moving things out of the heat template and changing their "owner", I'm thinking maybe we should switch gears and have 2 HOT templates:

1. One admin HOT that will create project, user, rolebindings, quotas, etc, like it's done here: https://github.com/redhat-gpe/novello-templates/blob/master/heat-templates/EMEA-PARTNERS-ansible_tower_summit-v1.0-bp/stack_admin.yaml
2. The user HEAT template.

IIUC:
- today, 1. is done by some ansible tasks (not heat) **and** the heat template. 2. is done in the heat template, as admin.  Which is confusing and wrong now that we discovered the issue with keypairs.
- with this PR, 1. is done entirely outside of the heat template as ansible tasks executed as admin user in our osp cloud-provider roles. and 2. is done in the heat template executed as user.

In the future, maybe we want to use HOT for the admin part as well since it would simplify the deletion.

Would that be a too big change to be done now ? Wdyt?

The changes in that PR are valid since they address the issue and fix them.

fridim:
> @fridim Can we wrap this up? Please take a look at the changes since the last review.

The changes in that PR seem valid and they address the issue and fix it.

I commented about another possibility, let me know what you think.

I'm happy to merge this and we create an issue to continue the discussion about 2 Heat templates, both for admin and user.

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
stencell: It is not urgent-urgent...but it should be done, especially before the next set of OpenStack clusters get deployed and used. I also didn't want this branch to get too far out of sync with what everything was tested against.

As for your suggestion, I'm not opposed, but not in favor either. I find HOT (and CFN and ARM and...) more complicated to use and maintain. I get the need for them when we are deploying a more complex set of resources that may be dependent on one another and have nuance in how they are instantiated and removed. 

However, for doing simple things like we are talking about here, I really find Ansible tasks much easier to create and maintain. They do the job well to create these resources. Teardown of these same resources is very predictable as well.

If you feel strongly about moving these into a HOT, I will do it...but I'm also quite satisfied with them being Ansible tasks.

ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2
```diff
@@ -35,7 +35,7 @@ resources:
     depends_on:
       - {{ guid }}-project_user
 {% endif %}
-
+#}
```
fridim, line 38: Let's remove that part instead of commenting it out.

ansible/roles-infra/infra-osp-resources-destroy/tasks/project.yml
```diff
@@ -2,9 +2,38 @@
...
+
+    - name: Delete user keypair
+      when: r_user_keypairs.stdout_lines | length > 0
+      shell: nova keypair-delete --user {{ osp_user_uuid }} {{ item }}
+      loop: "{{ r_user_keypairs.stdout_lines }}"
```
fridim, line 30: Why do we delete keypairs as admin here?

If they are part of the user Heat stack, when we delete the stack, they should be deleted there?

ansible/roles-infra/infra-osp-template-generate/tasks/generate_osp_template.yml
```diff
@@ -1,10 +1,10 @@
 - set_fact:
-    heat_user_password_generated: "{{lookup('password', '/dev/null length=12 chars=ascii_letters,digits')}}"
+    # heat_user_password_generated: "{{lookup('password', '/dev/null length=12 chars=ascii_letters,digits')}}"
```
fridim, line 2: Let's remove the line

ansible/roles-infra/infra-osp-template-generate/tasks/generate_osp_template.yml
```diff
@@ -1,10 +1,10 @@
...
     heat_master_template: "{{output_dir}}/{{ env_type }}.{{ guid }}.{{cloud_provider}}_cloud_master_template.yaml"
     heat_nested_template: "{{output_dir}}/{{ env_type }}.{{ guid }}.{{cloud_provider}}_cloud_nested_template.yaml"
 
-- set_fact:
-    heat_user_password: "{{ heat_user_password_provided | default(heat_user_password_generated) }}"
+# - set_fact:
+#     heat_user_password: "{{ heat_user_password_provided | default(heat_user_password_generated) }}"
```
fridim, line 7: same

ansible/roles-infra/infra-osp-resources-destroy/tasks/project.yml
stencell: We do not know the password for the user during teardown so I cannot do things as the user. We have to delete things as admin, that is the only credential we know. This allows us to delete the keypair "on behalf" of the user.

fridim: OK but when the admin deletes the user stack, the keys are gone, no?

stencell:
Tested successfully on both `studentvm` and `ocp4-cluster` configs.

PR #1795, opened 2020-05-25
Fix ImageStream issue.

DuncanDoyle:
##### SUMMARY
Fixes an ImagePull problem with the FSI Client Onboarding demo after images have been permanently moved to registry.redhat.io, which requires a login.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp-workload-fsi-client-onboarding-demo

##### ADDITIONAL INFORMATION
The script, like some other scripts that use RHPAM and BPM Suite, now checks for the availability of the IS in the OpenShift namespace, and if the IS is not there yet, adds it. We need to user the "openshift" namespace, as the builder service account in that namespace has pull access to registry.redhat.io.

### DIFF
--- a/ansible/roles/ocp-workload-fsi-client-onboarding-demo/defaults/main.yml
+++ b/ansible/roles/ocp-workload-fsi-client-onboarding-demo/defaults/main.yml
@@ -23,3 +23,8 @@ build_status_delay: 20
 
 deploy_status_retries: 15
 deploy_status_delay: 20
+
+
+bpms_version_tag: 6.4.x
+bpms_imagestreams_yml: https://raw.githubusercontent.com/jboss-container-images/jboss-processserver-6-openshift-image/{{bpms_version_tag}}/templates/processserver64-image-stream.json"
+bpms_imagestreams_tag: 1.0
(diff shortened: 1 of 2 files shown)


ansible/roles/ocp-workload-fsi-client-onboarding-demo/defaults/main.yml
```diff
@@ -23,3 +23,8 @@ build_status_delay: 20
 
 deploy_status_retries: 15
 deploy_status_delay: 20
+
+
+bpms_version_tag: 6.4.x
+bpms_imagestreams_yml: https://raw.githubusercontent.com/jboss-container-images/jboss-processserver-6-openshift-image/{{bpms_version_tag}}/templates/processserver64-image-stream.json"
+bpms_imagestreams_tag: 1.0
```
fridim, line 30: If applicable, can you please use the full qualified variable names ?

  for example:

- `ocp_workload_fsi_client_onboarding_demo_bpms_version_tag`
-  `deploy_status_delay`

This prevent collisions.

fridim: i meant  for second var:

-  `ocp_workload_fsi_client_onboarding_demo_deploy_status_delay`

PR #1803, opened 2020-05-26
Fix aws credential in ocp4-workshop destroy playbook

fridim:
##### SUMMARY
It happens that the `openshift-install destroy` comamnd loops forever with this kind of error messages:

```
time="2020-05-26T14:17:29Z" level=debug msg="search for and delete matching instances by tag matching aws.Filter{\"openshiftClusterID\":\"66564c10-bc6f-4c81-a373-719a03722447\"}"
time="2020-05-26T14:17:29Z" level=info msg="describe instances: UnauthorizedOperation: You are not authorized to perform this operation.\n\tstatus code: 403, request id: 0eb0b0e6-913d-4d2f-8abd-7f2e60b5f548"
```

This happens when the AWS credential file `.aws/credentials` is modified on the
clientVM between provision and destroy.

This commit will fix that by ensuring the credentials are passed as environment
variables. The ocp4-workshop being aws-only, it's OK to specify the environment variables there.

Also fix the logic of the condition of the task waiting for the PID of `openshift-install`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- ocp4-workshop config

##### ADDITIONAL INFORMATION
I tested this change with the following:

1. Deploy an ocp4-workshop
2. Intentionally corrupt the `.aws/credentials` file on the clientvm
3. Run the destroy playbook and see that it fails
4. Apply this change
5. Run again and see that it succeeds

### DIFF
--- a/ansible/configs/ocp4-ha-lab/destroy_env.yml
+++ b/ansible/configs/ocp4-ha-lab/destroy_env.yml
@@ -175,7 +175,7 @@
           register: r_check_openshift_install_running
           failed_when: r_check_openshift_install_running.rc == 0
           changed_when: false
-          until: r_check_openshift_install_running.failed
+          until: r_check_openshift_install_running is succeeded
           retries: 360
           delay: 10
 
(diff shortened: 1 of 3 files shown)


fridim:
> As an aside: students still have aws cli access to interrogate resources?

Yes. I left that untouched.

wkulhanek:
@newgoliath That's different. This is for workshop. Users don't get bastion access (usually) to the workshop. And if they do they should not need admin credentials.

PR #1805, opened 2020-05-27
Change to prepend local directories to PATH in bash profile

jkupferer:
##### SUMMARY

This PR updates three related bastion roles to change the `.bash_profile` to add `~/.local/bin` as well as reordering the additions to `PATH` so that `/usr/local/bin` and `~/bin` override executables in other locations.

This is particularly relevant for ansible usage, where `ansible` may be installed with `pip3 install --user --upgrade ansible` and so end up in `~/.local/bin`. In this case we want the user copy to override the system copy.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Workload roles `bastion`, `bastion-lite`, and `ocp-client-vm`.

### DIFF
--- a/ansible/roles/bastion-lite/files/bash_profile
+++ b/ansible/roles/bastion-lite/files/bash_profile
@@ -7,5 +7,5 @@ fi
 
 # User specific environment and startup programs
 
-export PATH=$PATH:$HOME/bin:/usr/local/bin:/usr/local/maven/bin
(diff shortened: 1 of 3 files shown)


stencell:
@jkupferer Another change went in recently (#1971 ) that got rid of the `bash_profile` file completely in these roles...so I think this PR is no longer needed in it's current form?

stencell:
@jkupferer Shall we close this?

PR #1836, opened 2020-06-03
added retry & delay to k8s blocks

austinmartinh:
JIRA: https://issues.redhat.com/browse/INTLY-8171


fridim:
This need to be patched. You miss the `until` condition for the retries to work.

see https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html

> You must set the until parameter if you want a task to retry. If until is not defined, the value for the retries parameter is forced to 1.

@austinmartinh

austinmartinh:
I misunderstood that, will do.

PR #1841, opened 2020-06-03
Revert "added retry & delay to k8s blocks"

fridim:
Reverts redhat-cop/agnosticd#1836


fridim:
right, closing, not needed anymore.

PR #1848, opened 2020-06-04
use k8s to deploy imagestream; include admin URL in user.info

jamesfalkner:
##### SUMMARY
Minor fix to make workload idempotent when deploying imagestream. Also added extra URL in `user.info` for username distribution admin console.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`ocp4-workload-ccnrd`

### DIFF
--- a/ansible/roles/ocp4-workload-ccnrd/tasks/post_workload.yml
+++ b/ansible/roles/ocp4-workload-ccnrd/tasks/post_workload.yml
@@ -10,7 +10,7 @@
   debug:
     msg:
     - "user.info: "
-    - "user.info: URL for attendees to get a username assigned to them and links to labs:"
+    - "user.info: URL for attendees to get a username assigned to them and get personaliized links to labs:"
     - "user.info: "
     - "user.info: http://get-a-username-labs-infra.{{ route_subdomain }}"
     - "user.info: "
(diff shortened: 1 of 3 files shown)


ansible/roles/ocp4-workload-ccnrd/tasks/post_workload.yml
```diff
@@ -15,6 +15,12 @@
     - "user.info: http://get-a-username-labs-infra.{{ route_subdomain }}"
     - "user.info: "
     - "user.info: You should share this URL (or a shortlink for it) -- It is all they will need to get started!"
+    - "user.info: "
```
stencell, line 18: Not required, but I'd like to see this old way updated to the new way if you are already updating. You can use the `agnosticd_user_info` module to print out `user.info` and preserve functionality today and also save the same as data that can be used in bookbag/Babylon eventually.

https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/ocp4-disconnected-osp-lab/post_software.yml#L10-L39

jamesfalkner: Thanks for the tip, will do next time I'm in here - what are the allowed keys for the `agnosticd_user_info.data` element?

stencell: `msg`
`data`
`user`

https://github.com/redhat-cop/agnosticd/blob/development/ansible/library/agnosticd_user_info.py

https://github.com/redhat-cop/agnosticd/blob/development/ansible/action_plugins/agnosticd_user_info.py

PR #1851, opened 2020-06-04
Update Preparing_your_workstation.adoc

amayagil:
Added AWS prerequisites

##### SUMMARY

Added AWS prerequisites

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/Preparing_your_workstation.adoc
+++ b/docs/Preparing_your_workstation.adoc
@@ -32,10 +32,12 @@ include::../tools/builds/readme.adoc[leveloffset=3]
 Please see the https://github.com/redhat-cop/agnosticd/blob/development/docs/First_OSP_Env_walkthrough.adoc[OpenStack tutorial]. 
 
 === Workstation setup for AWS
+This is a configuration guide of the packages and repos you need to install the tools
 
 ==== Prerequisites
-
-This is a configuration guide of the packages and repos you need to install the tools, but a previoulsy existing AWS account is needed.
+For AWS, AgnosticD setup will require:
+- Progrmmatic access account and password
+- Route53 DNS Domain
 
 ==== Steps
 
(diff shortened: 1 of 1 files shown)


docs/Preparing_your_workstation.adoc
```diff
@@ -32,10 +32,12 @@ include::../tools/builds/readme.adoc[leveloffset=3]
 Please see the https://github.com/redhat-cop/agnosticd/blob/development/docs/First_OSP_Env_walkthrough.adoc[OpenStack tutorial]. 
 
 === Workstation setup for AWS
+This is a configuration guide of the packages and repos you need to install the tools
 
 ==== Prerequisites
-
-This is a configuration guide of the packages and repos you need to install the tools, but a previoulsy existing AWS account is needed.
+For AWS, AgnosticD setup will require:
+- Progrmmatic access account and password
```
stencell, line 39: typo

docs/Preparing_your_workstation.adoc
```diff
@@ -57,58 +59,58 @@ The lint tools are optional but are recommended tools to check the quality of yo
...
 
 #Install pywinrm if you plan to deploy windows VMs
 #pip install pywinrm
 
 # Enable epel repositories for Ansible
-cd /tmp
-wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-sudo yum -y install `ls *epel*.rpm`
+$ cd /tmp
+$ wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
stencell, line 77: I'm not in favor of using EPEL for anything that is not absolutely necessary. What is the need for it here to prepare your workstation? I think this might be legacy documented steps since you can easily get ansible via pip now.

stencell:
I'm super meh on these instruction updates. It seems like there might be a bunch in here for how things were done a year or two ago. Have you run through all of this and validated that everything is necessary or that there is not a "more better" way to do it now?

docs/Preparing_your_workstation.adoc
tonykay: I agree. I actually think whilst it is more work we may need 3 "workstation" docs linked from a master: RHEL7, RHEL8/Fedora<x>, and Mac. Plus it may be better to be opinionated e.g. python3, pip for boot etc and I think we may want to include a requirements file for your "workstation" (archaic term?). @wilson-walrus you've also been looking at this?

wilson-walrus: I am also in favor of removing EPEL where not 100% needed. 

I think having a separate path for the various operating systems is a great idea too. I also think it would be beneficial to have a master list required packages for people what will want to install in their own way.

amayagil:
I've gone thru those steps and they work, I'm not aware of a better way to do it, I am not into developing agD, just testing docs to see if they can be easily followed.

Also agree in having different docs for different OS, in this case, I just could test the RHEL steps.

docs/Preparing_your_workstation.adoc
NonyNo3lle: @wilson-walrus @amayagil to discuss separately in a call before merging or closing. After all we are not using this doc for the training.

stencell:
Where are we at with this one? Is this update going to be abandoned or ?

@amayagil @wilson-walrus

PR #1872, opened 2020-06-08
Revert "Removing credentials so lab users can't access them"

stencell:
Reverts redhat-cop/agnosticd#1865

Changes that touch these common roles and configs need more attention and validated test results. I use this config daily and this broke the destroy for AWS using the `ec2` provider.

```
TASK [Call Role to destroy the OpenShift cluster] ***************************************************************************************
Monday 08 June 2020  08:33:59 -0700 (0:00:01.191)       0:00:18.918 ***********
ERROR! Unable to retrieve file contents
Could not find or access '/Users/nstephan/gpte/upstream-agnosticd/ansible/configs/ocp4-cluster/ec2_prereqs.yml' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

@vincepower @rhjcd @fridim


stencell:
Holding on this for a few while @vincepower looks for a fix.

vincepower:
Fixed in PR 1875

stencell:
Closing in favor of #1875

PR #1874, opened 2020-06-08
[WIP] [DONTMERGE] [osp-migration] Use uuid as project name and user created

agonzalezrh:
Instead to use the guid for the project name and username, it uses the uuid generated by CF as project name and username created

### DIFF
--- a/ansible/configs/osp-migration/destroy_env.yml
+++ b/ansible/configs/osp-migration/destroy_env.yml
@@ -14,11 +14,18 @@
     OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
     OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
   tasks:
-    - set_fact:
+    - name: Set project name based on guid if uuid doesn't exist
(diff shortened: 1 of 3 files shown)


stencell:
What is the reason for this change? Do we need to change because the `guid` is insufficient? This will cause a deviation when people go to look and troubleshoot things. The change is okay, but we should be making something like that with a wholistic plan, not just this.

agonzalezrh:
@marcosmamorim @rut31337 please discuss with @stencell 

I will add this as WIP

marcosmamorim:
@stencell correct, `guid` is insufficient! Username and Project name are primary keys on OpenStack, this PR affects osp-migration only. If you have another suggestion to prevent duplicate user/project, we're here to listen.

stencell:
Is anything happening with this? If not, we should go ahead and close it.

fridim:
Is this PR still wanted ?

GUID is supposed to be unique in cloudforms, right ?
What are the chances to have a guid-collision downstream then ?

stencell:
No responses on this for months, please reopen if the change is still warranted.

PR #1891, opened 2020-06-09
Update project_resources.yml

rut31337:
##### SUMMARY
Set openstack volumes state to error and detached before deleting the volume.  There is an issue where Heat templates created via agnosticD do not maintain dependency with all volumes.  This breaks deletion.  To fix it, the volume status must be set to error or available then detached from any instance.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-osp-resources-destroy/tasks/project_resources.yml

##### ADDITIONAL INFORMATION
```
openstack volume set --state error <vol id>
openstack volume set --detached <vol id>
```

### DIFF
--- a/ansible/roles-infra/infra-osp-resources-destroy/tasks/project_resources.yml
+++ b/ansible/roles-infra/infra-osp-resources-destroy/tasks/project_resources.yml
@@ -47,7 +47,18 @@
         - debug:
             var: _all_volumes
             verbosity: 2
-
+            
+        - name: Reset all volumes state to error
+          command: >-
+            openstack volume set --state error {{ item }}
+          loop: "{{ _all_volumes }}"
+          
+        - name: Detach all volumes
+          command: >-
+            openstack volume set --detached {{ item }}
+          loop: "{{ _all_volumes }}"
+          ignore_errors: yes
+          
         - name: Delete all volumes
           command: >-
             openstack volume delete {{ item }}


ansible/roles-infra/infra-osp-resources-destroy/tasks/project_resources.yml
```diff
@@ -47,7 +47,18 @@
...
+        - name: Reset all volumes state to error
+          command: >-
+            openstack volume set --state error {{ item }}
+          loop: "{{ _all_volumes }}"
+          
+        - name: Detach all volumes
+          command: >-
+            openstack volume set --detached {{ item }}
+          loop: "{{ _all_volumes }}"
+          ignore_errors: yes
```
jkupferer, line 60: Why ignore errors?

rut31337: Because the volume may already be detached.  This is just making sure it is detached.

jkupferer: `ignore_errors` should only be used when an issue should present as an error, but treated as a warning. When looking at the log, tasks with `ignore_errors` look like something went wrong.

The better way to do this is to used a `failed_when` condition so that we control whether it shows as an error at all rather than showing as an ignored error.

rut31337: I read up on it but I dont know how to use failed_when to make it ignore the error.  I assume the command may exit with a non 0 if the volume is already detached.  This is ok.  I dont want it to abort the running playbook.  Is there an example you can point me to?

jkupferer: The simplest is: `failed_when: false`, to say no condition is a failure.
A fancier approach would be to register a variable and check the stdout/stderr:
```
        - name: Detach all volumes
          command: >-
            openstack volume set --detached {{ item }}
          loop: "{{ _all_volumes }}"
          register: r_detatch_volume
          failed_when: >-
            r_detatch_volume.rc != 0 and
            'volume is not attached' in r_detatch_volume.stderr
```
This is just an example. I don't know what the stdout/stderr may really show in this case.

PR #1899, opened 2020-06-10
OSP Cloud Provider fixes for Sandboxes

stencell:
##### SUMMARY
With the recent changes in #1763 it made the `osp-sandbox` harder to use. Specifically with the credentials people were using and forcing into `osp_auth_user`, we now expect the user that deploys the stacks to be `guid-user` to avoid using the admin account. There were probably a number of ways to fix this, including scraping and rebuilding the `osp-sandbox` config all together but this seems like a decent compromise to minimize more changes to the cloud provider with what amounts to a very small use case with sandbox development vs prod deployments. 

With these changes, the OSP cloud provider will now optionally create an additional user called `sandbox-guid-user` when the `osp_create_sandbox` var is set. The expectation and documented use of this user will be for lab/config developers working in a sandbox environment to use it when they are deploying an agnosticd config inside an OSP sandbox. They will need to provide these 3 vars when doing so:
```
guid: sandbox-<real guid>
heat_user_password_provided: <the password from clouds.yaml>
osp_project_create: false
```
This also updates the `sap-hana` config, since it uses a customized heat template. It removes the parts that are no longer required to bring it inline with the standard template we use for the OSP cloud provider.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
osp-sandbox
sap-hana
infra-osp-project-create
infra-osp-resources-destroy

@wilson-walrus will test and document these changes
@fridim Lets discuss this to make sure I'm not missing anything
@makentenza FYI since this was being done after you brought it up and you will be the guinea pig


ansible/roles-infra/infra-osp-resources-destroy/tasks/main.yml
```diff
@@ -18,10 +19,26 @@
...
+      #     OS_AUTH_URL: "{{ osp_auth_url }}"
+      #     OS_USERNAME: "{{ osp_auth_username }}"
+      #     OS_PASSWORD: "{{ osp_auth_password }}"
+      #     OS_PROJECT_NAME: "{{ osp_project_name }}"
+      #     OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
+      #     OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
+      loop:
+      - "{{ guid }}-user"
+      - sandbox-{{ guid }}-user
+
```
fridim, line 39: When passing a variable to a whole set of tasks with `loop` + `include_tasks`, it's better to use a dedicated variable:

```yaml
loop_control:
  loop_var: _keypair_owner
```

stencell:
This has been updated based on some testing and conversations. Now, the sandbox developer would need to define these things:

`guid`: whateveryouwant

`osp_auth_username_member`: The user that you get with the sandbox and is in email/clouds.yaml. It is used to actually create things.

`osp_auth_password_member`: Same as username - get it from your email or clouds.yaml

`osp_project_create`: This is not new...continue setting this to false

`osp_project_name/id`: these are not new. continue setting them

@wilson-walrus If you can, please test on Monday as we discussed so this can be merged to development. We'll have to followup on the solution for the identification and teardown of the sandbox stacks.

stencell:
The latest changes should fix the cleanup as well. Once this is tested by @wilson-walrus we can merge.

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
```diff
@@ -7,21 +7,18 @@
...
-          command: >-
-            openstack stack show -f json base-stack-{{ guid }}
-          register: r_show
-          failed_when: false
-          changed_when: false
-          tags: heat
+  - name: Get a list of stacks in the project
+    command: >-
+      openstack stack list --property tenant={{ osp_project_info[0].id }} -f value --column ID
+    register: r_stacks
```
fridim, line 13: can we add a comment about why `command` is necessary here VS built-in ansible module?

ansible/cloud_providers/osp_infrastructure_deployment.yml
```diff
@@ -40,10 +33,13 @@
...
+    osp_auth_password_member: "{{ heat_user_password }}"
   environment:
     OS_AUTH_URL: "{{ osp_auth_url }}"
-    OS_USERNAME: "{{ osp_auth_username }}"
-    OS_PASSWORD: "{{ osp_auth_password }}"
+    OS_USERNAME: "{{ osp_auth_username_member }}"
+    OS_PASSWORD: "{{ osp_auth_password_member }}"
     OS_PROJECT_NAME: "{{ osp_project_name }}"
     OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
     OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
```
fridim, line 45: Should we make it cleaner, like the previous block, and have those inside each role instead of globally here ?

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
```diff
@@ -1,79 +1,118 @@
...
+  set_fact:
+    heat_user_password_generated: "{{lookup('password', '/dev/null length=12 chars=ascii_letters,digits')}}"
+
+- name: Set heat_user_password
+  set_fact:
+    heat_user_password: "{{ heat_user_password_provided | default(heat_user_password_generated) }}"
+
+- name: Set fact for new user password
+  set_fact:
+    osp_auth_password_member: "{{ heat_user_password_provided | default(heat_user_password_generated) }}"
```
fridim, line 12: equivalent to the simpler:

```yaml
 osp_auth_password_member: "{{ heat_user_password }}"
```

Is that what we want ?  If i want to provide `osp_auth_password_member`, i need to actually set:

* `osp_auth_password_member`
* `heat_user_password`

That sounds strange to me.

Also, now, the user is not created in the Heat template, so there is no reason to keep `heat_user_password`, right ?

We should just replace every occurence of `heat_user_password` with `osp_auth_password_member`. Wdyt ?

i checked, `heat_user_password` is not used in agnosticv.\

In agnosticd:

```
[fridim@lolo agnosticd]$ grep heat_user_password * -r
ansible/roles-infra/infra-osp-template-generate/templates/nested_version/cloud_template_master.j2:      password: {{ heat_user_password }}
ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2:      password: {{ heat_user_password }}
ansible/roles-infra/infra-osp-template-generate/tasks/generate_osp_template.yml:    heat_user_password_generated: "{{lookup('password', '/dev/null length=12 chars=ascii_letters,digits')}}"
ansible/roles-infra/infra-osp-template-generate/tasks/generate_osp_template.yml:    heat_user_password: "{{ heat_user_password_provided | default(heat_user_password_generated) }}"
ansible/roles/host-ocp4-provisioner/templates/clouds.yaml.j2:      password: "{{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/rhel8lab/post_software.yml:          bookbag_os_password: "{{ hostvars.localhost.heat_user_password }}"
ansible/configs/ansible-tower-implementation/files/cloud_providers/osp_cloud_template_master.j2:      password: {{ heat_user_password }}
ansible/configs/sap-rhvh/files/cloud_providers/osp_cloud_template_master.j2:      password: {{ heat_user_password }}
ansible/configs/ocp4-disconnected-osp-lab/software.yml:        #     OS_PASSWORD: "{{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2:      password: "{{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/osp-sandbox/files/clouds.yaml.j2:      password: "{{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/osp-sandbox/files/secret.yaml.j2:osp_auth_password: "{{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/osp-sandbox/post_software.yml:        - "user.info:       password: {{ hostvars['localhost']['heat_user_password'] }}"
ansible/configs/sap-hana/files/cloud_providers/osp_cloud_template_master.j2:      password: {{ heat_user_password }}
ansible/configs/multi-cloud-capsule/files/cloud_providers/osp_cloud_template_master.j2:      password: {{ heat_user_password }}```

File: ansible/roles-infra/infra-osp-project-create/tasks/main.yaml [hunk 5]
```diff
@@ -1,79 +1,118 @@
...
+          subnet: "{{ quota_subnets }}"
+          router: "{{ quota_routers }}"
+          floatingip: "{{ quota_fip }}"
+          security_group: "{{ quota_sg }}"
+          security_group_rule: "{{ quota_sg_rules }}"
+
+      - name: Create user in new project
+        os_user:
+          state: present
+          name: "{{ guid }}-user"
```
fridim, line 69: shouldn't it be `osp_auth_username_member` ?

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
```diff
@@ -1,79 +1,118 @@
...
-        #pool: #when Octavia is available
-        network: "{{ quota_networks }}"
-        subnet: "{{ quota_subnets }}"
-        router: "{{ quota_routers }}"
-        floatingip: "{{ quota_fip }}"
-        security_group: "{{ quota_sg }}"
-        security_group_rule: "{{ quota_sg_rules }}"
+      - name: Add member role to user
+        os_user_role:
+          user: "{{ guid }}-user"
```
fridim, line 76: shouldn't it be `osp_auth_username_member` ?

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
```diff
@@ -1,79 +1,118 @@
...
+          role: _member_
+          project: "{{ osp_project_name }}"
 
-    - set_fact:
-        heat_user_password_generated: "{{lookup('password', '/dev/null length=12 chars=ascii_letters,digits')}}"
+      - name: Add Swift role to user
+        os_user_role:
+          user: "{{ guid }}-user"
```
fridim, line 82: shouldn't it be `osp_auth_username_member` ?

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
```diff
@@ -1,79 +1,118 @@
...
+          msg: No project '{{ osp_project_info }}' found
+
+  - name: Add user if sandbox
+    when: osp_create_sandbox
+    block:
+    - name: Create sandbox user in project
       os_user:
         state: present
-        name: "{{ guid }}-user"
+        name: sandbox-{{ guid }}-user
```
fridim, line 109: Why do we need the `sandbox-{{ guid }}-user` for now ? Wasn't it a trick to be able to use just the "special" guid `sandbox-{{guid}}` at the beginning of the PR, before we introduced the `..._member` user ?

i'm OK to have an additional user, but i can't remember why we need it ?

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
stencell: Comments added.

ansible/cloud_providers/osp_infrastructure_deployment.yml
stencell: We only need these for the one task in this play in order to get the instance info with `os_instance_facts`. We could move it down to the task level, but there is no role to move it into. I'm impartial on this.

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
stencell: I kept `heat_user_password` in here really for compatibility and to minimize change since this is used elsewhere. Ultimately we should no longer need it and `osp_auth_password_member` should be sufficient, but the changes in this PR were already getting bigger than I wanted.

I'd prefer to keep both as is and I will add a comment to `heat_user_password` that it is being deprecated and will be removed later?

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
stencell: yar, updated

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
stencell: updated

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
stencell: updated

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
stencell: No, I think you are correct on this. Removing this and cleaning up the keypair deletion as well.

wilson-walrus:
Tested deployment of sandbox followed by deployment of config onto the resulting sandbox-project. Both deployments completed & destroyed without issue or error.

stencell:
@fridim Please approve and merge this when you get online tomorrow so we can close this out. If we need to make any further tweaks that are not major we can after I think.

ansible/roles-infra/infra-osp-project-create/tasks/main.yaml
fridim: OK, let's track this in an issue.

fridim:
Approved. 

I cannot merge because there are conflicts and i can't resolve them from the github webui.

PR #1487, opened 2020-04-02
Add the capability to gather the openshift installer to ansible host

gmontalvoy:
##### SUMMARY
Fix #511 
After Openshift is installed, the install log can be found at {{ output_dir  }}

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
post-software.yml at both roles, ocp4-workshop and ocp4-cluster

##### ADDITIONAL INFORMATION
After validation, it can be applied to other ocp4 roles

PR #1937, opened 2020-06-15
Get OCP4 installation logs in a block

### LINKED ISSUE #511: in OCP4 configs, automatically grab the openshift_install.log
**Is your feature request related to a problem? Please describe.**
As a developer, i need access to the logs when deployment or deletion fails.

**Describe the solution you'd like**
After openshift-install commands, grab the .openshift_install.log file, gzip it, and copy it to `{{ output_dir }}`.

**Describe alternatives you've considered**
Today, developer needs to SSH to the clientvm to tail or cat the log. => tedious.

**Additional context**
configs:
- ocp4-workshop
- ocp4-shared

gmontalvoy:
#### SUMMARY

Related with PR #1487
Fixes #511

#### ISSUE TYPE
Feature Pull Request

#### COMPONENT NAME

post-software.yml at both roles, ocp4-workshop and ocp4-cluster
#### ADDITIONAL INFORMATION

After validation, it can be applied to other ocp4 roles


gmontalvoy:
> I don't understand the logic now. :-(

The point is to get .openshift_install.log file even if the installation itself fails.

Taking in consideration there will be always an .openshift_install.log from every Openshift4 config, not every 4 config follow the same logic, I created an approach to both configs (ocp4-workshop and ocp4-cluster to see both examples)

If it fits, I will do the same for the rest of configs.

PR #1960, opened 2020-06-17
Revert "Get OCP4 installation logs in a block"

stencell:
Reverts redhat-cop/agnosticd#1937

I'm not sure how this was tested. It is breaking things because the syntax for the block doesn't appear to be correct. I also don't understand why we have the same tasks in both software and post-software phase?

I don't have the bandwidth to test the fix, but it should be as simple as a `-` and some spaces.

```
The error appears to have been in '/tmp/ocp4-workshop-046f/ansible_agnostic_deployer/ansible/configs/ocp4-workshop/software.yml': line 117, column 11, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

        - name: Make a copy of the cluster install config
          ^ here
```

### DIFF
--- a/ansible/configs/ocp4-cluster/post_software.yml
+++ b/ansible/configs/ocp4-cluster/post_software.yml
@@ -8,18 +8,6 @@
   tasks:
   - debug:
       msg: "Post-Software Steps starting"
-  
-  - name: Gzip Install log
-    archive:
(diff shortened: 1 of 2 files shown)


gmontalvoy:
There were 2 different configs affected, ocp4-cluster and ocp4-workshop, syntax check failed on ocp4-cluster only, working on it. I will split PRs in the future in depends of the config.

PR #1970, opened 2020-06-17
Enable Lifecycle Hooks (and implement start hook for ocp4-cluster)

wkulhanek:
##### SUMMARY
Enable lifecycle hooks. These should work for all lifecycle actions (start / stop / status).

There is a default hook in the agnosticd directory that's just a no-op.

ocp4-cluster now implements a post start hook to approve any CSRs for clusters newer than 4.4.7 that have been sleeping through the 24h cert rotation window.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Global
ocp4-cluster

##### ADDITIONAL INFORMATION
Needed to implement lifecycle for ocp4-cluster config - and this PR avoids doing a one off for the config.

@fridim Can you double check that this should work for sandboxed environments? I can't test that from my laptop...

### DIFF
--- a/ansible/lifecycle_entry_point.yml
+++ b/ansible/lifecycle_entry_point.yml
@@ -9,4 +9,12 @@
        })
     }}
 
+# Call livecycle_post_hook_ACTION
+- import_playbook: >-
+    {{ lookup('first_found', {
+         'files': [ 'lifecycle_hook_post_' + ACTION + '.yml', 'lifecycle_hook.yml' ],
+         'paths': ['configs/' + env_type, playbook_dir]
+       })
+    }}
+
 - import_playbook: completion_callback.yml
--- a/ansible/lifecycle_entry_point.yml
+++ b/ansible/lifecycle_entry_point.yml
@@ -11,12 +11,12 @@
        })
     }}
 
-# Call livecycle_post_hook_ACTION
 - name: Call Lifecycle Post Hook
   vars:
     _params:
       files:
-        - "{{ configs }}/{{ env_type }}/lifecycle_hook_post_{{ ACTION }}.yml"
+        - configs/{{ env_type }}/lifecycle_hook_post_{{ ACTION }}.yml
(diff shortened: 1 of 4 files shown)


ansible/lifecycle_entry_point.yml
```diff
@@ -9,4 +9,12 @@
        })
     }}
 
+# Call livecycle_post_hook_ACTION
+- import_playbook: >-
+    {{ lookup('first_found', {
+         'files': [ 'lifecycle_hook_post_' + ACTION + '.yml', 'lifecycle_hook.yml' ],
```
fridim, line 15: Shouldn't the generic hook in the config be named lifecycle_hook_post.yml  ?

`         'files': [ 'lifecycle_hook_post_' + ACTION + '.yml', 'lifecycle_hook_post.yml' ],`

ansible/lifecycle_hook.yml
```diff
@@ -0,0 +1,9 @@
+- name: Lifecycle Hook (fallback)
+  hosts: localhost
+  run_once: true
+  become: false
+  gather_facts: false
+  tasks:
+  - name: Placeholder
+    debug:
+      msg: "No custom lifecycle hook provided. This is the lifecycle hook placeholder"
```
fridim, line 9: instead of creating an empty hook, can't we just use `/dev/null` ?  (i didn't try)

ansible/lifecycle_entry_point.yml
wkulhanek: The idea was that this logic should work for all actions.
If naming livecycle_hook_post we'd have to have 3 defaults.
lifecycle_hook_post, lifecyclc_hook_pre, lifecycle_hook_status.
With just lifecycle_hook we have one default for the post hook regardless of action.

ansible/lifecycle_hook.yml
wkulhanek: I like having the output there. Makes it obvious in the log that no hook was called.

PR #1971, opened 2020-06-17
Add cloud user

wkulhanek:
##### SUMMARY
Add CLOUD_USER variable to both /etc/skel/.bashrc and $HOME/{{ ansible_user }}.bashrc.
Also fix that GUID is not set for {{ ansible_user }}
Remove stern logging tool
Remove files that are nowhere used in the role.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
bastion-lite

##### ADDITIONAL INFORMATION
This enables multi cloud deployments. For example the OCP4 Configuration Lab environment can be deployed to either AWS or OpenStack. Depending on the cloud the path from which to copy the kubeconfig is different (/home/ec2-user or /home/cloud-user). Rather than requiring students to determine which cloud they are on provide an Environment variable. That way instructions can be written like this:

```
. The cluster has been installed by the default user account for the cloud your environment is running on. This could be AWS, OpenStack or any other cloud. An environment variable, *$CLOUDUSER*, has been provided to help witht his task.
+
Copy the installation directory from the default cloud user's home directory to your home directory.
+
[source,sh,subs="{markup-in-source}"]
----------------------------------------------
$ *sudo cp -R /home/${CLOUDUSER}/cluster-$GUID $HOME*
$ *sudo chown -R $(whoami):users $HOME/cluster-$GUID*
----------------------------------------------
```


ansible/roles/bastion-lite/tasks/main.yml
```diff
@@ -64,16 +61,20 @@
...
   lineinfile:
-    path: "/etc/skel/.bashrc"
+    path: "{{ item }}"
     regexp: "^export GUID"
     line: "export GUID={{ guid }}"
+  loop:
+  - "/etc/skel/.bashrc"
+  - "/home/{{ ansible_user }}/.bashrc"
+
+- name: Add CLOUD_USER to /etc/skel/.bashrc and /home/{{ ansible_user }}/.bashrc
```
stencell, line 73: Small nit...update the var in the name of the task to match the var you are using (no _)

wkulhanek: Updated

PR #1993, opened 2020-06-19
Change role ocp-infra-osp-fip to support nsupdate algorithm in nsupdate calls

fdupont-redhat:
##### SUMMARY
Follow-up of https://github.com/redhat-cop/agnosticd/pull/1968 to add the `key_algorithm` to `nsupdate` call in `ocp-infra-osp-fip` role.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp-infra-osp-fip

### DIFF
--- a/ansible/roles/ocp-infra-osp-fip/defaults/main.yml
+++ b/ansible/roles/ocp-infra-osp-fip/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+key_algorithm: hmac-md5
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@stencell agreed. I added the `key_algorithm` to `defaults/main.yml`.

ansible/roles/ocp-infra-osp-fip/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+key_algorithm: hmac-md5
```
stencell, line 2: The var name is wrong. It should be `ddns_key_algorithm` to match what you are using in the task.

fdupont-redhat: Oops. Fixed.

PR #1995, opened 2020-06-19
infra-osp-resources-destroy - Fix Heat stack deletion

fdupont-redhat:
##### SUMMARY
The OpenStack roles now use `osp_auth_username_member` and `osp_auth_password_member` as credentials. The Heat stack deletion tasks file was still using `osp_auth_username` and `osp_auth_password`. This pull request updates the variable names and allows to use the _old_ names for backward compatibility.

The project detection may also fail in the case where it was not created by AgnosticD. Because the user can provide an existing project id, the pull request also allows to fall back to `osp_project_id` variable.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
infra-osp-template-create

### DIFF
--- a/ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
+++ b/ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
@@ -5,8 +5,8 @@
 
 - environment:
     OS_AUTH_URL: "{{ osp_auth_url }}"
-    OS_USERNAME: "{{ osp_auth_username_member | d(osp_auth_username) }}"
-    OS_PASSWORD: "{{ osp_auth_password_member | d(osp_auth_password) }}"
+    OS_USERNAME: "{{ osp_auth_username_member | default(osp_auth_username) }}"
+    OS_PASSWORD: "{{ osp_auth_password_member | default(osp_auth_password) }}"
     OS_PROJECT_NAME: "{{ osp_project_name }}"
     OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
     OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
--- a/ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
+++ b/ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
@@ -1,4 +1,8 @@
 ---
+- name: Get OpenStack project id
+  set_fact:
+    osp_project_id: "{{ osp_project_id | default(osp_project_info[0].id) }}"
+
 - environment:
     OS_AUTH_URL: "{{ osp_auth_url }}"
     OS_USERNAME: "{{ osp_auth_username_member | d(osp_auth_username) }}"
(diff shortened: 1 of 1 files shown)


stencell:
actually I take this back...I didn't read thoroughly. It *might* work. I'm not 100% confident, though. What happens if someone has `osp_auth_username_member` defined and NOT password? Then we default to the admin password and nothing works.

fdupont-redhat:
I thought the goal of `osp_auth_*_member` was to provide a non admin user account to still be able to deploy the configs when not creating the project. So, IMHO it doesn't make sense to declare `osp_auth_username_member` and not `osp_auth_password_member`.

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
```diff
@@ -1,16 +1,16 @@
...
+    OS_PASSWORD: "{{ osp_auth_password_member | d(osp_auth_password) }}"
     OS_PROJECT_NAME: "{{ osp_project_name }}"
     OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
     OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
   block:
   # There is no module that allows retrieving this data.
   - name: Get a list of stacks in the project
     command: >-
-      openstack stack list --property tenant={{ osp_project_info[0].id }} -f value --column ID
+      openstack stack list --property tenant={{ osp_project_info[0].id | d(osp_project_id) }} -f value --column ID
```
stencell, line 13: I think these should be reversed. Use `osp_project_id` if defined...otherwise fall back to `default(osp_project_info[0].id`. We'll only get `osp_project_info` if `osp_project_id` is not defined anyhow.

Also, I think this could be set as a fact in a task before to make it a little easier to read. WDYT?

stencell:
Ehhh...I'm torn because I don't trust things that make sense to always happen. 

Honestly this role exists to cleanup everything - project, resources, all of it. That we are using it to try to cleanup just a subset of these things inside of a project is not wrong...but not what it was intended to solve originally. Really, you are just running the tasks to delete the heat stack. What happens to any other resources that you created _outside_ of that stack while you were deployed? What is cleaning those up? We'll catch them when we delete the sandbox with the purge...but I just don't see a huge amount of sense in this working the same way inside a sandbox. Literally you could just do `openstack stack delete XXX` and get the same functionality this is currently providing inside a sandbox.

fdupont-redhat:
The other resources are destroyed by the config `destroy_env.yml` tasks. A lot of configs rely on the default stack deletion to remove all the instances, networks, floating IPs, etc... created by the stack. So, deletion of the stack should also be done when the project is not created. Using `openstack stack delete` puts you out of the AgnosticD flow.

stencell:
This whole role was built to tear down everything - project, resources, heat stacks, keypairs, users...everything. To do all of those things, we use the admin credentials - both to ensure we have access to delete everything and because those are the only credentials we *know*. Then we added conditionals to maybe do those things...or maybe not.

When a config is deployed in a sandbox and only uses a regular non-admin user, this whole teardown becomes quite useless. Out of the 5 "phases" in this role, only 1 will actually run - deleting the heat stack. Which is the same thing a sandbox user could just do manually with that command. I do not see any reason or benefit for someone running this *inside* a sandbox when the main conditional, `osp_project_create` is set to false and most of the phases are skipped anyhow. 

That said, I do agree with keeping this in the agnosticd workflow even though it is not super practical or useful to "clean up" everything inside a project and will probably give people a false sense of completion...but this sandbox usage is not the primary use case, so that is good enough for now.

@fdupont-redhat I think if you address the last comment and flip those vars around (and change to `default`) this can be merged.

ansible/roles-infra/infra-osp-resources-destroy/tasks/heat.yml
fdupont-redhat: That was my initial plan :) It should be good now.

PR #2009, opened 2020-06-19
ocp4-cluster - Fix FIP DNS records deletion

fdupont-redhat:
##### SUMMARY
When the DNS records for the OCP4 floating IPs are created by AgnosticD, the names are `api.cluster-{{ guid }}` and `*.apps.cluster-{{ guid }}`. In the destruction tasks, the `cluster-` part is missing, so the records are not removed. This pull request fixes that.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-cluster


stencell:
This is a good catch! This might need to be updated in a couple more places as well...let me run a couple of tests.

PR #2018, opened 2020-06-19
OSP - User _member variable for username and password

fdupont-redhat:
##### SUMMARY
Recently, the `osp_auth_username_member` and `osp_auth_password_member` were introduced. The `ocp4-disconnected-osp-lab` and `host-ocp4-provisioner` roles don't use it to expand the `clouds.yaml.j2` template. This pull request adds the support for them.

The `ocp4-disconnected-osp-lab` also didn't support the ability to use an existing project, so the pull request checks if `osp_project_create` is set and uses existing user if yes, like `host-ocp4-provisioner` does.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-disconnected-osp-lab
host-ocp4-provisioner

### DIFF
--- a/ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
+++ b/ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
@@ -2,15 +2,10 @@ clouds:
   {{ osp_project_name }}:
     auth:
       auth_url: "{{ osp_auth_url }}"
-{% if osp_project_create | bool %}
-      username: "{{ guid }}-user"
-      password: "{{ hostvars['localhost']['heat_user_password'] }}"
-{% else %}
       username: "{{ osp_auth_username_member }}"
       password: "{{ osp_auth_password_member }}"
-{% endif %}
       project_name: "{{ osp_project_name }}"
-      project_id: "{{ hostvars['localhost']['osp_project_info'][0].id | d(osp_project_id) }}"
+      project_id: "{{ hostvars['localhost']['osp_project_info'][0].id | default(osp_project_id) }}"
       user_domain_name: "Default"
     region_name: "regionOne"
     interface: "public"
--- a/ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
+++ b/ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
@@ -2,11 +2,16 @@ clouds:
   {{ osp_project_name }}:
     auth:
       auth_url: "{{ osp_auth_url }}"
+{% if osp_project_create | bool %}
       username: "{{ guid }}-user"
-      project_name: "{{ osp_project_name }}"
-      project_id: "{{ hostvars['localhost']['osp_project_info'][0].id }}"
-      user_domain_name: "Default"
       password: "{{ hostvars['localhost']['heat_user_password'] }}"
+{% else %}
+      username: "{{ osp_auth_username_member }}"
+      password: "{{ osp_auth_password_member }}"
+{% endif %}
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@tonykay updated the PR to use `d()`.

ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
```diff
@@ -2,11 +2,16 @@ clouds:
   {{ osp_project_name }}:
     auth:
       auth_url: "{{ osp_auth_url }}"
+{% if osp_project_create | bool %}
```
stencell, line 5: I don't think I like this. We should just use `osp_auth_username_member` going forward. Even if we are creating a project, we set that var at the beginning of the role so it is always available. There is no reason here to define the user as `guid-user` again.

Same for the password...`osp_auth_password_member` should be fine whether the project is being created or not.

ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
```diff
@@ -2,11 +2,16 @@ clouds:
...
       username: "{{ guid }}-user"
+      password: "{{ hostvars['localhost']['heat_user_password'] }}"
+{% else %}
+      username: "{{ osp_auth_username_member }}"
+      password: "{{ osp_auth_password_member }}"
+{% endif %}
       project_name: "{{ osp_project_name }}"
-      project_id: "{{ hostvars['localhost']['osp_project_info'][0].id }}"
+      project_id: "{{ hostvars['localhost']['osp_project_info'][0].id | d(osp_project_id) }}"
```
stencell, line 13: please change this to `default` instead of `d`

ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
fdupont-redhat: I wanted to stay backward compatible, but I'm fine with that :)

ansible/configs/ocp4-disconnected-osp-lab/files/clouds.yaml.j2
fdupont-redhat: Done

ansible/roles/host-ocp4-provisioner/templates/clouds.yaml.j2
```diff
@@ -6,8 +6,8 @@ clouds:
       username: "{{ guid }}-user"
       password: "{{ hostvars['localhost']['heat_user_password'] }}"
 {% else %}
-      username: "{{ osp_auth_username }}"
-      password: "{{ osp_auth_password }}"
+      username: "{{ osp_auth_username_member | d(osp_auth_username) }}"
```
stencell, line 9: I just don't think we need this. `osp_auth_username|password_member` is always going to be set now to either a user defined value...or [what is defined here](https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-osp-project-create/tasks/main.yaml#L12-L18). We can get rid of this whole conditional and just replace the currently hardcoded `username|password` with the `osp_auth_username|password_member` vars.

I don't even think we should default back to `osp_auth_username|password` any longer. Those should _only_ be used for initially creating the project, user, etc. and then not again until we delete everything.

fdupont-redhat: I agree. My understanding of this role is that it should always be called once the cloud resources have been deployed, hence a user with `_member_` role exists.

PR #2019, opened 2020-06-19
Ocp4 workload kubevirt allow time for subscription status

fdupont-redhat:
##### SUMMARY
It may happen that the Subscription CR is not picked by the controller when we check its status, so the `status` property doesn't exist yet. This pull request allows an empty `status` property.

It also adds a small pause to be sure that the operator group is created.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_kubevirt

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_kubevirt/tasks/workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_kubevirt/tasks/workload.yml
@@ -13,6 +13,10 @@
     state: present
     definition: "{{ lookup('file', 'operatorgroup.yaml') }}"
 
+- name: Waiting 5 seconds to ensure the operator group is created
+  pause:
+    seconds: 5
+
 - name: Create subscription for HCO
   k8s:
     state: present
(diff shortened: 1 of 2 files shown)


ansible/roles_ocp_workloads/ocp4_workload_kubevirt/tasks/workload.yml
```diff
@@ -13,6 +13,10 @@
     state: present
     definition: "{{ lookup('file', 'operatorgroup.yaml') }}"
 
+- name: Waiting 5 seconds to ensure the operator group is created
```
stencell, line 16: We should not wait an arbitrary amount of time like this. If you want to see if the operator group (or anything) is created, add a task that retrieves the `operatorgroup` and runs retries until it is there.

mali-chainzee:
> @mali-chainzee is identified as the author of this role, so I'd like to have his approval. If he is no longer the author/maintainer, the `meta/main.yml` will need to be updated.

hi, I approve this PR

ansible/roles_ocp_workloads/ocp4_workload_kubevirt/tasks/workload.yml
fdupont-redhat: I agree with you. I've changed it.

fdupont-redhat:
No offense @mali-chainzee, but I'm taking over ownership of the role, as well as maintenance.

PR #2021, opened 2020-06-19
WIP: Add additional recovery options to ocp4-cluster (may not be needed with 4.4.8, test ongoing)

wkulhanek:
##### SUMMARY
Add additional recovery logic for OCP 4.4.7 + to ocp4-cluster.

In 4.4.7+ sometimes the OpenShift API Server Pods don't restart as they should. Killing them fixes the problem.

Once this is tested this will also need to be added to ocp4-workshop.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-cluster

### DIFF
--- a/ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
+++ b/ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
@@ -87,6 +87,7 @@
         kind: ClusterVersion
         name: version
       register: r_cluster_version
+
     - name: Set ocp4_cluster_version fact
       set_fact:
         ocp4_cluster_version: "{{ r_cluster_version.resources[0].status.history[0].version }}"
--- a/ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
+++ b/ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
@@ -79,3 +79,43 @@
       when: r_new_csrs.resources | length > 0
       command: "oc adm certificate approve {{ item.metadata.name }}"
       loop: "{{ r_new_csrs.resources }}"
+
+    # Get OpenShift Version to determine if we need additional recovery actions
+    - name: Get OpenShift ClusterVersion
+      k8s_facts:
+        api_version: config.openshift.io/v1
+        kind: ClusterVersion
+        name: version
+      register: r_cluster_version
+    - name: Set ocp4_cluster_version fact
+      set_fact:
+        ocp4_cluster_version: "{{ r_cluster_version.resources[0].status.history[0].version }}"
+
+    # Only do the following recovery actions on 4.4.7 and higher 4.4 OpenShift versions
+    # Before 4.4.7 the cluster would still be broken. Starting in 4.5 this won't be necessary
+    # Also only do when previously there were certificates to be approved
+    - name: Workaround for OCP 4.4 bug (API Servers need restarting)
+      when:
+      - ocp4_cluster_version is version_compare('4.4.7', '>=')
+      - ocp4_cluster_version is version_compare('4.5.0', '<')
+      - r_csrs.resources | length > 0 or r_new_csrs.resources | length > 0
+      block:
+      - name: Wait 5 minutes for things to settle
+        pause:
+          seconds: 300
+
+      - name: Find openshift-apiserver pods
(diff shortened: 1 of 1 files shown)


ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
```diff
@@ -79,3 +79,43 @@
...
       loop: "{{ r_new_csrs.resources }}"
+
+    # Get OpenShift Version to determine if we need additional recovery actions
+    - name: Get OpenShift ClusterVersion
+      k8s_facts:
+        api_version: config.openshift.io/v1
+        kind: ClusterVersion
+        name: version
+      register: r_cluster_version
+    - name: Set ocp4_cluster_version fact
```
stencell, line 90: please add line above to maintain formatting throughout?

ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
```diff
@@ -79,3 +79,43 @@
...
+    # Only do the following recovery actions on 4.4.7 and higher 4.4 OpenShift versions
+    # Before 4.4.7 the cluster would still be broken. Starting in 4.5 this won't be necessary
+    # Also only do when previously there were certificates to be approved
+    - name: Workaround for OCP 4.4 bug (API Servers need restarting)
+      when:
+      - ocp4_cluster_version is version_compare('4.4.7', '>=')
+      - ocp4_cluster_version is version_compare('4.5.0', '<')
+      - r_csrs.resources | length > 0 or r_new_csrs.resources | length > 0
+      block:
+      - name: Wait 5 minutes for things to settle
```
stencell, line 103: Is there something else we can look at here? I don't like picking a number like this and hoping it is settled...what if it takes longer for some reason (we've run into this often enough before)? Is there something that can be identified as "settled"?

wkulhanek: Not really. Unless you want to start parsing COs status which is super messy.

ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
wkulhanek: If you insist. I like to keep things together that belong together.

wkulhanek:
Added newline.

This is temp fix until 4.5 comes out. It only runs when CSRs got approved. This will be good enough.

ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
stencell: I could have sworn we discussed this in the code quality contributors guide, but I don't see it now hmmmmm. Still, but that logic, it is hard to draw the line between what belongs together and what doesn't...kind of subjective and this makes it more readable IMO.

stencell:
4.4.8 introduced new changes to automatically recover control plane certificates. So that will be tested as discussed.

ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml
jkupferer: Something like?
```
  - name: Wait for cluster operators to settle
    k8s_info:
      api_version: config.openshift.io/v1
      kind: ClusterOperator
    register: r_get_co
    failed_when: >-
      r_get_co | json_query(_condition_unavailable_query) | length > 0 or
      r_get_co | json_query(_condition_progressing_query) | length > 0 or
      r_get_co | json_query(_condition_degraded_query) | length > 0
    vars:
      _condition_unavailable_query: >-
        resources[].status.conditions[?type=='Available' && status=='False'][]
      _condition_progressing_query: >-
        resources[].status.conditions[?type=='Progressing' && status=='True'][]
      _condition_degraded_query: >-
        resources[].status.conditions[?type=='Degraded' && status=='True'][]
    until: r_get_co is success
    retries: 100
    delay: 30
```

wkulhanek: I think we may not need this at all anymore. Will know Monday.

wkulhanek:
Not needed with 4.4.8+

PR #2053, opened 2020-06-23
New configuration for SAP SMART workshop

makentenza:
##### SUMMARY
This new AgnosticD config will support the workshops on SAP SMART, to showcase the Smart Management capabilities with SAP Digital Core included as part of the RHEL for SAP Solutions subscription"

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
sap-smart


ansible/configs/sap-smart/files/cloud_providers/osp_cloud_template_master.j2
```diff
@@ -0,0 +1,232 @@
...
+
+resources:
+
+  {{ guid }}-infra_key:
+    type: OS::Nova::KeyPair
+    properties:
+      name: {{ guid }}-infra_key
+      save_private_key: true
+
+{% if osp_project_create | bool %}
```
stencell, line 18: This should be removed as well as the `project_role_user` section. These are handled in an earlier playbook now.

ansible/configs/sap-smart/post_software.yml
```diff
@@ -0,0 +1,198 @@
+- name: Step 00xxxxx post software
```
stencell, line 1: What is this section for targeting `support` hosts? If it is not needed and not doing anything, we shoudl start with the first actual play. The debug tasks can still be included in that play if you want.

ansible/configs/sap-smart/post_software.yml
```diff
@@ -0,0 +1,198 @@
...
+  serial: 1
+  become: True
+  gather_facts: True
+  tasks:
+
+    - include_role:
+        name: redhat_sap.sap_rhsm
+
+    - name: Enable required repository for Katello Agent
+      command: subscription-manager repos --enable="satellite-tools-6.7-for-rhel-8-x86_64-e4s-rpms"
```
stencell, line 89: Can we use `rhsm_repository` module here?

ansible/configs/sap-smart/pre_software.yml
```diff
@@ -0,0 +1,86 @@
+- name: Configure RHEL 8 hosts with Repositories 
+  hosts:
+    - bastions:towers
```
stencell, line 3: For consistency in formatting across the config, can we keep the hosts list on one line for all of these plays?

ansible/configs/sap-smart/default_vars.yml
```diff
@@ -0,0 +1,531 @@
...
+
+# This is where the ssh_config file will be created, this file is used to
+# define the communication method to all the hosts in the deployment
+deploy_local_ssh_config_location: "{{output_dir}}/"
+
+install_bastion: true
+install_common: true
+
+# Variable used to identify the request being made by a Red Hatter or Partner (external)
+rh_internal: true
```
stencell, line 38: Since it seems based on the task that you are using this var for that you are concerned about the SAP software being made available to externals, would it not be safer to default this to `false` and then explicitly allow it in the agnosticv config for the internal catalog item - assuming you are planning on different ones?

ansible/configs/sap-smart/default_vars.yml
```diff
@@ -0,0 +1,531 @@
...
+        group: BastionSG
+        rule_type: Ingress
+      - name: HanaTCPPPortsPrivate
+        description: "Only from bastion"
+        from_port: 1
+        to_port: 65535
+        protocol: tcp
+        group: BastionSG
+        rule_type: Ingress
+  - name: BastionSG
```
stencell, line 120: Hmm, aren't these rules going to restrict incoming connections to the bastion even except from instances in the HanaSG group? Meaning you can't SSH to the bastion from the outside.

ansible/configs/sap-smart/sat_workload.yml
```diff
@@ -0,0 +1,126 @@
...
+    line: 127.0.0.1 access.redhat.com
+    create: yes
+
+- name: "Copy subscription manifest .zip file"
+  copy:
+    src: "{{ manifest_file_path }}"
+    dest: /root/manifest.zip
+
+- name: "Add manifest to Satellite"
+  command: >
```
stencell, line 55: Can you look at your use of command module like this? You are using folding block operator, but in some you are leaving the `\` between lines and others you aren't. If you aren't using literal block operator, I don't think you need the `\`.

ansible/configs/sap-smart/sat_workload.yml
```diff
@@ -0,0 +1,126 @@
+- name: 'Install required packages'
+  yum:
+    name: '{{ item }}'
+    state: installed
+  loop:
+    - satellite
+    - firewalld
+    - python-firewall
+
+- name: 'Ensure firewalld is running'
```
stencell, line 10: Just out of curiosity, why are you enabling and configuring firewalld here and then disabling it at the end of this task list?

ansible/configs/sap-smart/files/cloud_providers/osp_cloud_template_master.j2
makentenza: Done

ansible/configs/sap-smart/post_software.yml
makentenza: This is coming from the original sample config I used in the past to create mines. Just removed this.

ansible/configs/sap-smart/post_software.yml
makentenza: Sure, moved!

ansible/configs/sap-smart/pre_software.yml
makentenza: Yes, I did not notice that. Changed all that apply

ansible/configs/sap-smart/default_vars.yml
makentenza: Yeah, agree

ansible/configs/sap-smart/default_vars.yml
makentenza: No, external traffic to the bastion is open

ansible/configs/sap-smart/sat_workload.yml
makentenza: Yes, just removed that, not sure why I added the operator in the first place when is not required

ansible/configs/sap-smart/sat_workload.yml
makentenza: There is an easy explanation, I was doing some testing and I forgot to remove it ;-)

makentenza:
Hey @stencell thanks for reviewing this! Just applied the suggested changes after your review.

ansible/configs/sap-smart/sat_workload.yml
stencell: haha, ok. If you don't need it can you also not install it in the task before as well? As well as get rid of the task at the end that stops it.

ansible/configs/sap-smart/post_software.yml
```diff
@@ -0,0 +1,195 @@
...
+- name: Configure Ansible Tower
+  hosts: towers
+  gather_facts: False
+  become: yes
+  tasks:
+
+    - block:
+
+      - name: Ensure dependencies for tower-cli are installed
+        dnf:
```
stencell, line 25: I've noticed in a few spots you've used different modules. Can we standardize on any tasks that are installing packages? `package` module preferred...otherwise `yum` is okay i guess.

ansible/configs/sap-smart/sat_workload.yml
```diff
@@ -0,0 +1,122 @@
...
+  command: >
+     hammer
+       -u "{{ satellite_username }}"
+       -p "{{ satellite_password }}"
+       activation-key add-subscription
+       --name "{{ satellite_activation_key }}"
+       --subscription-id "1"
+       --organization "{{ satellite_organization }}"
+
+- name: Ensure firewalld is not running
```
stencell, line 118: This needs to be removed since you are not going to install it or run it.

ansible/configs/sap-smart/post_software.yml
makentenza: Done

ansible/configs/sap-smart/sat_workload.yml
makentenza: Done

makentenza:
@stencell just changed the commented bits

ansible/configs/sap-smart/default_vars.yml
stencell: that doesn't make sense, but I think I see the problem. You are using an out of date version of the OSP cloud provider template. So you are getting what you have defined here PLUS what is defined in the defaults. If that is the case, I was not expecting that to be munged together when the same SG name is used. Hmmm. I guess if it works, you can keep it, but I'd recommend specifying the SSH access in your config for that SG to avoid unintended removal of it in the future.

https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-osp-template-generate/templates/cloud_template_master.j2#L58-L61

stencell:
@makentenza Just one remaining. I'll leave it to you on whether you want to make sure the SG rules are specified in your config (vs relying on external defaults) as well as updating your cloud provider template. 

Otherwise, the only other request I have is that you add some author information to the README. That way we can make sure you (or whoever is specified) can approve any changes to this config going forward.

makentenza:
Hey @stencell as discussed I'll keep SG and template as-is (template is required since I need to mount existing images as secondary volumes which are not supported by the original template) and will create another PR for adding author information to all my configs

PR #2066, opened 2020-06-25
Add variable documentation for __meta__ and agnosticv_meta

jkupferer:
##### SUMMARY

Add variable documentation for __meta__ and agnosticv_meta and fix underscore evaluation in asciidoc.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

Variable documentation.

### DIFF
--- a/docs/Variables.adoc
+++ b/docs/Variables.adoc
@@ -33,19 +33,24 @@ Required variable, but may be set to `none` where not applicable.
 | Name
 | Description
 
-| `pass:[__meta__]`
-| Metadata used by the Babylon infrastructure when running AgnosticD.
-This variable should be ignored by workloads, configs, and roles.
-
-| `agonsticd_callback_token`
+| `agnosticd_callback_token`
 | Authentication Bearer token used to callback with information and data collected from `agnosticd_user_info`.
 This is used to integrate AgnosticD with external systems including Babylon and GUID Grabber.
 Default value `pass:[__meta__.callback.token]`.
 
-| `agonsticd_callback_url`
+| `agnosticd_callback_url`
 | API URL used to callback with information and data collected from `agnosticd_user_info`.
 This is used to integrate AgnosticD with external systems including Babylon and GUID Grabber.
 Default value `pass:[__meta__.callback.url]`.
+|============================
+
+=== Other Variables
+
+[options="header",cols="1,4"]
+|============================
+| `pass:[__meta__]`
+| Metadata used by the Babylon infrastructure when running AgnosticD.
+This variable should be ignored by workloads, configs, and roles.
 
 | `agnosticv_meta`
 | Metadata used by the OpenTLC and RHPDS admin host when running AgnosticD.
--- a/docs/Variables.adoc
+++ b/docs/Variables.adoc
@@ -48,6 +48,9 @@ Default value `pass:[__meta__.callback.url]`.
 
 [options="header",cols="1,4"]
 |============================
+| Name
+| Description
+
 | `pass:[__meta__]`
 | Metadata used by the Babylon infrastructure when running AgnosticD.
 This variable should be ignored by workloads, configs, and roles.
--- a/docs/Variables.adoc
+++ b/docs/Variables.adoc
@@ -30,13 +30,21 @@ Required variable, but may be set to `none` where not applicable.
 | Name
 | Description
 
+| `pass:[__meta__]`
+| Metadata used by the Babylon infrastructure when running AgnosticD.
+This variable should be ignored by workloads, configs, and roles.
+
 | `agonsticd_callback_token`
 | Authentication Bearer token used to callback with information and data collected from `agnosticd_user_info`.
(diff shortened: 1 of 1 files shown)


docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
...
+| Metadata used by the Babylon infrastructure when running AgnosticD.
+This variable should be ignored by workloads, configs, and roles.
+
 | `agonsticd_callback_token`
 | Authentication Bearer token used to callback with information and data collected from `agnosticd_user_info`.
 This is used to integrate AgnosticD with external systems including Babylon and GUID Grabber.
-Default value `__meta__.callback.token`.
+Default value `pass:[__meta__.callback.token]`.
 
 | `agonsticd_callback_url`
```
fridim, line 45: same here

docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
 | Name
 | Description
 
+| `pass:[__meta__]`
+| Metadata used by the Babylon infrastructure when running AgnosticD.
+This variable should be ignored by workloads, configs, and roles.
+
 | `agonsticd_callback_token`
```
fridim, line 40: there is a typo here  `agonsticd` -> `agnosticd`

docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
 | Name
 | Description
 
+| `pass:[__meta__]`
+| Metadata used by the Babylon infrastructure when running AgnosticD.
```
fridim, line 37: I'd be more specific here.

Let's explain exactly what it is. The relation is not obvious. agnosticd can be run from laptop and in that case, we shouldn't care about this special variable.
I don't think this one should be considered as a core variable. It's a Babylon variable, and it is used for callbacks.

ex: it is a variable passed to agnosticd by the Babylon infrastructure. It contains information about the context etc etc.
Maybe give an example of a place it is used?

docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
...
+This variable should be ignored by workloads, configs, and roles.
```
fridim, line 38: Good :+1: 
So one could ask, what is it used for ?

docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
...
+
 | `agonsticd_callback_token`
 | Authentication Bearer token used to callback with information and data collected from `agnosticd_user_info`.
 This is used to integrate AgnosticD with external systems including Babylon and GUID Grabber.
-Default value `__meta__.callback.token`.
+Default value `pass:[__meta__.callback.token]`.
```
fridim, line 43: What happen when `__meta__` (and thus `__meta__.callback.token`) is not defined?
=> Callback is ignored all together
I think we should make it clear

docs/Variables.adoc
```diff
@@ -30,13 +33,21 @@ Required variable, but may be set to `none` where not applicable.
...
 
 | `agonsticd_callback_url`
 | API URL used to callback with information and data collected from `agnosticd_user_info`.
 This is used to integrate AgnosticD with external systems including Babylon and GUID Grabber.
-Default value `__meta__.callback.url`.
+Default value `pass:[__meta__.callback.url]`.
+
+| `agnosticv_meta`
+| Metadata used by the OpenTLC and RHPDS admin host when running AgnosticD.
+This variable should be ignored by workloads, configs, and roles.
```
fridim, line 52: is this a core variable ? if it should be ignore, where is it used ?
What does it do ?

docs/Variables.adoc
jkupferer: I wish we weren't using `__meta__` at all in agnosticd. I don't think it is appropriate for anything in agnosticd to reference any metadata settings. It violates the whole concept of it being metadata. Because this is used for babylon really all I want to say here is don't use it in agnosticd.

docs/Variables.adoc
jkupferer: One could ask, but the answer should be go look at the Babylon docs, it isn't used by agnosticd.

docs/Variables.adoc
jkupferer: I want to remove the use of `__meta__` in this case and have the babylon_anarchy_governor just set `agnosticd_callback_url` and `agnosticd_callback_token`.

docs/Variables.adoc
jkupferer: Same here. A developer may notice it is set and be tempted to use it, but all that I want to say here is don't use it, don't worry about it.

docs/Variables.adoc
fridim: Ok, i agree. It wasn't clear enough since it's listed as "core variables"

PR #2089, opened 2020-06-29
Change Contributing guidelines

fridim:
##### SUMMARY
In contributing.adoc, point to redhat-cop guidelines.

- Mention redhat-cop guidelines
- Are issues mandatory for PR ?
- Specify where to document config variables

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/docs/Contributing.adoc
+++ b/docs/Contributing.adoc
@@ -43,10 +43,14 @@ These are our contribution guidelines for helping out with the project. Any sugg
 ** Use `k8s` and `k8s_info` modules as much as possible and not `oc` command.
 . All tasks AND plays must have names.
 . Roles must have documented variables. This can be documented in either:
-** Role defaults/main.yml (preferred)
+** Role `defaults/main.yml` (preferred)
 ** Role README file
 ** Choose one or the other. If you are commenting your variables in the defaults, your README should just point to this file. Do not split your variable documentation across these.
 . Roles must have `meta/main.yml` with author/contact information for support.
+. Configs must have documented variables. This can be documented in either:
+** Config `default_vars*.yml` (preferred)
+** Config README file
+** Choose one or the other. If you are commenting your variables in the `default_vars.yml`, your README should just point to this file. Do not split your variable documentation across these.
 . Be extra careful with external dependencies. Identify them and make sure the versions are **pinned** (use versions, tags, commit ID, etc.).
 ** External Git repos
 ** Libraries/Modules
--- a/docs/Contributing.adoc
+++ b/docs/Contributing.adoc
@@ -9,13 +9,13 @@ These are our contribution guidelines for helping out with the project. Any sugg
 * The Ansible link:https://docs.ansible.com/ansible/latest/community/code_of_conduct.html[Code of Conduct] still applies.
 * For git messages, branch names, etc., follow link:https://github.com/redhat-cop/agnosticd/blob/development/docs/git-style-guide.adoc[Git Style Guide].
 * Pull Requests should reference an *issue* whenever possible.
-** Completely new functionality e.g. a new role or new config etc, must reference an issue. 
+** Completely new functionality e.g. a new role or new config etc, must reference an issue.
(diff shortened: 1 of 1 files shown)


docs/Contributing.adoc
```diff
@@ -9,13 +9,13 @@ These are our contribution guidelines for helping out with the project. Any sugg
 * The Ansible link:https://docs.ansible.com/ansible/latest/community/code_of_conduct.html[Code of Conduct] still applies.
 * For git messages, branch names, etc., follow link:https://github.com/redhat-cop/agnosticd/blob/development/docs/git-style-guide.adoc[Git Style Guide].
 * Pull Requests should reference an *issue* whenever possible.
-** Completely new functionality e.g. a new role or new config etc, must reference an issue. 
+** Completely new functionality e.g. a new role or new config etc, must reference an issue.
```
wkulhanek, line 12: Hm. Do we really want to mandate that?

stencell: I agree...After thinking about this, I am not for it. I think it should be the opposite. If you are touching, fixing, changing, an *existing* config or role, you should need an issue. Completely new roles or configs should not need an issue, but a well crafted description. Almost no one does this and I'm starting to ask more people to get in the habit.

Now, with changes to existing configs/roles that require an issue...that can get tedious, but we can't break the rules just because someone says "oh this is just one thing no big deal". Issues for changes are either enforced or they are not. Is having a well crafted description in the PR sufficient to negate the need for an issue? Do we make people copy and paste that description to an issue just to reference it? That is busy work. No thanks.

I think there will be some nuance here.

Add - completely new role or config. No issue needed. PR must have well crafted description with enough detail for someone to understand exactly what the role or config is doing. It must also follow other rules.

Change - changing an existing role as an elective or non-essential improvement. Does this need an issue?

Fix - How is this different than change? Is it break/fix? If anything needs an issue open, it would be Fix, but this seems very subjective to whoever is submitting.

Remove - Again, a well crafted description should explain why a config/role is being removed. Of course, I have yet to see this unicorn in the wild, so probably the lease important of the three :)

I reckon this comes down to - what is forcing the opening of issues really doing for any of these? I'm not against using them when people want to report an issue, but considering the speed at which things are reported and must be fixed I think it is going to become an inconvenient step *at this time* to force it.

fridim: I agree we usually don't provide an issue and a PR with a good explanation and description can be self-sufficient.
i pushed a change. Let me know what you think now.

docs/Contributing.adoc
```diff
@@ -9,13 +9,13 @@ These are our contribution guidelines for helping out with the project. Any sugg
 * The Ansible link:https://docs.ansible.com/ansible/latest/community/code_of_conduct.html[Code of Conduct] still applies.
 * For git messages, branch names, etc., follow link:https://github.com/redhat-cop/agnosticd/blob/development/docs/git-style-guide.adoc[Git Style Guide].
 * Pull Requests should reference an *issue* whenever possible.
```
stencell, line 11: I'm still not feeling right about this. It is kind a vague and unless it is more of a detailed requirement, no one is going to do it. 

We agree *all* PRs should have well-crafted description. Do we want to require an issue for a certain type of PR like a fix? Does the issue even matter if the reason behind the PR and what it fixes is in the description?

docs/Contributing.adoc
```diff
@@ -28,6 +28,7 @@ These are our contribution guidelines for helping out with the project. Any sugg
 * Owner of existing role or config must be set as reviewer.
 ** If new role or config, owner must be identified.
 * If your Pull Request is not ready for review, open it as link:https://github.blog/2019-02-14-introducing-draft-pull-requests/[Draft] or prefix with `WIP` in the title.
+* AgnosticD being part of the Red Hat Community of Practices, the link:https://redhat-cop.github.io/contrib/[Red Hat CoP Contibution Guidelines] applies.
```
stencell, line 31: Change to 

`AgnosticD is part of the Red Hat Community of Practices; the link:https://redhat-cop.github.io/contrib/[Red Hat CoP Contribution Guidelines] apply.

PR #2145, opened 2020-07-08
Update Openshift Pipelines setup

mbogoevici:
- Use OpenShift Pipelines v1.0.1 (Red Hat)
- Update pipeline definition to account for Tekton 'kn' task changes


ansible/roles/ocp4-workload-ml-workflows-infra-summit2020/templates/pipelines/subscription.yml.j2
```diff
@@ -4,8 +4,9 @@ metadata:
   name: openshift-pipelines-operator
   namespace: openshift-operators
 spec:
-  channel: dev-preview
+  channel: ocp-4.4
   installPlanApproval: Automatic
```
wkulhanek, line 8: If you set Automatic you don't need the startingCSV....

PR #2164, opened 2020-07-12
[DONT MERGE] change role infra-ibm-rg-create infra-ibm-ssh-key infra-ibm-workspace-create to ensure idempotency

gmontalvoy:
##### SUMMARY
ibm cloudprovider idempotency tested to the workspace creation.
No development branch is involved, to be reviewed by @stencell 

##### ISSUE TYPE
- Feature Pull Request


ansible/roles-infra/infra-ibm-rg-create/tasks/main.yml
```diff
@@ -1,45 +1,46 @@
...
     var: resource_group_exists
 
-- name: Get the ID of resource group if it exists
-  when: resource_group_exists
-  set_fact:
-    resource_group_id: "{{ (r_rg_list.stdout | from_json) | json_query(rg_id_query) }}"
-  vars:
-    rg_id_query: >
-      [?name=='{{ guid }}-rg'].id|[0]
+# I don't see why do we need this
```
stencell, line 27: ID is more precise than name and I'm not sure if we will use it again later. IBM is not consistent with their tools and whether they want a name or ID. It doesn't hurt to have, so I am going to leave it.

ansible/roles-infra/infra-ibm-ssh-key/tasks/main.yml
```diff
@@ -13,11 +13,12 @@
...
 #   register: r_ssh_key_fingerprint
 
+# If we only need one key on the sandbox, we can use the task below to do both, check and create
+# Idempotency tested
 - name: Create SSH key if it does not exist
   when: (r_ssh_keys.stdout | from_json) | json_query(ssh_key_query) | length == 0
   command: >-
-    ibmcloud is keyc {{ guid }}-key "{{ infra_ibm_ssh_key_pubssh_key }}"
-    --json --resource-group-id {{ resource_group_id }}
+    ibmcloud is key-create {{ guid }}-key "{{ infra_ibm_ssh_key_pubssh_key }} --resource-group-name {{ guid }}-rg"
```
stencell, line 21: I'm going to keep this with ID for now since we have it from the previous RG create and/or retrieve.

ansible/roles-infra/infra-ibm-ssh-key/tasks/main.yml
```diff
@@ -13,11 +13,12 @@
 #     | cut -d " " -f2
 #   register: r_ssh_key_fingerprint
 
+# If we only need one key on the sandbox, we can use the task below to do both, check and create
+# Idempotency tested
```
stencell, line 17: This will not work. It actually checks the fingerprint of the public key you are uploading, so if you have a key - no matter what it is named or what RG it is in - and try to upload another, it fails. This is why i have the previous task roughed out.

ansible/roles-infra/infra-ibm-upload-plan/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
```
stencell, line 1: I think this does not need to be another role. I'll merge this in with the template create or template generate.

ansible/roles-infra/infra-ibm-rg-create/tasks/main.yml
gmontalvoy: Agreed, commit will follow.

ansible/roles-infra/infra-ibm-ssh-key/tasks/main.yml
gmontalvoy: Agreed, commit will follow.

ansible/roles-infra/infra-ibm-ssh-key/tasks/main.yml
gmontalvoy: It will fail if you use the same public_key content to create a new one but should work if you use a new public_key that does not hit with the previously created one.

ansible/roles-infra/infra-ibm-upload-plan/tasks/main.yml
gmontalvoy: Can we revisit this?
I'm not sure how to handle to call the tf definition upload, plan generation and the sandbox creation itself within the template creation and generate.

Would be possible to rename the role and then create another task file into the role to be imported by the main.yml file?

stencell: Yes, let us discuss more. I think we will merge a bunch of these together. There are too many shared registered vars where it doesn't make sense to have them as separate roles. If we register a var in role 1 and then need that var in role 2, it either needs to be a default (and we don't know what it will be) or they should be 1 role.

PR #2186, opened 2020-07-17
Adds admin when gitea operator is deployed

thoraxe:
The operator role didn't create an admin user for gitea, so you couldn't
really do anything. This adds a var for the user and password and makes
sure that the admin is created idempotently (although it won't fix or
change a password if run a second time).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4_workload_gitea_operator

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_gitea_operator/defaults/main.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_gitea_operator/defaults/main.yml
@@ -36,3 +36,6 @@ ocp4_workload_gitea_operator_ssl_route: true
 
 # Route name for the Gitea Route. Only set if you know what you are doing!
 # ocp4_workload_gitea_operator_gitea_route: gitea.apps.clusterdomain
+
+ocp4_workload_gitea_operator_admin_user: gitea
+ocp4_workload_gitea_operator_admin_password: gitea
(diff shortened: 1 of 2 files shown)


ansible/roles_ocp_workloads/ocp4_workload_gitea_operator/defaults/main.yml
```diff
@@ -36,3 +36,6 @@ ocp4_workload_gitea_operator_ssl_route: true
 
 # Route name for the Gitea Route. Only set if you know what you are doing!
 # ocp4_workload_gitea_operator_gitea_route: gitea.apps.clusterdomain
+
+ocp4_workload_gitea_operator_admin_user: gitea
```
wkulhanek, line 40: Should we have a variable to enable/disable creation of admin users? E.g. ocp4_workload_gitea_operator_install_admin_user: true

ansible/roles_ocp_workloads/ocp4_workload_gitea_operator/tasks/workload.yml
```diff
@@ -52,6 +52,42 @@
     - r_gitea_deployment.resources[0].status.availableReplicas is defined
     - r_gitea_deployment.resources[0].status.availableReplicas | int == r_gitea_deployment.resources[0].spec.replicas | int
 
+- name: search for gitea pod
```
wkulhanek, line 55: I'd put all that into a block under a

when: ocp4_workload_gitea_operator_install_admin_user | bool

condition.

ansible/roles_ocp_workloads/ocp4_workload_gitea_operator/defaults/main.yml
thoraxe: I would think that you would always want an admin user. Without an admin user you can't create actual users, people would have to register, and etc.

I can't envision a situation where having the admin user would be something you didn't want.

PR #2189, opened 2020-07-18
Fix occurences of variable 'update_all_packages', rename it to 'update_packages'

fridim:
The variable `update_all_packages` doesn't exist. In the role 'common' it is
simply called `update_packages`.

This commit if applied, will rename the variables, or just remove it in case the
other variable is already in the file.

sample_vars are usually safe to change, default_vars need a little more extra
precautions.

##### ISSUE TYPE

- Bugfix Pull Request


stencell:
This looks safe to me.

PR #2190, opened 2020-07-20
Use specific EPEL repo when current OS is RHEL8.1.

ggbecker:
##### SUMMARY
Use specific EPEL repo when installing packages on RHEL8.1 to avoid any missing dependencies when installing `xrdp` packages.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
rhel-graphical

### DIFF
--- a/ansible/roles/rhel-graphical/tasks/main.yml
+++ b/ansible/roles/rhel-graphical/tasks/main.yml
@@ -85,6 +85,17 @@
         - "xorg-*"
         - coreutils
         - wget
+
+    # xrdp-selinux requires selinux-policy and newer versions of this
+    # package in EPEL require newer selinux-policy which is not available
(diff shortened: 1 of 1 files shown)


stencell:
We're very much against using EPEL since it has broken us so many times in the past. Is there any other alternative here or is this the only way?

ggbecker:
> We're very much against using EPEL since it has broken us so many times in the past. Is there any other alternative here or is this the only way?

We are considering to remove EPEL dependencies (xrdp packages) from this role. It's mainly for remote access to the GUI using RDP. But it's obvious the problems it can cause in the future, we also support VNC for remote access and this one can be installed using regular RHEL repositories.

ggbecker:
Closing in favour of #2248

PR #2214, opened 2020-07-23
Change role infra-osp-create-inventory

stencell:
##### SUMMARY
Ansible seems to randomly select the private IP when there are multiple IPs available to OpenStack instances with multiple NICs. This is specifically affecting the `kni-osp` config where the `private_v4` would sometimes be the pxe network and sometimes the app network - when it always needs to be the app network in this case.

With this change, a config can specify the following two vars to make sure we select the right IP to use. This will default to false, so it will only run if enabled by the config.

`mutli_network` - bool
`multi_network_primary` - full name of the network that you want the IPs selected from

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
infra-osp-create-inventory

### DIFF
--- a/ansible/roles-infra/infra-osp-create-inventory/tasks/main.yml
+++ b/ansible/roles-infra/infra-osp-create-inventory/tasks/main.yml
@@ -35,18 +35,13 @@
       name: "{{ server | json_query(_name_selector) | default(server.name) }}"
       original_name: "{{ server.name }}"
       groups:
-      #TODO: remove thos tag_*
-      # - "tag_Project_{{stack_tag}}"
-      # - "tag_{{ stack_tag}} | default('unknowns') }}"
       - "{{ server.metadata.ostype | default('unknowns') }}"
+      - "{{ server.metadata.AnsibleGroup | default('') }}"
       ansible_user: "{{ ansible_user }}"
       remote_user: "{{ remote_user }}"
-      # ansible_ssh_private_key_file: "{{item['key_name']}}"
-      # key_name: "{{item['key_name']}}"
       state: "{{ server.status }}"
       instance_id: "{{ server.id }}"
       isolated: "{{ server.metadata.isolated | default(false) }}"
-      # private_dns_name: "{{item['private_dns_name']}}"
       private_ip_address: "{{ server.private_v4 }}"
       public_ip_address: "{{ server.public_v4 | default('') }}"
       image_id: "{{ server.image.id | default('') }}"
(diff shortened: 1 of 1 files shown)


ansible/roles-infra/infra-osp-create-inventory/tasks/main.yml
```diff
@@ -61,6 +56,22 @@
     - create_inventory
     - must
 
+  # This should be used if you have VMs with multiple NICs.
+  # Make sure you have multi_network_primary variable set indicating the
+  # name of the network you want to pull the IP from
+  - name: Make sure private IP is set if using multiple networks
+    when: multi_network | default(false)
```
wkulhanek, line 63: Does this need a `| bool` at the end?

fridim: if multi_network is a string, like "false", it gives the following warning:

```
[DEPRECATION WARNING]: evaluating 'f2' as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the 
future. Also see CONDITIONAL_BARE_VARS configuration toggle. This feature will be removed in version 2.12. Deprecation warnings can be disabled by
 setting deprecation_warnings=False in ansible.cfg.```

So, yes, i think we want `|bool`

File: ansible/roles-infra/infra-osp-create-inventory/tasks/main.yml [hunk 2]
```diff
@@ -35,18 +35,13 @@
       name: "{{ server | json_query(_name_selector) | default(server.name) }}"
       original_name: "{{ server.name }}"
       groups:
-      #TODO: remove thos tag_*
-      # - "tag_Project_{{stack_tag}}"
-      # - "tag_{{ stack_tag}} | default('unknowns') }}"
       - "{{ server.metadata.ostype | default('unknowns') }}"
+      - "{{ server.metadata.AnsibleGroup | default('') }}"
```
fridim, line 39: Why do you add this line ?

it is done later in the code

stencell: Whoops, I was testing adding that in the same task and forgot to remove it. I still can't understand why this doesn't work but I don't have time to fix it.

stencell:
> Don't really understand what's happening. :-) But I assume @stencell tested this. Only one defensive programming comment.

What is not clear? This won't affect anything that doesn't set `multi_network` to true, so everything else will work as it always has. I did test with studentVM, which does not enable this var and it worked fine.

PR #2303, opened 2020-08-05
[Config][Change] - DIL Streaming workshop - Replace Infinispan with Data Grid operator

weimeilin79:
##### Replace Infinispan with Data Grid operator

Problem in Inifinispan 2.0.22. Decide to switch to DataGrid G version. 

##### COMPONENT NAME
Data Pool in LAB 3 workshop


ansible/roles/ocp4-workload-dil-streaming/tasks/provision_dg.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+- name: Evaluate Data Grid Operator Group
+  k8s:
+    state: present
+    resource_definition: "{{ lookup('template', 'datagrid-operatorgroup.yaml.j2') }}"
+
+- name: Evaluate Data Grid Subscription
+  k8s:
+    state: present
+    resource_definition: "{{ lookup('template', 'datagrid-subscription.yaml.j2') }}"
```
wkulhanek, line 17: This really should have checks to ensure the operator installed successfully.
Example: https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_serverless/tasks/workload.yml#L13-L79

Also is there a risk of the operators changing? If so a Catalog Snapshot should be used to prevent environments from breaking: https://github.com/redhat-cop/agnosticd/blob/development/docs/Operator_Catalog_Snapshots.adoc

wkulhanek: Will be addressed in subsequent PRs. This is a break fix for a broken environment

PR #2337, opened 2020-08-12
Fix missing variables in disco lab

wkulhanek:


stencell:
Must've been a while since we tagged an update. These are definitely needed if we pull in all the recent changes.

PR #2339, opened 2020-08-12
Fix ocp4 cluster shutdown

wkulhanek:
##### SUMMARY
@rut31337 noticed that *ocp4-cluster* config did not get deleted properly (on AWS) if the environment had been idled (as in VMs stopped).

* For OCP4 Cluster split the `destroy_env` playbook into 4 playbooks - one for each cloud - which dramatically simplified the logic in each playbook.
** Call the cloud specific logic from ocp4-cluster's main destroy_env.yml playbook.
* AWS: Added logic start instances if they are shut down.
* OpenStack: Added logic to add metadata tags to instances (because installer does not) in order to support proper lifecycle (start / stop of environments)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-cluster

All destroy logic


ansible/lifecycle_ec2.yml
```diff
@@ -11,6 +11,7 @@
         wait: no
         filters:
           "tag:guid": "{{ guid }}"
+          instance-state-name: running
```
stencell, line 14: Out of curiousity, why are these needed. If you tell the module to stop the instance...it knows to stop it if it is running and not stop it if if it stopped.

wkulhanek: It's a safeguard when this is run shortly after provisioning. Because then the terminated bootstrap VM is still there - and the module will throw up...

ansible/destroy.yml
```diff
@@ -20,8 +20,13 @@
 
 - import_playbook: >-
     {{ lookup('first_found', {
-         'files': [ 'destroy_env.yml', cloud_provider + '_destroy_env.yml' ],
-         'paths': [ 'configs/' + env_type, 'cloud_providers/' ]
+         'files': [ 'destroy_env.yml',
+                    'destroy_env_' + cloud_provider + '.yml',
```
stencell, line 24: I'm not opposed to this change...but rather than making the change here, can't we just leave the generic `destroy_env.yml` and import the right set of tasks on conditional? That follows the pattern we use in many other places.

If ec2, import destroy_env_ec2.yml
If osp, import destroy_env_osp.yml
etc.

wkulhanek: Hm. That's not a bad idea. I'll change and test.

PR #2340, opened 2020-08-13
Update osp cloud provider for heat stack throttling

stencell:
##### SUMMARY
These changes will update the OSP cloud provider to limit the number of concurrent HOT stack creations and updates that can happen in parallel. There are two variables that can be provided, which are currently default to a value of 10:

```
max_update_in_progress
max_create_in_progress
```

Upon pulling the list of current stacks, the deployment will check to make sure the number of in progress doesn't exceed the defined threshold. If it does, it will back off and check again 2-5 minutes later which is offset as a random value between 120-300.

These changes also preserve functionality when this is being used by a non-admin user in a sandbox or other type of environment. The task that checks for the number of stacks will default to the "member" credentials, so it will only see the stacks it has access to - which is not all of them.

Running this as an admin will have visibility to all of the stacks and is magical. 🌈 🦄 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
infra-osp-template-create


ansible/roles-infra/infra-osp-template-create/defaults/main.yml
```diff
@@ -10,3 +10,6 @@ cf_tags:
 cloud_tags: {}
 
 cloudformation_retries: 0
+
+max_update_in_progress: 10
+max_create_in_progress: 10
```
jkupferer, line 15: New variables should have prefixes that match the role, cloud provider, or config.

I think cloud provider prefix is appropriate here and would recommend variable naming `osp_max_update_in_progress` and `osp_max_create_in_progress`.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,98 @@
...
+  - r_stack_list.stdout | from_json | json_query(update_stacks) | length <= max_update_in_progress
+  retries: 10
+  delay: "{{ 300 | random(start=120) }}"
+  vars:
+    create_stacks: >-
+      [?"Stack Status"=='CREATE_IN_PROGRESS']
+    update_stacks: >-
+      [?"Stack Status"=='UPDATE_IN_PROGRESS']
+
+- debug:
```
jkupferer, line 23: Tasks should have names.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,98 @@
...
+    msg: The number of updating stacks is {{ r_stack_list.stdout | from_json | json_query(query) | count | default(0) }}
+  vars:
+    query: >
+      [?"Stack Status"=='UPDATE_IN_PROGRESS']
+  verbosity: 2
+
+- debug:
```
jkupferer, line 30: Tasks should have names

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,98 @@
...
-    - name: Copy infra_key content to output_dir
-      copy:
-        dest: "{{ infra_ssh_key }}"
-        content: "{{ infra_private_key_content }}"
-        mode: 0600
+  - name: Copy infra_key content to output_dir
+    copy:
+      dest: "{{ infra_ssh_key }}"
+      content: "{{ infra_private_key_content }}"
+      mode: 0600
```
jkupferer, line 98: As long as we are touching this line, can we change to use symbolic style, `u=rw,g=,o=`?
Symbolic is best practice as it is clearer for people reading the code and is less likely to to run into octal/decimal issues.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
rut31337: my understanding is that we are not going to support stack updating.  can we filter the query using `openstack stack list --property stack_status=CREATE_IN_PROGRESS -f json` and remove all the updating stuff?

ansible/roles-infra/infra-osp-template-create/defaults/main.yml
stencell: Fixed in 93ab338

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
stencell: Fixed in e233315

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
stencell: This line was only touched to fix indentation with the previous block removal. Changing this, even though small, is unnecessary and is outside of the scope of this PR so I am going to leave it for now.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,100 @@
...
+  - name: Grab outputs
+    command: >-
+      openstack stack show -f json base-stack-{{ guid }}
+    changed_when: false
+    register: r_hot_outputs
 
-    - debug:
-        var: r_hot_outputs
-        verbosity: 3
+  - debug:
```
jkupferer, line 75: Tasks should have names.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,100 @@
...
-    - set_fact:
-        hot_outputs: "{{ (r_hot_outputs.stdout | from_json) }}"
-    
-    - debug:
-        var: hot_outputs
-        verbosity: 3
+  - set_fact:
+      hot_outputs: "{{ (r_hot_outputs.stdout | from_json) }}"
+  
+  - debug:
```
jkupferer, line 82: Tasks should have names.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,100 @@
...
+      infra_private_key_content: "{{ hot_outputs | json_query(query) }}"
+    vars:
+      query: "outputs[?@.output_key=='{{ guid }}-infra_key'].output_value|[0]"
 
-    - name: Set infra_ssh_key path
-      set_fact:
-        infra_ssh_key: "{{ output_dir }}/{{ guid }}_infra_ssh_key.pem"
+  - name: Set infra_ssh_key path
+    set_fact:
+      infra_ssh_key: "{{ output_dir }}/{{ guid }}_infra_ssh_key.pem"
```
jkupferer, line 94: These `set_fact` tasks should likely be combined?

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -1,69 +1,100 @@
+- name: Check for running Heat stacks
+  command: >-
+    openstack stack list -f json
```
jkupferer, line 3: Are we sure this will work and that `openstack stack list` won't page results? Like first 50 of 200?

Should we perhaps filter the list by creating as @rut31337 suggested?

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
rut31337: Note `openstack stack list --property stack_status=CREATE_IN_PROGRESS --property stack_status=UPDATE_IN_PROGRESS` will NOT work.  It will show everything.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
stencell: I'm not touching things that are not in the scope of this PR. This can be fixed separately when someone has more time to test...for now, lets please focus on the feature being added.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
stencell: Moving this to its own comment since it is unrelated to what JK mentioned here.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
stencell: I have not seen this, but I am not sure. We need the full list since we are removing the ability to update a stack...I need the full list to determine whether the stack exists or not and whether we bail or proceed with creating a new one. Only pulling back a filtered list would break this.

stencell:
This should work as is now. It will make 2 API calls. One to get the list of stacks, which we use both for throttling and to validate that the stack doesn't already exist. The second will be to do a `stack show` once the stack is created to get all of our outputs.

If you feel that the full stack list is going to be detrimental, someone can suggest and commit a change that does 3 API calls. One to get a list of create_in_progress. One to check if the stack exists. One to get the stack outputs after stack creation.

I would roll with this until you actually see getting a full stack list is a problem vs overcomplicating the solution.

stencell:
Did anyone do an actual analysis to see if 3 API calls were more expensive than 2 or is this still just a guess?

@jkupferer Can you please post what configs you tested these new changes with to validate everything works? I typically like to see `ocp4-cluster` and/or `studentvm`. At least through the cloud provider pieces.

jkupferer:
@stencell I have not run through a test with these changes yet.

stencell:
@jkupferer I tested multiple deploys of `studentvm` and everything looks good. I can't approve my own PR though :)

PR #2344, opened 2020-08-17
[WIP] Update osp-migration with new tasks to find the external network ID of the cluster

agonzalezrh:
Currently, external network ID are defined on the secret for each cluster. With this PR it will find automatically this ID

### DIFF
--- a/ansible/configs/osp-migration/infra.yml
+++ b/ansible/configs/osp-migration/infra.yml
@@ -81,6 +81,23 @@
         create_objects_user: "{{ guid }}"
         create_objects_password: "{{ osp_migration_api_pass }}"
       when: use_openstack_centralized_router | default(False) == False
+      
+    - name: Gather OSP network external network
+      os_networks_facts:
+        name: "{{ osp_external_network | default('external') }}"
+        filters:
+          router:external: true
+      register: r_osp_n_facts
+
+    - name: Fail is there is not an external network found
+      assert:
+        that:
+          - (r_osp_n_facts.ansible_facts.openstack_networks | count) > 0
+        msg: "No OSP external network found"
+
+    - set_fact:
+        external_network:
+          "{{ (r_osp_n_facts.ansible_facts.openstack_networks | first).id }}"
 
     - name: Set authentication information to create objects
       set_fact:
--- a/ansible/configs/osp-migration/infra.yml
+++ b/ansible/configs/osp-migration/infra.yml
@@ -46,7 +46,7 @@
         parameters:
           project_name: "{{ osp_project_name }}"
           project_guid: "{{ guid }}"
-          project_description: "created:{{ ansible_date_time.epoch }}"
+          project_description: "created:{{ ansible_date_time.epoch | default('') }}"
           project_api_user: "{{ guid }}"
           project_api_pass: "{{ osp_migration_api_pass }}"
           blueprint: "{{ project }}"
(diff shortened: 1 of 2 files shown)


ansible/configs/osp-migration/infra.yml
```diff
@@ -46,7 +46,7 @@
         parameters:
           project_name: "{{ osp_project_name }}"
           project_guid: "{{ guid }}"
-          project_description: "created:{{ ansible_date_time.epoch }}"
+          project_description: "created:{{ ansible_date_time.epoch | default('') }}"
```
jkupferer, line 49: This should not need a default... If `ansible_date_time` is not set, then the `setup` module (or `gather_facts`) should be run.

ansible/configs/osp-migration/infra.yml
```diff
@@ -81,6 +81,23 @@
...
+      register: r_osp_n_facts
+
+    - name: Fail is there is not an external network found
+      assert:
+        that:
+          - (r_osp_n_facts.ansible_facts.openstack_networks | count) > 0
+        msg: "No OSP external network found"
+
+    - set_fact:
+        external_network:
```
jkupferer, line 99: Please include a name for every task.

Also, is it difficult to change this variable name to something like `osp_external_network`?

ansible/configs/osp-migration/infra.yml
```diff
@@ -81,6 +81,23 @@
         create_objects_user: "{{ guid }}"
         create_objects_password: "{{ osp_migration_api_pass }}"
       when: use_openstack_centralized_router | default(False) == False
+      
+    - name: Gather OSP network external network
+      os_networks_facts:
```
jkupferer, line 86: Please switch to using Ansible 2.9 and `os_networks_info`.

ansible/configs/osp-migration/infra.yml
```diff
@@ -81,6 +81,23 @@
...
+        name: "{{ osp_external_network | default('external') }}"
+        filters:
+          router:external: true
+      register: r_osp_n_facts
+
+    - name: Fail is there is not an external network found
+      assert:
+        that:
+          - (r_osp_n_facts.ansible_facts.openstack_networks | count) > 0
+        msg: "No OSP external network found"
```
jkupferer, line 96: This assertion can be combined with the previous task using `failed_when`:

```
    - name: Gather OSP network external network
      os_networks_facts:
        name: "{{ osp_external_network | default('external') }}"
        filters:
          router:external: true
      register: r_osp_n_facts
      failed_when:
      - (r_osp_n_facts.openstack_networks | count) == 0
```

ansible/configs/osp-migration/infra.yml
```diff
@@ -81,6 +81,23 @@
...
+
+    - set_fact:
+        external_network:
+          "{{ (r_osp_n_facts.ansible_facts.openstack_networks | first).id }}"
```
jkupferer, line 100: When updating to Ansible 2.9 `os_networks_info` it will no longer set `ansible_facts`. So this will be `r_osp_n_facts.openstack_networks`

stencell:
It will be scope creep on this PR, but this config is one that is NOT using the agnosticD OSP cloud provider. The `infra.yml` needs the throttling built into it or any of these Ravello-now-Novello items that are provisioned won't be restricted to a max concurrent.

@jkupferer @rut31337

stencell:
@agonzalezrh Is this still needed? It has been open for quite a while.

PR #2363, opened 2020-08-18
Change role infra-ec2-template-create to add uuid to cloud tags

jkupferer:
##### SUMMARY

Add uuid tag with default empty string to standard ec2 instance tags.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Infrastructure role infra-ec2-template-create

### DIFF
--- a/ansible/roles-infra/infra-ec2-template-create/defaults/main.yml
+++ b/ansible/roles-infra/infra-ec2-template-create/defaults/main.yml
@@ -5,6 +5,7 @@ cf_tags:
   owner: "{{ email | default(user) | default('unknown') }}"
   env_type: "{{ env_type }}"
   guid: "{{ guid }}"
+  uuid: "{{ uuid | default('') }}"
 
 # custom additional tags :
 cloud_tags: {}


stencell:
@rut31337 Didn't we run into a problem where we had too many tags? I can't remember what the number was, but if I am remembering correctly we should be careful about adding more like this.

@jkupferer Can we do `guid` or `uuid`? will there ever be both of them or always just one or the other?

jkupferer:
@stencell I think we do want both `guid` and `uuid`. The `guid` is really helpful for quickly finding things while the `uuid` is better suited for use in automation. There are other tags that we will be phasing out because as we begin supporting pre-provisioning environments for students. For example, `email` and `owner` tags should probably be going away in favor of looking this information up by the `uuid`.

stencell:
@jkupferer Ok. I would like to see one tag left that is easily and quickly identifiable, though. Either email or owner...we don't need both for that. In a pinch when you are trying to figure things out quickly, looking at the tags and immediately being able to identify who it belongs to is much better than having to then go look up information in another tool.

jkupferer:
@stencell We are moving to a world where lab environments will be provisioned before they are requested. For example, for OCP4 clusters the long term goal is to have pre-installed but otherwise not configured clusters ready to go layer on workloads. We won't reliably be able to go back and re-tag resources with owner or email... though this discussion gets beyond simply putting in the uuid tag.

jkupferer:
Practically speaking, everything is already getting a `uuid` tag. So adding this as a standard tag in agnosticd won't cause any trouble at least as far as tag count limits.

stencell:
@jkupferer Not everything is going to be installed in this pre-provisioned way...that will just be some of what is deployed. There are still people that will use agnosticd to deploy on-demand, whether it is from testing, labs, etc. as well as other organizations that have adopted its use that have nothing to do with the work you are talking about. We need to keep that in mind when. Having an owner tag is such a small thing but potentially so helpful in quick identification. I also find it highly unlikely that we are going to be deploying a bunch of AWS resources ahead of time just to sit there and do...nothing while we pay for it, so this will probably be a moot point wrt ec2 tags. 

But, as you said, this can be a different discussion.

I'd like @rut31337 to give an okay to this since he was involved in the number of tags when we ran into the problems previously. CloudForms passes a bunch of stuff in + what we define. I don't want this to break things by accident without a test done first.

rut31337:
We are already writing uuid from CF so I dont think we are adding any more tags?  I think the limit was like 4 or 5 tags.  And the lowest common denominator was route53.  EC2 can have a lot more tags, but everything gets tagged with the same cloud tags and Routr53 would barf, killing the stack.

jkupferer:
@rut31337 The trouble is that cloudforms injects `uuid` into the `cloud_tags` variable, but we need to override the `uuid` for babylon. So the new code filters out the `guid` and `uuid` sent from cloudforms...

https://github.com/redhat-cop/babylon/blob/master/playbooks/service-provision.yaml#L43-L45

So this PR is about replacing that functionality... I could hack it into another place, but it seems to me that putting this into agnosticd and handling it like we do guid is the best way to go.

prakhar1985 (approved):

stencell:
@prakhar1985 Why did you merge this when there was an ongoing discussion about whether to implement this change and no verification that it had been tested? It is not appropriate to come in and approve and merge something in this situation.

prakhar1985:
@stencell I thought @jkupferer had requested for approval no body could review it. It was a small change as Jonathan mentioned in chat. I know it’s not appropriate to merge anything which is not been tested or verified. Thanks for letting me know. Sorry for merging it if it breaks anything.

PR #2364, opened 2020-08-19
Update ocp4-workload-ccnrd for upgrade MTA 5.0 from RHAMT 4.3.1

danieloh30:
##### SUMMARY
RHMAT 4.3.1 product is changed to MTA 5.0 recently so CCN workshop(module 1) needs to upgrade for this.

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml
+++ b/ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml
@@ -1,31 +1,32 @@
 ---
-# deploy rhamt
-- name: Deploy RHAMT
+# deploy mta
+- name: Deploy MTA
   shell: oc process -f - | oc create -n labs-infra  -f -
   args:
-    stdin: "{{ lookup('file', './files/rhamt-template.yaml') }}"
+    stdin: "{{ lookup('file', './files/mta-template.yaml') }}"
 
-- name: TODO - scale RHAMT executor to quarter the number of users
-  shell: |
-         oc scale dc/rhamt-web-console-executor --replicas={{ ((num_users|int) / 3) | int }} -n labs-infra
-
-- name: remove liveness and readiness for RHAMT web console
-  shell: |
-         oc set probe dc/rhamt-web-console  -n labs-infra --remove --readiness --liveness
-
-# wait for RHAMT to be running
-- name: wait for RHAMT to be running
+- name: scale MTA executor to quarter the number of users
+  k8s_scale:
+    api_version: apps.openshift.io/v1
+    kind: DeploymentConfig
+    name: mta-web-console-executor
+    namespace: labs-infra
+    replicas: "{{ (num_users|int / 3) | int }}"
+    wait: no
+ 
+# wait for MTA to be running
+- name: wait for MTA to be running
   uri:
-    url: http://rhamt-web-console-labs-infra.{{ route_subdomain }}/rhamt-web/
+    url: http://mta-web-console-labs-infra.{{ route_subdomain }}/mta-web/
   register: result
   until: result.status == 200
   retries: "120"
   delay: "15"
 
-# Get admin token for rhamt's sso
+# Get admin token for mta's sso
 - name: Get SSO admin token
   uri:
-    url: https://secure-rhamt-web-console-labs-infra.{{ route_subdomain }}/auth/realms/master/protocol/openid-connect/token
+    url: https://secure-mta-web-console-labs-infra.{{ route_subdomain }}/auth/realms/master/protocol/openid-connect/token
     method: POST
     validate_certs: no
     body:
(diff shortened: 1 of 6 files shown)


ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml
```diff
@@ -1,31 +1,31 @@
...
-- name: Deploy RHAMT
+# deploy mta
+- name: Deploy MTA
   shell: oc process -f - | oc create -n labs-infra  -f -
   args:
-    stdin: "{{ lookup('file', './files/rhamt-template.yaml') }}"
+    stdin: "{{ lookup('file', './files/mta-template.yaml') }}"
 
-- name: TODO - scale RHAMT executor to quarter the number of users
+- name: TODO - scale MTA executor to quarter the number of users
```
stencell, line 8: Is this still TODO? Please update task name if not.

danieloh30: Removed.

ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml
```diff
@@ -1,31 +1,31 @@
...
   shell: oc process -f - | oc create -n labs-infra  -f -
   args:
-    stdin: "{{ lookup('file', './files/rhamt-template.yaml') }}"
+    stdin: "{{ lookup('file', './files/mta-template.yaml') }}"
 
-- name: TODO - scale RHAMT executor to quarter the number of users
+- name: scale MTA executor to quarter the number of users
   shell: |
-         oc scale dc/rhamt-web-console-executor --replicas={{ ((num_users|int) / 3) | int }} -n labs-infra
+         oc scale dc/mta-web-console-executor --replicas={{ ((num_users|int) / 3) | int }} -n labs-infra
```
stencell, line 10: Please convert this to use `k8s` module. Something as simple as this should be sufficient.

```
k8s:
  state: present
  apply: yes
  definition:
    apiVersion: apps.openshift.io/v1
    kind: DeploymentConfig
    metadata:
      name: mta-web-console-executor
      namespace: labs-infra
    spec:
      replicas: {{ ((num_users|int) / 3) | int }}
```

ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml
```diff
@@ -1,31 +1,31 @@
...
 
-- name: remove liveness and readiness for RHAMT web console
+- name: remove liveness and readiness for MTA web console
   shell: |
-         oc set probe dc/rhamt-web-console  -n labs-infra --remove --readiness --liveness
+         oc set probe dc/mta-web-console  -n labs-infra --remove --readiness --liveness
```
stencell, line 14: Please convert this to use `k8s` module. You can probably merge this with the previous task to do it in one with `strategic-merge` or `apply` in module.

danieloh30: Any example to convert oc set probe using k8 module?

stencell: actually before I go down that path...why do you not want any liveness or readiness probes? That seems like not a good idea to remove them unless they are causing some kind of issue?

danieloh30: because the pod never comes up for the liveness and readiness check.

danieloh30: @stencell I'm trying to test without oc set probe. Btw, I got an error b/c replicas: {{ ((num_users|int) / 3) | int }} doesn't work.
fatal: [bastion.6b94.example.opentlc.com]: FAILED! => {"reason": "We were unable to read either as JSON nor YAML, these are the errors we got from each:\nJSON: Expecting value: line 1 column 1 (char 0)\n\nSyntax Error while loading YAML.\n  found unacceptable key (unhashable type: 'AnsibleMapping')\n\nThe error appears to be in '/Users/danieloh/RedHat-Middleware-Workshops/agnosticd/ansible/roles/ocp4-workload-ccnrd/tasks/install-mta.yaml': line 19, column 20, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n      spec:\n        replicas: {{ ((num_users|int) / 3) | int }}\n                   ^ here\nWe could be wrong, but this one looks like it might be an issue with\nmissing quotes. Always quote template expression brackets when they\nstart a value. For instance:\n\n    with_items:\n      - {{ foo }}\n\nShould be written as:\n\n    with_items:\n      - \"{{ foo }}\"\n"}

danieloh30: Ok. I fixed this.

PR #2391, opened 2020-08-24
assign static ips from outside dhcp range for customer ports for ocp4…

augustrh:
…-cnvlab workloads

##### SUMMARY

To manage the fact that OVN is blocking DHCP even on ports without port security we need to ensure an IP is set. This allows us to set the IP using cloud-init and ensure the CNV VMs get an IP from the "public network" in the lab.

Normally we'd like to have neutron managing this (or standalone DHCP in a non-OSP env) but in Orange with OVN it has not been wokring (BZ coming soon).

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME

CNV Hands on Lab.

##### ADDITIONAL INFORMATION

```paste below

```


NonyNo3lle (approved):
Necessary workaround for CNV Lab Demo/Lab to work in upcoming hackaton

PR #2410, opened 2020-08-27
Add RHEL81NBDE AMI identifiers.

ggbecker:
##### SUMMARY

Add AMI for custom RHEL81 images to be used by NBDE exercise from RHEL Security Lab

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
rhel-security


fridim:
Hello @ggbecker 

Is this image public ?

Who is the owner of the image ?


Any chance you can add this to the `find_ami.sh` script ? https://github.com/redhat-cop/agnosticd/blob/development/tools/find_ami.sh

Indeed, the `aws_ami_region_mapping` variable is generated and not manually maintained.

ggbecker:
> Hello @ggbecker
> 
> Is this image public ?
> 
> Who is the owner of the image ?
> 
> Any chance you can add this to the `find_ami.sh` script ? https://github.com/redhat-cop/agnosticd/blob/development/tools/find_ami.sh
> 
> Indeed, the `aws_ami_region_mapping` variable is generated and not manually maintained.

I think the image is public, but it's on @sergio-correia 's AWS account. The idea was to transfer/copy to an events account so it is always available.

What should we do to handover this image to you?

stencell:
@ggbecker I think right now, let's not worry about the image being handed over. We can deal with that separately. Either way, this script will still need to be updated. 

Can you provide the owner account number where the images currently live? I want to follow the same pattern in the script to minimize changes.

When we do eventually pull the AMIs into our account, I will just update this again - but I don't want that to delay it.

stencell:
Actually, since it is public I was able to find the account. I'll see if I can get a change in tonight for the script to generate a new mapping to include this.

ggbecker:
> Actually, since it is public I was able to find the account. I'll see if I can get a change in tonight for the script to generate a new mapping to include this.

That's great. Thanks.

stencell:
This is being fixed in #2421

PR #2421, opened 2020-09-01
Update find_ami and ec2 image mappings

stencell:
##### SUMMARY

The rhel-security config requires a custom image. The `find_ami` script has been updated to find this image and the mappings have also been updated.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
infra-ec2-template-generate
tools/find_ami.sh


ggbecker:
@stencell just to make sure, even though the images were found from this account:

[`- 719622469867 # GPE (RHPDS prod)`](https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-aws-sandbox/defaults/main.yml#L41)

They are actually just shared from Sergio's personal account, and it still needs to be officially uploaded to a different account so Sergio can remove it from his own account.

stencell:
I copied them into this account last night, that is why we are pulling from this account and not the one you provided previously.

PR #2422, opened 2020-09-01
New role vscode-server

miteshget:
##### SUMMARY
A new role vscode-server which installs vscode on bastion/control node and serves it via https/http protocol using nginx daemon. 

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
Role- vscode-server


prakhar1985 (approved):
Looks good

PR #2429, opened 2020-09-01
 [Bug/Feature] New project repo pointing to RedHat-Middleware-Workshops 

weimeilin79:
##### SUMMARY
Removing CHE and OpenShift SSO Auth. 
Adding CHE workspace start up validation 
Update to new transferred REPO  RedHat-Middleware-Workshops 

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

CHE workspace Login update
Repo Updates


ansible/roles/ocp4-workload-dil-streaming/tasks/provision_dg.yaml
```diff
@@ -15,3 +15,16 @@
   k8s:
     state: present
     resource_definition: "{{ lookup('template', 'datagrid-subscription.yaml.j2') }}"
+    
+- name: Wait until Operator POD is in running state
```
wkulhanek, line 19: That's probably fine. Instead of the operator `pod` I would check the `deployment` and check that `.spec.replicas` is the same value as `.status.readyReplicas`. Like here: https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_pipelines/tasks/workload.yml#L100-L112

You could also instead check that the CSV is succeeded. Like here: https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_pipelines/tasks/workload.yml#L73-L98

And finally you could add the "Approve install plan if necessary" logic before that. Like here: https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles_ocp_workloads/ocp4_workload_pipelines/tasks/workload.yml#L37-L71

PR #2456, opened 2020-09-04
Fix the command used to retrieve the image registry url

kharyam:
##### SUMMARY

The shell command used to retrieve the OCP internal registry route is incorrect, resulting in a failure of the lab. Fixed the grep.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_dso

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_dso/tasks/infrastructure.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_dso/tasks/infrastructure.yml
@@ -25,11 +25,15 @@
   become: true
 
 - name: retrieve registry route
-  shell: "{{ ocp4_dso_openshift_cli }} get routes --all-namespaces | grep registry | grep default | grep -v console | awk '{print $3}'"
+  k8s_info:
+    api_version: route.openshift.io/v1
+    kind: route
+    name: default-route
+    namespace: openshift-image-registry
   register: image_route
 
 - set_fact:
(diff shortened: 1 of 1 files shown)


ansible/roles_ocp_workloads/ocp4_workload_dso/tasks/infrastructure.yml
```diff
@@ -25,7 +25,7 @@
   become: true
 
 - name: retrieve registry route
-  shell: "{{ ocp4_dso_openshift_cli }} get routes --all-namespaces | grep registry | grep default | grep -v console | awk '{print $3}'"
+  shell: "{{ ocp4_dso_openshift_cli }} get routes --all-namespaces | grep image-registry | awk '{print $3}'"
```
stencell, line 28: Please move this to the k8s_info module. We do not want to keep using `shell` and `command`.

Something like:

```
- k8s_info:
      api_version: imageregistry.operator.openshift.io/v1
      kind: Config
      name: cluster
    register: something
```

stencell (changes requested):
see comment.

ansible/roles_ocp_workloads/ocp4_workload_dso/tasks/infrastructure.yml
kharyam: @stencell Thank you, updated

stencell (approved):

PR #2471, opened 2020-09-08
Add role infra-osp-template-create stack retry

jkupferer:
##### SUMMARY

Add retry for OpenStack template creation. On failure the stack will be deleted and recreated.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role infra-osp-template-create

### DIFF
--- a/ansible/roles-infra/infra-osp-template-create/.yamllint
+++ b/ansible/roles-infra/infra-osp-template-create/.yamllint
@@ -1,0 +1,14 @@
+---
+extends: default
+
+rules:
+  comments:
+    require-starting-space: false
+  comments-indentation: disable
+  indentation:
+    level: warning
+    indent-sequences: consistent
+  line-length:
+    level: warning
+    max: 120
+    allow-non-breakable-inline-mappings: true
--- a/ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
+++ b/ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
@@ -2,86 +2,71 @@
 - name: Wait for less than {{ osp_max_create_in_progress }} Heat stacks in CREATE_IN_PROGRESS
   command: >-
     openstack stack list --property stack_status=CREATE_IN_PROGRESS --format json
-  environment:
-    OS_AUTH_URL: "{{ osp_auth_url }}"
-    OS_USERNAME: "{{ osp_auth_username | default(osp_auth_username_member) }}"
-    OS_PASSWORD: "{{ osp_auth_password | default(osp_auth_password_member) }}"
-    OS_PROJECT_NAME: "{{ osp_project_name }}"
-    OS_PROJECT_DOMAIN_ID: "{{ osp_auth_project_domain }}"
-    OS_USER_DOMAIN_NAME: "{{ osp_auth_user_domain }}"
+  environment: "{{ __infra_osp_template_create_environment }}"
   register: r_stack_list
-  until: 
+  until:
   - r_stack_list.stdout | from_json | count <= osp_max_create_in_progress
   retries: 10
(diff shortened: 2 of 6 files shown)


ansible/roles-infra/infra-osp-template-create/.yamllint
```diff
@@ -0,0 +1,14 @@
+---
```
wkulhanek, line 1: Why are we adding custom .yamllints everywhere? Isn't the one at the AgnosticD root level enough?

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
```diff
@@ -2,86 +2,71 @@
...
+  environment: "{{ __infra_osp_template_create_member_environment }}"
+  register: r_stack_show
+  changed_when: false
+  failed_when: false
+  tags: heat
+
+- when: >-
+    __stack_status in ['CREATE_FAILED', 'NONE']
+  vars:
+    __stack: "{{ r_stack_show.stdout | from_json if r_stack_show.rc == 0 else {} }}"
```
wkulhanek, line 29: Can we not use double underscores, please? How about r_ instead like everywhere else.

wkulhanek (dismissed):
Please replace the double underscore variables. They are hard to read.
Also why the extra yamllint?

ansible/roles-infra/infra-osp-template-create/.yamllint
jkupferer: We don't yet have the codebase to the level where it could pass a yamllint at the root level. We don't even have it to a place where it could pass for `ansible/roles-infra/`. So we continue to go bit by bit, working towards being able to run lint at a higher level.

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
jkupferer: The `r_` prefix is used for registered variables. This is just a task level variable.

The use of underscore is to confirm to the standards set by the roles standards working group, documented here:
https://github.com/oasis-roles/meta_standards, "internal variables (those that are not expected to be set by users) are to be prefixed by two underscores"

jkupferer: I would normally go with a single underscore, but the working group went with two.

fridim (approved):

ansible/roles-infra/infra-osp-template-create/tasks/create_osp_template.yml
fridim: Did we discussed about following the standards working group guidelines and rules ?
(i'm open, just i'm not sure we decided on that for agnosticd).

jkupferer: I don't believe we did. @tonykay and I participated in some of the roles standards development. The standards documented for oasis are fairly new and are being raised to more visibility within Red Hat. Their rules are quite close to our own. They don't have an `r_` prefix standard for registered variables... but I thought we were already on the same page with private vars starting with underscore?

wkulhanek: Why not follow the usual conventions and also prefix private variables with the config / workload name?

jkupferer: The `__stack` and `__stack_status` vars are declared at the task level and are not defined outside of the task, so I felt a short name was more readable and helps make it clear that these are scoped to the task.

In contrast, `__infra_osp_template_create_retry_count` and `__infra_osp_template_create_environment` and `__infra_osp_template_create_member_environment` are exposed outside of the role because of how they are declared and so those include the role name.

PR #2475, opened 2020-09-09
Bump secrets and configmaps quota. Change quota name.

wkulhanek:
##### SUMMARY

Quotas for Secrets were too low. Bumped secrets and configmaps.
ClusterResourceQuota used ocp_username - which for some users contains illegal characters ('_'). Removed ocp_username from the quota name.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_pam_case_mgmt_exp_elt


jkupferer (approved):

jkupferer:
Approved... but thinking now... should the old quota name be left in the remove workload for a time? I'm not sure how much of a concern it may be to miss a quota removal that was provisioned before this is merged and tagged.

wkulhanek:
Remove uses a zipped copy of the AgnosticD repo that was used to deploy. At least I think that's the case. @fridim  ??
So it should be OK.

wkulhanek:
I was also thinking I could change {{ ocp_username }} to {{ ocp_username | regexp_replace('_', '-') }}..... but wasn't sure if that was going to work... the problem is that some OpenTLC user IDs have '_' in their name - which is an illegal character for Kube resources.

PR #2485, opened 2020-09-10
Fix ansible config epel repo url

jkupferer:
##### SUMMARY

Fix epel repo base url in ansible configs.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Configs:
- ansible-gitOps-integration
- ansible-tower-intergation
- three-tier-app

### DIFF
--- a/ansible/configs/ansible-gitOps-integration/files/repos_template.j2
+++ b/ansible/configs/ansible-gitOps-integration/files/repos_template.j2
@@ -7,7 +7,7 @@ gpgcheck=0
 
 {% endfor %}
 [epel]
-name=Extra Packages for Enterprise Linux 7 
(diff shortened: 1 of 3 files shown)


tonykay:
Actually I'm not wild about this change and would much prefer we do each config one by one and more inline [with](https://github.com/redhat-cop/agnosticd/blob/2fc2606ed4877011e6e94440e177426c5b16d45c/ansible/configs/ansible-multitier-infra/files/repos_template.j2#L12. ) https://github.com/redhat-cop/agnosticd/blob/2fc2606ed4877011e6e94440e177426c5b16d45c/ansible/configs/ansible-multitier-infra/files/repos_template.j2#L12

PR #2492, opened 2020-09-11
New bookbag roles.

wkulhanek:
##### SUMMARY

This PR introduces two new bookbag workloads:

*ocp4_workload_bookbag_setup*: this workload builds a bookbag image from a git repository.
*ocp4_workload_bookbag_user*: this workload deploys a bookbag for a user in a new project from the image that the other role created.

Use case:
On a shared cluster use *ocp4_workload_bookbag_setup* as an `infra_workload` to create the image.
Then use *ocp4_workload_bookbag_setup* as a `student_workload` to deploy x number of bookbags

Second use case:
Create the image when setting up a shared cluster.
On demand when provisioning an environment deploy the workloads that make up the environment and the ocp4_workload_bookbag_user role to deploy the matching bookbag.

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
ocp4_workload_bookbag_setup
ocp4_workload_bookbag_user

Example variables:

\# Bookbag Setup
ocp4_workload_bookbag_setup_project: "bookbag"
ocp4_workload_bookbag_setup_project_display: "Bookbag RHTR2020 OCP4 Pipelines"
ocp4_workload_bookbag_setup_git_repo: https://github.com/redhat-gpte-devopsautomation/rhtr2020_pipelines
ocp4_workload_bookbag_setup_git_version: master
ocp4_workload_bookbag_setup_image_name: rhtr2020pipelines

\# Bookbag User
ocp4_workload_bookbag_user_project: "bookbag-wkrhtr"
ocp4_workload_bookbag_user_project_display: "Bookbag for user {{ ocp_username }}"
ocp4_workload_bookbag_user_deployment_name: bookbag
ocp4_workload_bookbag_image_name: image-registry.openshift-image-registry.svc:5000/bookbag/rhtr2020pipelines
ocp4_workload_bookbag_image_tag: latest
ocp4_workload_bookbag_image_namespace: "bookbag"

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
@@ -11,6 +11,13 @@ ocp4_workload_bookbag_user_project_display: "Bookbag for user {{ ocp_username }}
 ocp4_workload_bookbag_user_deployment_name: bookbag
 
 # Bookbag Image
-ocp4_workload_bookbag_image_namespace: "bookbags"
-ocp4_workload_bookbag_image_name: image-registry.openshift-image-registry.svc:5000/bookbag/bookbag
-ocp4_workload_bookbag_image_tag: latest
+ocp4_workload_bookbag_user_image_namespace: "bookbags"
+ocp4_workload_bookbag_user_image_name: image-registry.openshift-image-registry.svc:5000/bookbag/bookbag
+ocp4_workload_bookbag_user_image_tag: latest
+
+# Bookbag configuration
+ocp4_workload_bookbag_user_console_deploy: true
+ocp4_workload_bookbag_user_console_image: quay.io/openshift/origin-console:4.5
+ocp4_workload_bookbag_user_console_branding: openshift
+# Role to grant to the service account
+ocp4_workload_bookbag_user_role: edit
--- a/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
@@ -1,0 +1,16 @@
+---
+become_override: false
+ocp_username: opentlc-mgr
+silent: false
+
+# Project in which to deploy the bookbag deployment
+ocp4_workload_bookbag_user_project: "bookbag-{{ guid }}"
+ocp4_workload_bookbag_user_project_display: "Bookbag for user {{ ocp_username }}"
+
+# Name of bookbag deployment
+ocp4_workload_bookbag_user_deployment_name: bookbag
+
+# Bookbag Image
+ocp4_workload_bookbag_image_namespace: "bookbags"
+ocp4_workload_bookbag_image_name: image-registry.openshift-image-registry.svc:5000/bookbag/bookbag
+ocp4_workload_bookbag_image_tag: latest
--- a/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
@@ -17,7 +17,14 @@ ocp4_workload_bookbag_user_image_tag: latest
 
 # Bookbag configuration
 ocp4_workload_bookbag_user_console_deploy: true
-ocp4_workload_bookbag_user_console_image: quay.io/openshift/origin-console:4.5
 ocp4_workload_bookbag_user_console_branding: openshift
 # Role to grant to the service account
 ocp4_workload_bookbag_user_role: edit
+# The workload uses the same console image that is used in the
+# deployed cluster. By setting the next variable that can be overridden.
+# Leave empty to use the matching cluster console image.
+ocp4_workload_bookbag_user_console_image_override: ""
+# ocp4_workload_bookbag_user_console_image_override: quay.io/openshift/origin-console:4.5
+
+ocp4_workload_bookbag_user_create_pvc: false
+ocp4_workload_bookbag_user_pvc_size: 1Gi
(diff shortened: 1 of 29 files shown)


ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
```diff
@@ -0,0 +1,72 @@
...
+- name: Get User Info
+  set_fact:
+    _user_info_messages: >-
+      {{ lookup('file', hostvars.localhost.output_dir ~ '/user-info.yaml', errors='ignore')
+      | from_yaml }}
+
+- name: Fallback for User Info
+  when: _user_info_messages | default({}) | length == 0
+  set_fact:
+    _user_info_messages: {}
```
jkupferer, line 23: Should default to an empty list.

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Set workshop default vars
+  set_fact:
+    _workshop_default_vars:
+      GUID: "{{ guid }}"
+      USERNAME: "{{ ocp_username }}"
+      user_info_messages: "{{ _user_info_messages }}"
```
jkupferer, line 30: For workshop vars, the user_info_messages list should be joined with newlines to a single string.

jkupferer (changes requested):

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
wkulhanek: Changed to []

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: Fallback for User Info
+  when: _user_info_messages | default({}) | length == 0
+  set_fact:
+    _user_info_messages: {}
+
+- name: Set workshop default vars
+  set_fact:
+    _workshop_default_vars:
+      GUID: "{{ guid }}"
```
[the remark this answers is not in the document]
wkulhanek: updated

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
wkulhanek: Done

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
```diff
@@ -0,0 +1,72 @@
...
+  k8s:
+    state: present
+    definition: "{{ lookup('template', item ) | from_yaml }}"
+  loop:
+  - ./templates/project.yaml.j2
+  - ./templates/service_account.yaml.j2
+  - ./templates/rolebinding.yaml.j2
+  - ./templates/service.yaml.j2
+  - ./templates/route.yaml.j2
+  - ./templates/deployment.yaml.j2
```
jkupferer, line 47: I've been thinking about this as I'm putting together a demo...

I don't think I like using Jinja here. The trouble is that folks doing bookbag development won't be using Jinja, they will be using the OpenShift template in the demo repo. If our deployment doesn't use the same approach then it is bound to diverge and cause problems down the road.

A particular concern is use of the OpenShift console in bookbag. The deployment of the console involves using an init-container and another regular container in the bookbag pod. So that will make for pretty big differences between how some labs are deployed, and more potential trouble if we are not letting the bookbag developer control the deployment definition.

jkupferer (changes requested):

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/tasks/workload.yaml
wkulhanek: I disagree. We should not give folks a template to customize. They can customize the content via the workshop directory. And they can customize the image via the Dockerfile. What else could they need?

And init containers?? These YAML templates are 1:1 copies from the example repo template... There most definitely is no init container: https://github.com/redhat-gpte-devopsautomation/bookbag-template/blob/master/deploy-template.yaml#L65

wkulhanek: Added init Container

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
```diff
@@ -0,0 +1,23 @@
...
+ocp4_workload_bookbag_user_deployment_name: bookbag
+
+# Bookbag Image
+ocp4_workload_bookbag_user_image_namespace: "bookbags"
+ocp4_workload_bookbag_user_image_name: image-registry.openshift-image-registry.svc:5000/bookbag/bookbag
+ocp4_workload_bookbag_user_image_tag: latest
+
+# Bookbag configuration
+ocp4_workload_bookbag_user_console_deploy: true
+ocp4_workload_bookbag_user_console_image: quay.io/openshift/origin-console:4.5
```
jkupferer, line 20: The console image should match the OpenShift cluster version... can we make the default reference the cluster version?

jkupferer (changes requested):

ansible/roles_ocp_workloads/ocp4_workload_bookbag_user/defaults/main.yaml
wkulhanek: Changed to "override". Extracting image from deployed console if override not specified.

jkupferer (approved):

PR #2502, opened 2020-09-15
AWS CloudFormation: auto-select AvailabilityZone

fridim:
This commit, if applied, adds a feature to auto detect availbility zones (AZ)
that can host all the instance types defined in the `instances` variable.

If the AZ is omitted in the CloudFormation template, AWS doesn't guarantee to pick the right AZ and that sometimes results in errors of the type:

> "resource_status_reason: Your requested instance type (c3.large) is not supported in your requested Availability Zone (us-east-1a). Please retry your request by not specifying an Availability Zone or choosing us-east-1b, us-east-1d, us-east-1c, us-east-1e."

In order to fix that, this commit adds a set of tasks in the
infra-ec2-template-generate role that do the following:

- Get all the instance types
- For each instance type, get all AZs that support it
- Calculate the intersection of all AZs => all AZs that can host all
  instance types
- Select the first and set `aws_availability_zone`

This feature is enabled by default and can be disabled by setting the following variable:

```yaml
  infra_ec2_template_generate_auto_select_availability_zone: false
```

The tasks are skipped when any of those conditions is met:

- `infra_ec2_template_generate_auto_select_availability_zone` is `false`
- `aws_availability_zone` is defined
- several subnets are found in `instances`: subnets can be created in
  different AZs, in that case we don't want to interfere with the config.
- AWS CLI doesn't support the `describe-instance-type-offerings` feature

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`infra-ec2-template-generate` role

FYI @rut31337

### DIFF
--- a/ansible/roles-infra/infra-ec2-template-generate/defaults/main.yml
+++ b/ansible/roles-infra/infra-ec2-template-generate/defaults/main.yml
@@ -361,3 +361,8 @@ aws_dns_mapping:
       domain: "sa-east-1.compute.internal"
     "ap-south-1":
       domain: "ap-south-1.compute.internal"
+
+# When infra_ec2_template_generate_auto_select_availability_zone is set to true,
+# the role will set aws_availability_zone to an availability that can host
+# all instances types, unless several subnets are used.
+infra_ec2_template_generate_auto_select_availability_zone: true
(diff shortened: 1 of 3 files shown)


wkulhanek:
That is awesome! Should enable us to use t3 and **a (like m5a) instances now.

wkulhanek (approved):

rut31337:
Perfect thank you!

PR #2503, opened 2020-09-15
Do not gather facts when not necessary

fridim:
Improve performance, do not gather localhost facts.

- Enhancement Pull Request

### DIFF
--- a/ansible/configs/amq-messaging-foundations/post_infra.yml
+++ b/ansible/configs/amq-messaging-foundations/post_infra.yml
@@ -2,6 +2,7 @@
   hosts: localhost
   connection: local
   become: false
+  gather_facts: false
   tags:
   - step002
   - post_infrastructure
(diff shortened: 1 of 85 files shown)


stencell:
While I generally think this is going to be a benign change, was there any validation down for these plays to ensure that they didn't actually use facts that would be gathered with that set to true?

fridim:
It's a fair concern.

I looked at the tasks in the plays, and i didn't see any use of local facts.

I also think we should not need localhost facts anyway, and if we do i'd rather like it to be clearly specify with:

```yaml
# fact FOOBAR  is required
gather_facts: true
```

But to be clear, i haven't tested all those configs.

jkupferer (approved):
I'm a bit nervous about this change, but I like it in principle. Role tasks can always run `setup` later when they need a fact that hasn't been gathered.

jkupferer:
When done right, we should filter what facts we gather using `gather_subset` and include a `when` clause so that we only gather facts as needed:
```
- name: Run setup if gather_facts hasn't been run
  when: ansible_distribution_major_version is not defined
  setup:
    gather_subset: min
```

PR #2535, opened 2020-09-17
Change infra-ec2-template-generate, fix label in loop

fridim:
This commit fixes the following error:

```
TASK [infra-ec2-template-generate : Fail if return code is not 0] **************
failed: [localhost] (item={u'skipped': True, u'_type': u't3a.medium', u'ansible_loop_var': u'_type', u'changed': False, u'skip_reason': u'Conditional result was False'}) => {"ansible_loop_var": "item", "changed": false, "item": {"_type": "t3a.medium", "ansible_loop_var": "_type", "changed": false, "skip_reason": "Conditional result was False", "skipped": true}, "msg": "Failed to template loop_control.label: 'dict object' has no attribute 'cmd'", "skip_reason": "Conditional result was False"}
```


wkulhanek (approved):

ansible/roles-infra/infra-ec2-template-generate/tasks/locate_availability_zones.yml
```diff
@@ -84,7 +84,7 @@
       when: item.rc != 0
       loop: "{{ r_possible_azs.results }}"
       loop_control:
-        label: "{{ item.cmd }}"
+        label: "{{ item.cmd | default(item) }}"
```
jkupferer, line 87: I don't believe this will fix the issue, because it will just fail on `item.rc` in the `when`.

jkupferer (changes requested):

ansible/roles-infra/infra-ec2-template-generate/tasks/locate_availability_zones.yml
jkupferer: We really should get away from using `blocks` like this with `when` clauses. The ansible code is much cleaner using `include_tasks` with `when`, because this pattern avoids evaluating tasks like this. It makes the ansible output cleaner and avoids weird when clause evaluation like this.

fridim: Good point

jkupferer (approved):

PR #2556, opened 2020-09-21
Modify index while deploying multiple instances

makentenza:
##### SUMMARY

This PR address the issues generated after the following change in the AgnosticD core code: https://github.com/redhat-cop/agnosticd/pull/1001

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-hana
sap-rhvh

##### ADDITIONAL INFORMATION
This is an issue affecting multiple CIs in RHPDS and LABS using 2 different configs from AgnosticD. Submitting a PR to solve multiple configs as this is the same issue affecting both configs.


makentenza:
@rut31337 this is to address one of the issues we need to solve today

fridim (approved):

PR #2578, opened 2020-09-23
Add AWS infra support to sap-hana config

makentenza:
#### SUMMARY

Enable the config to be deployed into AWS infra.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
sap-hana


NonyNo3lle:
@jkupferer @fridim please take a look at HEAT --> CloudFormations and general multi cloud-provider logic. Thanks!

ansible/configs/sap-hana/default_vars_ec2.yml
```diff
@@ -0,0 +1,332 @@
...
..."128 GiB\"\n        mount_point: \"/hana/data\"\n        state: present\n      - name: log\n        size: \"64 GiB\"\n        mount_point: \"/hana/log\"\n        state: present\n      - name: shared\n        size: \"256 GiB\"\n        mount_point: \"/hana/shared\"\n        state: present\n      - name: sap\n        size: \"50 GiB\"\n        mount_point: \"/usr/sap\"\n        state: present\nsap_hana_deployment_bundle_path: /software/HANA_installation\nsap_hana_deployment_bundle_sar_file_name: IMDB_SERVER20_046_0-80002031.SAR\nsap_hana_deployment_sapcar_path: /software/SAPCAR\nsap_hana_deployment_sapcar_file_name: SAPCAR_1311-80000935.EXE\nsap_hana_deployment_root_password: \"R3dh4t123!\"\nsap_hana_deployment_sapadm_password: \"R3dh4t123!\"\nsap_hana_deployment_hana_sid: RHE\nsap_hana_deployment_hana_instance_number: \"00\"\nsap_hana_deployment_hana_env_type: development\nsap_hana_deployment_hana_mem_restrict: \"n\"\nsap_hana_deployment_common_master_password: \"R3dh4t123!\"\nsap_hana_deployment_sidadm_password: \"R3dh4t123!\"\nsap_hana_deployment_hana_db_system_password: \"R3dh4t123!\"\nsap_hana_deployment_ase_user_password: \"R3dh4t123!\"\nsap_hana_deployment_apply_license: false'
```
fridim, line 178: you have a password in there, it should be a variable with a generated default

```
              {{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}
```

and we'll override it in agnosticv

If there is a way to break those `variables` into several lines for readability, it would be much better

ansible/configs/sap-hana/default_vars_ec2.yml
```diff
@@ -0,0 +1,332 @@
...
+  - tmux
+  - bind-utils
+  - wget
+  - git
+  - vim-enhanced
+  - at
+
+### Ansible Tower default variables ###
+
+ansible_tower:
```
fridim, line 163: can't you move `ansible_tower` to `default_vars.yml` ?

And have a variable `ansible_tower_disks` that would be different in `default_vars_osp.yml` and `default_vars_ec2.yml` ?

ansible/configs/sap-hana/default_vars_ec2.yml
```diff
@@ -0,0 +1,332 @@
...
+      - job_template: 'sap-hostagent'
+        success_nodes:
+        - job_template: 'sap-preconfigure'
+          success_nodes:
+          - job_template: 'sap-hana-preconfigure'
+            success_nodes:
+            - job_template: 'sap-hana-deployment'
+              success_nodes:
+              - job_template: 'sap-s4hana-deployment'
+          - job_template: 'sap-netweaver-preconfigure'
```
fridim, line 332: is this ec2 specific ?
no => `defaults_vars.yml`

ansible/configs/sap-hana/files/cloud_providers/ec2_cloud_template.j2
```diff
@@ -0,0 +1,373 @@
...
+          Value: {{instance['name']}}.{{aws_dns_zone_private_chomped}} 
+        - Key: publicname
+          Value: {{instance['name']}}.{{aws_dns_zone_public_prefix|d('')}}{{subdomain_base }}
+    {% else %}
+        - Key: Name
+          Value: {{instance['name']}}{{loop.index}}
+        - Key: internaldns
+          Value: {{instance['name']}}{{loop.index}}.{{aws_dns_zone_private_chomped}}
+        - Key: publicname
+          Value: {{instance['name']}}{{loop.index}}.{{aws_dns_zone_public_prefix|d('')}}{{ subdomain_base}}
```
fridim, line 228: The only difference between your template and the default one is those 2 lines.

Do you really need them ?

If yes, i think we can safely add them to the default template and then you can just remove yours from the config.

ansible/configs/sap-hana/files/hosts_template.j2
```diff
@@ -32,4 +33,41 @@ ansible_user={{remote_user}}
...
 sap_domain=automation.local
+{% endif %}
+
+{% if cloud_provider == "ec2" %}
+[hanas]
+{% if instances[3].count == 2 %}
+hana1
+hana2
+{% else %}
+hana1
```
fridim, line 44: Why don't you set `hana_instance_count` and `s4hana_instance_count` for ec2 too, so your host template would be simpler ?
You shouldn't  need to duplicate the whole block for ec2.

ansible/configs/sap-hana/print_info_ec2.yml
```diff
@@ -0,0 +1,47 @@
+- name: Set agnosticd user info messages
+  agnosticd_user_info:
+    msg: "{{ item }}"
+  loop: >-
+    {{
+      [
+        "You can access your bastion via SSH:",
+        "",
+        "ssh cloud-user@{{ bastion_public_dns_chomped }}",
```
fridim, line 9: `s/cloud-user/{{remote_user}}`

ansible/configs/sap-hana/print_info_ec2.yml
```diff
@@ -0,0 +1,47 @@
...
+        "Username: admin",
+        "Password: {{ tower_password }}"
+      ] if tower_instance_count > 0 else []
+      )
+    }}
+
+- name: Set agnosticd user info data for bastion
+  agnosticd_user_info:
+    data:
+      bastion_ssh_command: "ssh cloud-user@{{ bastion_public_dns_chomped }}"
```
fridim, line 36: not cloud-user but `{{ remote_user }}`

ansible/configs/sap-hana/print_info_osp.yml
```diff
@@ -0,0 +1,45 @@
...
+      bastion_ssh_command: "ssh cloud-user@bastion-{{ guid }}.{{ guid }}.{{ osp_cluster_dns_zone }}"
+      bastion_ssh_key: "{{ lookup('file', '{{ output_dir }}/{{ guid }}_infra_ssh_key.pem') }}"
+
+- name: Set agnosticd user info data for tower
+  when: tower_instance_count > 0
+  agnosticd_user_info:
+    data:
+      tower_url: "https://tower-{{ guid }}.{{ guid }}.{{ osp_cluster_dns_zone }}"
+      tower_username: "admin"
+      tower_password: "{{ tower_password }}"
```
fridim, line 45: do you really need 2 print_info.yml files, one for each cloud provider ?

Things that differ could be set in the default_vars_...  files

for example in the common print_info.yml:
```yaml
bastion_ssh_command: "ssh {{ remote_user }}@{{ bastion_public_dns_chomped }}"
```

Then 

default_vars_osp:
```yaml
bastion_public_dns_chomped: bastion-{{ guid }}.{{ guid }}.{{ osp_cluster_dns_zone }}"
```

ansible/configs/sap-hana/sample_vars_ec2.yml
```diff
@@ -0,0 +1,35 @@
+cloud_provider: ec2
+env_type: sap-hana
+aws_region: us-east-1
+HostedZoneId: Z09717862MUAYG7P3BER6
+email: mak@redhat.com
+software_to_deploy: none
+aws_access_key_id: XXXXXXXXXXXX
+aws_secret_access_key: XXXXXXXXXXXX
```
fridim, line 8: those should come from secrets

I would not put secrets here in a versioned directory, but rather include the secret when deploying, like:

```
ansible-playbook main.yaml -e @configs/sap../sample_vars.yml  -e @~/mysecrets/sandbox1234.yml
```

ansible/configs/sap-hana/sample_vars_ec2.yml
```diff
@@ -0,0 +1,35 @@
...
+
+subdomain_base_short: "{{ guid }}"
+subdomain_base_suffix: ".630d.sandbox256.opentlc.com"
+subdomain_base: "{{subdomain_base_short}}{{subdomain_base_suffix}}"
+
+repo_method: satellite
+satellite_org: Red_Hat_GPTE_Labs
+satellite_activationkey: XXXXXXXXXX
+satellite_url: XXXXXXXXXXXXXXX
```
fridim, line 17: same

ansible/configs/sap-hana/sample_vars_ec2.yml
```diff
@@ -0,0 +1,35 @@
...
+use_content_view: true
+
+rh_internal: true
+tower_instance_count: 1
+tower_password: XXXXXXXXXXXXXX
```
fridim, line 22: same

ansible/configs/sap-hana/software.yml
```diff
@@ -66,18 +67,25 @@
...
         state: directory
         owner: cloud-user
+      when: cloud_provider == "osp"
+
+    - name: Ensure /nfs directory is owned by 'ec2-user'
+      file:
+        path: /nfs
+        state: directory
+        owner: ec2-user
+      when: cloud_provider == "ec2"
```
fridim, line 88: you could just use `{{ remote_user }}` and have one task

fridim (changes requested):

ansible/configs/sap-hana/files/hosts_template.j2
makentenza: Even if I add `hana_instance_count` and `s4hana_instance_count` the hostnames for OSP and AWS are different so I still need to keep 2 versions

fridim: it's very weird to me that you don't loop over instances, get the name, then depending on the count create the entries.
Something like:
```jinja2

{% for instance in instances %}
[{{ instance.name }}s]
{% for c in range(1,(instance['count'] |int)+1) %}
{{instance['name']}}{{loop.index}}
{% endfor %}
{% endfor %}
```

fridim (approved):

ansible/configs/sap-hana/pre_infra.yml
```diff
@@ -28,3 +28,13 @@
...
+
+    - name: Find out SnapShot ID for SAP Software
+      ec2_snapshot_info:
+        filters:
+          description: "{{ sap_software_snap_name }}"
+      register: sap_snap
+
+    - name: Set local fact for sap_software_snap_id
+      set_fact:
+        sap_software_snap_id: "{{ sap_snap.snapshots[0].snapshot_id }}"
```
fridim, line 40: Did you test when the snapshot is not found ?
What should be our strategy if the snapshot is not found ?

ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
```diff
@@ -250,6 +250,9 @@ Resources:
             VolumeType: "{{ vol[cloud_provider].type }}"
           {% else %}
             VolumeType: "{{ aws_default_volume_type }}"
+          {% endif %}
+          {% if vol.snapshop is defined %}
```
fridim, line 254: typo here. snapshot, not snapshop

Let's use `snapshot_id` to be explicit 

`instances[].volumes[].snapshot_id`

ansible/configs/sap-hana/default_vars_ec2.yml
```diff
@@ -0,0 +1,163 @@
...
+      value: "bastions"
+    - key: "ostype"
+      value: "linux"
+    - key: ansible_python_interpreter
+      value: /usr/libexec/platform-python
+  rootfs_size: "{{ rootfs_size_bastion }}"
+  volumes:
+    - name: /dev/sdg
+      size: 100
+      snapshop: "{{ sap_software_snap_id }}"
```
fridim, line 87: typo

also please replace with `snapshot_id`

fridim (changes requested):

ansible/configs/sap-hana/pre_infra.yml
```diff
@@ -28,3 +28,13 @@
       args:
         creates: "{{output_dir}}/{{env_authorized_key}}.pub"
       when: set_env_authorized_key | bool
+
+    - name: Find out SnapShot ID for SAP Software
+      ec2_snapshot_info:
+        filters:
+          description: "{{ sap_software_snap_name }}"
```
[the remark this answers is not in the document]
makentenza: Yes, this is going to be used for multiple configs

ansible/configs/sap-hana/pre_infra.yml
makentenza: The deployment must fail then, will capture that

ansible/configs/sap-hana/default_vars_ec2.yml
```diff
@@ -31,8 +31,9 @@ rootfs_size_hana: 200
 
 sap_extra_device: xvdf
 sap_software_device: xvdg
-sap_software_snap_name: "sap-software-v1.0"
-sap_software_snap_id: snap-12345abc123456
+sap_software_snapshot_name: "sap-software-v1.0"
+sap_software_snapshot_id: snap-12345abc123456
+sap_software_snapshot_owner_id: "337334234598"
```
fridim, line 36: those 2 are mandatory variables, so i'd put them in the sample vars instead, unless they could have a good default

fridim (approved):

PR #2597, opened 2020-09-25
Remove snapshot owner filter

makentenza:
##### SUMMARY

Remove one of the filters to search for sap software snapshot

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-hana


makentenza:
@rut31337 as requested I've removed the filter

ansible/configs/sap-hana/pre_infra.yml
```diff
@@ -32,8 +32,7 @@
     - name: Find out SnapShot ID for SAP Software
       ec2_snapshot_info:
         filters:
-          description: "{{ sap_software_snapshot_name }}"
-          owner-id: "{{ sap_software_snapshot_owner_id }}" 
+          description: "{{ sap_software_snapshot_name }}"
```
fridim, line 35: There is a security issue about using only the description here.

We need to ensure the origin of the snapshot.

fridim (changes requested):

ansible/configs/sap-hana/pre_infra.yml
makentenza: @rut31337 please agree with @fridim on this. As long as we know the owner we can add that variable in AgnosticV, I don't see why should not be able to, but let us know what's your point of view here.

makentenza:
Already solved on a different PR

PR #2599, opened 2020-09-28
Add agnosticd_virtualenv role

jkupferer:
##### SUMMARY

Add role `agnosticd_virtualenv` with hook in `setup_runtime.yml` to create a python virtualenv for ansible execution on localhost if configured and not already present.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`setup_runtime.yml` playbook
`agnosticd_virtualenv` role

### DIFF
--- a/ansible/roles/agnosticd_virtualenv/.yamllint
+++ b/ansible/roles/agnosticd_virtualenv/.yamllint
@@ -1,0 +1,14 @@
+---
+extends: default
+
+rules:
+  comments:
+    require-starting-space: false
+  comments-indentation: disable
+  indentation:
+    level: warning
+    indent-sequences: whatever
+  line-length:
+    level: warning
+    max: 500
+    allow-non-breakable-inline-mappings: true
--- a/ansible/roles/agnosticd_virtualenv/.yamllint
+++ b/ansible/roles/agnosticd_virtualenv/.yamllint
@@ -7,8 +7,7 @@ rules:
   comments-indentation: disable
   indentation:
     level: warning
-    indent-sequences: whatever
+    indent-sequences: consistent
   line-length:
-    level: warning
-    max: 500
+    max: 120
     allow-non-breakable-inline-mappings: true
--- a/ansible/roles/agnosticd_virtualenv/README.adoc
+++ b/ansible/roles/agnosticd_virtualenv/README.adoc
@@ -7,7 +7,7 @@ This is role runs from setup_runtime with a default set of Python requirements t
 == Configuration
 
 `agnosticd_virtualenv_requirements` - Python requirements in list format as used by `pip` module `name` argument.
-Default value loaded from `configs/{{ env_type }}/requirements.txt` if it exists.
+Default value loaded from lines in `configs/{{ env_type }}/requirements.txt` if it exists.
 
 `agnosticd_virtualenv_base_dir` - Directory in which to create virtualenvs. Default `/tmp`.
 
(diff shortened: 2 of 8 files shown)


tonykay (approved):

ansible/roles/agnosticd_virtualenv/README.adoc
```diff
@@ -0,0 +1,18 @@
+= agnosticd_virtualenv
+
+Create a Python virtualenv for a host and set `ansible_python_interpreter` so that further task execution on the host will occur with this virtualenv.
+
+This is role runs from setup_runtime with a default set of Python requirements that can be set in a configs `requirements.txt`.
+
+== Configuration
+
+`agnosticd_virtualenv_requirements` - Python requirements in list format as used by `pip` module `name` argument.
+Default value loaded from lines in `configs/{{ env_type }}/requirements.txt` if it exists.
```
fridim, line 10: `requirements.yml` is for roles

Can we make that one more explicit ? maybe `python-requirements.txt` ?

ansible/roles/agnosticd_virtualenv/tasks/create.yml
```diff
@@ -0,0 +1,25 @@
...
+  pip:
+    name: "{{ agnosticd_virtualenv_requirements }}"
+    virtualenv: "{{ r_agnosticd_virtualenv_tmp.path }}"
+    virtualenv_python: "{{ agnosticd_virtualenv_python }}"
+
+# Ansible has no file move or rename module, so command will have to do
+- name: Move agnosticd virtualenv tmp to {{ agnosticd_virtualenv }}
+  command: mv {{ r_agnosticd_virtualenv_tmp.path }} {{ agnosticd_virtualenv }}
+  # Failed to mv virtualenv most likely means that another agnosticd run created it first
+  failed_when: false
```
fridim, line 20: Why do you create a temp directory and don't work directly in the target dir `agnosticd_virtualenv` ?

fridim (changes requested):

ansible/roles/agnosticd_virtualenv/.yamllint
```diff
@@ -0,0 +1,14 @@
+---
+extends: default
+
+rules:
+  comments:
+    require-starting-space: false
+  comments-indentation: disable
+  indentation:
+    level: warning
+    indent-sequences: whatever
```
fridim, line 10: `consistent` instead, no ?

ansible/roles/agnosticd_virtualenv/README.adoc
jkupferer: Makes sense, changed.

ansible/roles/agnosticd_virtualenv/tasks/create.yml
jkupferer: There is a race condition where two ansible runs on the same host that will use the same virtualenv start around the same time. Whichever one gets there second may see the virtualenv directory exists and proceed when it is actually still creating, so this is a safe design pattern to make sure the creation of the virtualenv is atomic and avoid errors if we do end up in this situation.

fridim: OK, that's what i guess,  but i thought it would be very unlikely.

> may see the virtualenv directory exists and proceed 

That would not be too much of a problem, only for that consecutive simultaneous run, the future runs would be fine.
On the other hand, I guess 2 simultaneous runs could build the virtualenv at the same time and could leave a broken virtualenv. That would be annoying.

I was not really sure if it was worth the additional tasks, but i guess now that it's written.. :)

9strands:
Cleaning old PR entries at end of 2022. Please reopen if this is still required.

PR #2613, opened 2020-09-29
Serverless: increase timeout for CSV waiting

guimou:
Several times already my deployment has failed because the Serverless CSV is not ready in a timely manner.
Proposed change is to increase retries from 15 to 30.


NonyNo3lle (approved):
LGTM

PR #2622, opened 2020-09-30
Add AWS infra support to SAP Integration config

makentenza:
##### SUMMARY

Extend sap-integration config so it can be deployed in AWS

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
sap-integration


NonyNo3lle:
@fridim can you check this one, please?

ansible/configs/sap-integration/default_vars.yml
```diff
@@ -100,15 +18,19 @@ project_tag: "{{ env_type }}-{{ guid }}"
 
 # Why is this config being deployed? Override for specific environments
 # Some valid: development, ilt, production, event
-purpose: development
+purpose: production
```
fridim, line 21: why change default purpose ?  development seems good to me as a default, and is then overridden in agnosticV

ansible/configs/sap-integration/software.yml
```diff
@@ -14,104 +14,110 @@
...
-        path: "{{ item }}"
-      loop:
-        - '/nfs/HANA_installation/'
-        - '/nfs/S4HANA_installation/'
-        - '/nfs/SAPCAR/'
-        - '/nfs/SAPHOSTAGENT/'
-      when: not rh_internal
+    - name: OSP NFS specific tasks
+      when: cloud_provider is match("osp")
+      block:
```
fridim, line 19: Please prefer `include_tasks` instead of `block` so the tasks are not evaluated or displayed in the logs in case the cloud provider is not osp

makentenza:
Thanks for reviewing @fridim. I've made the requested changes now.

fridim (approved):

PR #2655, opened 2020-10-01
Prevent failures while Satellite tasks are blocked

makentenza:
##### SUMMARY

While deploying multiple instances at the same time the deployment fails waiting fro Sat to unlock some tasks. Adding a retry option to the RHEL lock task to prevent this.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-hana


ansible/configs/sap-hana/pre_software.yml
```diff
@@ -17,7 +17,17 @@
...
   gather_facts: False
   tasks:
-    - command: subscription-manager release --set=8.1
+    - name: Lock RHEL release to SAP supported one
+      redhat_subscription:
+        state: present
+        server_hostname: "{{ set_repositories_satellite_url }}"
+        activationkey: "{{ set_repositories_satellite_activationkey }}"
+        org_id: "{{ set_repositories_satellite_org | default(satellite_org) }}"
+        release: 8.1
```
fridim, line 26: Should this be a variable ? it depend on the image used, right ?
(even if currently this lab only supports 8.1)

ansible/configs/sap-hana/pre_software.yml
```diff
@@ -17,7 +17,17 @@
...
+      register: lock_result
+      until: lock_result.rc == 0
```
fridim, line 28: return code ?

Since you're not using `command` anymore, i'd use this instead:

```yaml
until: lock_result is succeeded
```

fridim (approved):

PR #2688, opened 2020-10-01
yamllint formatting cleanup for role ocp4-workload-cost-uploader

ritesh97-rh:
##### SUMMARY
For the specified roles:

##### ISSUE TYPE
- Code Cleanup Request

##### COMPONENT NAME
ocp4-workload-cost-uploader

### DIFF
--- a/ansible/roles/ocp4-workload-cost-uploader/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-cost-uploader/tasks/workload.yml
@@ -80,7 +80,7 @@
     creates: "/home/ec2-user/korekuta-master/"
 
 - name: Run Korekuta Setup for this cluster
-  command: |
+  command: >-
     /home/ec2-user/korekuta-master/ocp_usage.sh --setup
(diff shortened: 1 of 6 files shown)


ansible/roles/ocp4-workload-cost-uploader/tasks/workload.yml
```diff
@@ -74,17 +74,17 @@
...
 - name: Unzip master.zip
   unarchive:
-    remote_src: yes
+    remote_src: true
     src: "/home/ec2-user/korekuta-master.zip"
     dest: "/home/ec2-user/"
     creates: "/home/ec2-user/korekuta-master/"
 
 - name: Run Korekuta Setup for this cluster
   command: |
```
jkupferer, line 83: This is kind of weird and surprising that this works! Normally line breaks in a command would be a problem and the `>-` would be required, but I guess that is not how the `command` module works... really though, this should not have line breaks in the command. Let's switch this.

jkupferer (changes requested):

jkupferer (approved):

PR #2699, opened 2020-10-02
yamllint formatting cleanup for role ocp4-workload-idm

ritesh97-rh:
##### SUMMARY
For the specified roles:

##### ISSUE TYPE
- Code Cleanup Request

##### COMPONENT NAME
ocp4-workload-idm

### DIFF
--- a/ansible/roles/ocp4-workload-idm/files/oauth-ldap.yaml
+++ b/ansible/roles/ocp4-workload-idm/files/oauth-ldap.yaml
@@ -1,10 +1,11 @@
+---
 apiVersion: config.openshift.io/v1
 kind: OAuth
 metadata:
   name: cluster
 spec:
   identityProviders:
-  - name: {{ idm_domain_name }}
+  - name: {{idm_domain_name}}
     mappingMethod: claim
     type: LDAP
     ldap:
--- a/ansible/roles/ocp4-workload-idm/.yamllint
+++ b/ansible/roles/ocp4-workload-idm/.yamllint
@@ -1,0 +1,13 @@
+---
+extends: default
+
+rules:
+  comments:
+    require-starting-space: false
+    min-spaces-from-content: 1
+  comments-indentation: disable
+  indentation:
+    indent-sequences: consistent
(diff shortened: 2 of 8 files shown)


ansible/roles/ocp4-workload-idm/files/oauth-ldap.yaml
```diff
@@ -1,10 +1,11 @@
+---
 apiVersion: config.openshift.io/v1
 kind: OAuth
 metadata:
   name: cluster
 spec:
   identityProviders:
-  - name: {{ idm_domain_name }}
+  - name: {{idm_domain_name}}
```
jkupferer, line 8: Why did we lose the spaces inside the brackets here?

jkupferer (changes requested):

ansible/roles/ocp4-workload-idm/files/oauth-ldap.yaml
tonykay: Yes I agree. As the opportunity arises it would be good to always insert spaces around the var names (and expand constructs like `d(foo)` to `default(foo)`, also spaces around `foo|bool` to `foo | bool`. All makes it more readable.

ritesh97-rh: I'm not sure .. yamllint was complaining about it. does it need to be in quotes (")?

ritesh97-rh: `./files/oauth-ldap.yaml
  8:13      error    too many spaces inside braces  (braces)
  8:29      error    too many spaces inside braces  (braces)`

That's what is says when I add the spaces back. Not sure why.

jkupferer: Ah! I see the issue here. This is a jinja template, not a normal YAML file. The file name should end in .j2

jkupferer (changes requested):

PR #2745, opened 2020-10-03
yamllint formatting cleanup for role ocp4-workload-quay-operator

ritesh97-rh:
##### SUMMARY
For the specified roles:

##### ISSUE TYPE
- Code Cleanup Request

##### COMPONENT NAME
ocp4-workload-quay-operator

### DIFF
--- a/ansible/roles/ocp4-workload-quay-operator/tasks/workload.yml
+++ b/ansible/roles/ocp4-workload-quay-operator/tasks/workload.yml
@@ -30,9 +30,12 @@
 
 - name: Generate Quay Superuser Password
   when:
-  - ocp4_workload_quay_operator.superuser_password is not defined or ocp4_workload_quay_operator.superuser_password|length == 0
+  - >-
+    ocp4_workload_quay_operator.superuser_password is not defined
+    or ocp4_workload_quay_operator.superuser_password | length == 0
   set_fact:
-    ocp4_workload_quay_operator_superuser_password: "{{ lookup('password', '/dev/null length=12 chars=ascii_letters') }}"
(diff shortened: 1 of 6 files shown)


ansible/roles/ocp4-workload-quay-operator/tasks/workload.yml
```diff
@@ -30,9 +30,12 @@
 
 - name: Generate Quay Superuser Password
   when:
-  - ocp4_workload_quay_operator.superuser_password is not defined or ocp4_workload_quay_operator.superuser_password|length == 0
+  - >-
```
stencell, line 33: This reads strange using the folded style block. How about something like:

```
when: ocp4_workload_quay_operator.superuser_password is not defined or
      ocp4_workload_quay_operator.superuser_password | length == 0
```

stencell (changes requested):
One small change requested, otherwise is good.

ansible/roles/ocp4-workload-quay-operator/tasks/workload.yml
ritesh97-rh: Removed the list notation `-`. But had to preserve the `>-` operator due to line length

stencell: Ok, I'll leave this since it looks better. Remember though, if you have a workload that just makes it hard to write these due to line length it is okay to change the yamllint file to be longer or just disable it. We don't want to use the letter of the rule to defeat the spirit of the intent.

stencell (approved):

PR #2790, opened 2020-10-06
Change sapa-hana, Fix dependencies of resources in sap-hana HOT

fridim:
This commit is in the same vein as #2789

Add missing dependencies in resources in the Heat Template


makentenza (approved):
LGTM

jkupferer (approved):

PR #2793, opened 2020-10-06
Change KNI config, fix osp_project_id in clouds.yml.j2 template

fridim:
The variable isn't defined by default, fallback to use the fact osp_project_info.


marcosmamorim (approved):
LGTM

PR #2843, opened 2020-10-07
yamllint formatting cleanup for ocp-workload-parksmap-demo

ritesh97-rh:
##### SUMMARY
For the specified roles:

##### ISSUE TYPE
- Code Cleanup Request

##### COMPONENT NAME
ocp-workload-parksmap-demo

### DIFF
--- a/ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
+++ b/ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
@@ -72,7 +72,6 @@ items:
           ports:
           - containerPort: 8080
             protocol: TCP
-       #  resources: {}
           terminationMessagePath: /dev/termination-log
           terminationMessagePolicy: File
           resources:
--- a/ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
+++ b/ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
@@ -1,3 +1,4 @@
+---
 kind: List
 apiVersion: v1
 metadata: {}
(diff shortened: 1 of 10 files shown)


ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
```diff
@@ -71,7 +72,7 @@ items:
           ports:
           - containerPort: 8080
             protocol: TCP
-          resources: {}
+       #  resources: {}
```
stencell, line 75: Why was this commented out compared to original?

stencell (changes requested):
see comment to address.

ansible/roles/ocp-workload-parksmap-demo/files/workshop-parksmap.yaml
ritesh97-rh: `resources` appears twice as as key in the same scope (open the whole file; don't think its visible in the snippet). The second one has a value, so I commented this one out .. should verify before merging.

stencell: These are different. Even though it is empty in both cases and shouldn't make a difference, we also should not be changing something like this - that is up to the workload owner.

`.spec.strategy.resources`
vs
`.spec.template.spec.containers.resources`

ritesh97-rh: Are you sure? There's another `resources: ` on line 77 that has limits and requests in it. I think that's the one yamllint says is conflicting.

stencell: I beg your pardon...you are absolutely right, I missed that one.

In this case, can you just the delete the one with the empty dict? That should resolve this.

stencell (approved):
lgtm

PR #2858, opened 2020-10-07
yamllint formatting cleanup for ocp-workload-starter-workshop

ritesh97-rh:
##### SUMMARY
For the specified roles:

##### ISSUE TYPE
- Code Cleanup Request

##### COMPONENT NAME
ocp-workload-starter-workshop


ansible/roles/ocp-workload-starter-workshop/tasks/per_user.yml
```diff
@@ -83,16 +83,20 @@
       tags: gogs
 
     - name: push catalog to git repository in Gogs for user "{{ my_user }}"
-      shell: |
-        git remote add {{ my_user }} http://{{ my_user }}:{{ user_gogs_password }}@{{ gogs_hostname }}/{{ my_user }}/{{ reponame_local }}.git
+      shell: >-
```
stencell, line 86: Why was this changed from literal to folded? It was not good before to be in a single `shell` task, but this will break it as is.

ansible/roles/ocp-workload-starter-workshop/tasks/remove_workload.yml
```diff
@@ -15,14 +15,20 @@
 - name: Delete user projects
   command: "oc delete project {{ item }}"
   ignore_errors: true
-  with_sequence: start={{user_count_start}} end={{ user_count_end }} format={{ user_format }}
+  with_sequence:
```
stencell, line 18: Not certain it will work formatted like this. It will need to be tested.

stencell (changes requested):
Testing and changes as noted.

ansible/roles/ocp-workload-starter-workshop/tasks/workload.yml
```diff
@@ -63,7 +63,10 @@
   tags: gogs
 
 - include_tasks: per_user.yml
-  with_sequence: start={{user_count_start}} end={{ user_count_end }} format={{ user_format }}
+  with_sequence:
```
stencell, line 66: This should not work.

ansible/roles/ocp-workload-starter-workshop/tasks/remove_workload.yml
```diff
@@ -15,14 +15,20 @@
...
+    end: "{{ user_count_end }}"
+    format: "{{ user_format }}"
   tags: always
 
 # Remove users quota
 - name: Delete user quota
   command: oc delete clusterquota/clusterquota-{{ item }}
   ignore_errors: true
-  with_sequence: start={{user_count_start}} end={{ user_count_end }} format={{ user_format }}
+  with_sequence:
```
stencell, line 28: This should not work.

ansible/roles/ocp-workload-starter-workshop/tasks/per_user.yml
ritesh97-rh: Line length constraints bc of the add url. Should not break .. double line in `>` converts to single new line. The yaml to json conversion looks the same. Could also separate the commands using `;` or `&&` as we would in shell.

stencell: but if you use folded, it will just string all of these lines together with a whitespace between them. These are all separate commands and, as you noted, this can be fixed with an `&&` option, but as of right now this would be broken.

ritesh97-rh: Double newline in folded = single newline in output. Single newline in `|` operator is the same as double newline in `>-` operator. That's why single newlines were converted to double newlines when the operator was changed from '|' to '>-`. It should not break.  The yaml to json conversion also looks the same.

stencell: Oh, you are right!

stencell (approved):
necessary changes made.

PR #2984, opened 2020-10-12
[kni-osp] permission fix and DNS setup issues fix

hgeaydem:


hgeaydem:
just spotted an issue while testing, will reopen when fixed

PR #2992, opened 2020-10-12
Cleanup role ocp4-workload-debezium-demo for yamllint

jkupferer:
##### SUMMARY

Style formatting changes to ocp4-workload-debezium-demo

Replaces previous yamllint PR closed due to conflict.

##### ISSUE TYPE

- Cleanup Pull Request

##### COMPONENT NAME

role ocp4-workload-debezium-demo

### DIFF
--- a/ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
+++ b/ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
@@ -39,23 +39,18 @@
     resource_definition: "{{ lookup('template', 'mssql-server-linux.yaml.j2') }}"
 
 - name: Wait for Database to Start
-  k8s_facts:
+  k8s_info:
     kind: Deployment
     name: mssql-server-linux
     namespace: '{{ e2e_namespace }}'
     api_version: apps/v1
-  register: mssql_resource
-  until: >-
-    (mssql_resource.resources is defined)
-    and (mssql_resource.resources|length>0)
-    and (mssql_resource.resources[0].status is defined)
-    and (mssql_resource.resources[0].status.readyReplicas is defined)
-    and (mssql_resource.resources[0].status.readyReplicas > 0)
+  register: r_k8s_info
+  until: r_k8s_info.resources[0].status.readyReplicas | default(0) > 0
   retries: 30
   delay: 10
 
 - name: Configure CDC in Database
-  command: ->
+  command: >-
     oc -n {{ e2e_namespace | quote }} exec deployment/mssql-server-linux --
     /opt/mssql-tools/bin/sqlcmd -S mssql-server-linux -U sa -P 'Password!' -i /opt/workshop/data.sql
 
@@ -75,18 +70,13 @@
     resource_definition: "{{ lookup('template', 'elasticsearch.yaml.j2') }}"
 
 - name: Wait for Kafka Cluster to start
-  k8s_facts:
+  k8s_info:
     kind: Kafka
     name: demo
     namespace: '{{ e2e_namespace }}'
     api_version: kafka.strimzi.io/v1beta1
-  register: kafka_resource
-  until: >-
-    (kafka_resource.resources is defined)
-    and (kafka_resource.resources|length>0)
-    and (kafka_resource.resources[0].status is defined)
-    and (kafka_resource.resources[0].status.conditions is defined)
-    and (kafka_resource.resources[0].status.conditions[0].type == 'Ready')
+  register: r_k8s_info
+  until: r_k8s_info.resources[0].status.conditions[0].type | default('') == 'Ready'
   retries: 30
   delay: 10
 
@@ -143,18 +133,13 @@
     part_of: 'replicated'
 
 - name: Wait for Database to Start
-  k8s_facts:
+  k8s_info:
     kind: Deployment
     name: postgresql
     namespace: '{{ e2e_namespace }}'
     api_version: apps/v1
-  register: pgsql_resource
-  until: >-
-    (pgsql_resource.resources is defined)
-    and (pgsql_resource.resources|length>0)
-    and (pgsql_resource.resources[0].status is defined)
-    and (pgsql_resource.resources[0].status.readyReplicas is defined)
-    and (pgsql_resource.resources[0].status.readyReplicas > 0)
+  register: r_k8s_info
+  until: r_k8s_info.resources[0].status.readyReplicas | default(0) > 0
   retries: 30
   delay: 10
 
--- a/ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
+++ b/ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
@@ -32,13 +32,13 @@
     namespace: '{{ e2e_namespace }}'
     resource_definition: "{{ lookup('template', 'configmap-data-sql.yaml.j2') }}"
 
-- name: Deploy MS SQL Server Database 
+- name: Deploy MS SQL Server Database
   k8s:
     state: present
     namespace: '{{ e2e_namespace }}'
     resource_definition: "{{ lookup('template', 'mssql-server-linux.yaml.j2') }}"
 
-- name: Wait for Database to Start 
+- name: Wait for Database to Start
   k8s_facts:
     kind: Deployment
     name: mssql-server-linux
@@ -55,12 +55,18 @@
   delay: 10
 
 - name: Configure CDC in Database
-  shell: "oc -n {{ e2e_namespace }} exec deployment/mssql-server-linux -- /opt/mssql-tools/bin/sqlcmd -S mssql-server-linux -U sa -P 'Password!' -i /opt/workshop/data.sql"
+  command: ->
+    oc -n {{ e2e_namespace | quote }} exec deployment/mssql-server-linux --
+    /opt/mssql-tools/bin/sqlcmd -S mssql-server-linux -U sa -P 'Password!' -i /opt/workshop/data.sql
 
 - name: Provision ECK Operator
   k8s:
     state: present
-    resource_definition: "{{ lookup('url', 'https://download.elastic.co/downloads/eck/{{ ocp4_workload_debezium_demo_eck_version }}/all-in-one.yaml', split_lines=False) }}"
+    resource_definition: >-
+      {{ lookup('url',
+        'https://download.elastic.co/downloads/eck/{{ ocp4_workload_debezium_demo_eck_version }}/all-in-one.yaml',
+        split_lines=False)
+      }}
 
 - name: Provision Elasticsearch
   k8s:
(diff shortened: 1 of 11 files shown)


ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
```diff
@@ -55,12 +55,18 @@
   delay: 10
 
 - name: Configure CDC in Database
-  shell: "oc -n {{ e2e_namespace }} exec deployment/mssql-server-linux -- /opt/mssql-tools/bin/sqlcmd -S mssql-server-linux -U sa -P 'Password!' -i /opt/workshop/data.sql"
+  command: ->
```
stencell, line 58: Should be `>-`

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
```diff
@@ -76,11 +82,11 @@
     api_version: kafka.strimzi.io/v1beta1
   register: kafka_resource
   until: >-
-    (kafka_resource.resources is defined) 
+    (kafka_resource.resources is defined)
```
stencell, line 85: Can this just be turned into a list which will be a logical and? This is okay, but I think the list is more readable.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_mysql_demo.yaml
```diff
@@ -40,11 +40,11 @@
     api_version: kafka.strimzi.io/v1beta1
   register: kafka_resource
   until: >-
-    (kafka_resource.resources is defined) 
+    (kafka_resource.resources is defined)
```
stencell, line 43: same, turn into a list (if you decide to).

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_registry.yaml
```diff
@@ -43,11 +44,11 @@
     api_version: kafka.strimzi.io/v1beta1
   register: kafka_resource
   until: >-
-    (kafka_resource.resources is defined) 
+    (kafka_resource.resources is defined)
```
stencell, line 47: same, turn into a list (if you decide to).

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_webapp.yaml
```diff
@@ -17,20 +17,24 @@
 
 - name: Download example files
   unarchive:
-      src: '{{ocp4_workload_debezium_demo_webapp_operator_resources}}'
-      dest: '{{ webapp_operator_tmp }}'
-      remote_src: yes
+    src: '{{ocp4_workload_debezium_demo_webapp_operator_resources}}'
```
stencell, line 20: missing space after `{{`

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_webapp.yaml
```diff
@@ -45,8 +49,8 @@
     api_version: route.openshift.io/v1
   register: webapp_secure_route
   until: >-
-    (webapp_secure_route.resources is defined) 
-    and (webapp_secure_route.resources|length>0) 
+    (webapp_secure_route.resources is defined)
```
stencell, line 52: turn this into a list to make it more readable and consistent

stencell (changes requested):
few comments to address

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
jkupferer: Ouch! Good catch!

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
jkupferer: Oh, I think I can do much better than that... I was trying to make minimal changes, but now that you mention...

jkupferer:
@stencell Can I get another look at this one?

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
```diff
@@ -32,35 +32,36 @@
...
     api_version: apps/v1
-  register: mssql_resource
-  until: >-
-    (mssql_resource.resources is defined)
-    and (mssql_resource.resources|length>0)
-    and (mssql_resource.resources[0].status is defined)
-    and (mssql_resource.resources[0].status.readyReplicas is defined)
-    and (mssql_resource.resources[0].status.readyReplicas > 0)
+  register: r_k8s_info
+  until: r_k8s_info.resources[0].status.readyReplicas | default(0) > 0
```
stencell, line 48: The main concern I have with shrinking this so much is when Ansible checks on the object and one of them doesn't exist yet. So, if it checks for the existence of the `Deployment` and it doesn't exist yet, it will fail. If the Deployment exists, but the `readyReplicas` field doesn't, it will fail. This is not to be confused with the `readyReplicas` being equal to zero. It is why we defined multiple levels to check as the objects are rolling out.

Probably less of a problem with Deployments vs something like a CSV...but this can still cause problems if the cluster is a bit slow.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_mysql_demo.yaml
```diff
@@ -33,18 +33,13 @@
...
     api_version: kafka.strimzi.io/v1beta1
-  register: kafka_resource
-  until: >-
-    (kafka_resource.resources is defined) 
-    and (kafka_resource.resources|length>0)
-    and (kafka_resource.resources[0].status is defined) 
-    and (kafka_resource.resources[0].status.conditions is defined)
-    and (kafka_resource.resources[0].status.conditions[0].type == 'Ready') 
+  register: r_k8s_info
+  until: r_k8s_info.resources[0].status.conditions[0].type | default('') == 'Ready'
```
stencell, line 42: Same as previous comment.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_registry.yaml
```diff
@@ -36,18 +37,13 @@
...
     namespace: '{{ ocp4_workload_debezium_demo_registry_namespace }}'
     api_version: kafka.strimzi.io/v1beta1
-  register: kafka_resource
-  until: >-
-    (kafka_resource.resources is defined) 
-    and (kafka_resource.resources|length>0)
-    and (kafka_resource.resources[0].status is defined) 
-    and (kafka_resource.resources[0].status.conditions is defined)
-    and (kafka_resource.resources[0].status.conditions[0].type == 'Ready') 
+  register: k8s_info
```
stencell, line 45: If these vars are going to be renamed, can we be consistent? In the earlier task you are using `r_k8s_info` and here `k8s_info`. I think if you register these, they should be a little more specific to what is being checked like `r_crd_apicurio` or something in this case.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
wkulhanek: All this *and* stuff isn't even necessary. Just make it a list of *when* conditions in the order to be checked.

fridim: starting at ansible 2.8  that shortcut works.

```yaml
- hosts: localhost
  gather_facts: false
  vars:
    a:
      b:
        c:
          d: ok
  tasks:
    - debug:
        msg: "ok"
      when: a.c.e.f.g|default(0) > 0
```

with 2.7.18, error

```TASK [debug] **************************************************************************************************************************************
Tuesday 24 November 2020  15:26:12 +0100 (0:00:00.088)       0:00:00.088 ****** 
fatal: [localhost]: FAILED! => 
  msg: |-
    The conditional check 'a.c.e.f.g|default(0) > 0' failed. The error was: error while evaluating conditional (a.c.e.f.g|default(0) > 0): 'dict object' has no attribute 'c'
  
    The error appears to have been in '/tmp/test.yaml': line 9, column 7, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
      tasks:
        - debug:
          ^ here
```

with 2.8.0, it works.

I think we use at least ansible 2.9 for all our deployments, right ?

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_registry.yaml
fridim: +1

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_webapp.yaml
```diff
@@ -17,42 +17,44 @@
...
   k8s:
     state: present
     namespace: '{{ webapp_namespace }}'
-    src: '{{ webapp_operator_tmp }}/tutorial-web-app-operator-{{ ocp4_workload_debezium_demo_webapp_operator_release_tag }}/deploy/{{ item }}'
-  with_items: "{{ ocp4_workload_debezium_demo_webapp_operator_resource_items }}"
+    src: >-
+      {{ '{}/tutorial-web-app-operator-{}/deploy/{}'.format(
+        webapp_operator_tmp, ocp4_workload_debezium_demo_webapp_operator_release_tag, item
+      ) }}
+  loop: "{{ ocp4_workload_debezium_demo_webapp_operator_resource_items }}"
```
fridim, line 32: i don't think using `.format()` is more clear here.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_registry.yaml
jkupferer: Fixed to use the `r_k8s_info` consistently. I think using a generic variable name is appropriate here because it highlights that this registered variable is not used for anything else beyond this task. It helps underscore that it is just used here and done with it.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_webapp.yaml
jkupferer: Fair enough. It weighs in at 137 characters with just inline. Which seems not unreasonable. I'm looking to suggest 150 character line length cutoff in the future.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_e2e_demo.yaml
jkupferer: Right. We are targeting ansible 2.9, so we can just shortcut past all these checks. It was only older versions where `default()` wouldn't catch the error on undefined.

ansible/roles/ocp4-workload-debezium-demo/tasks/provision_mysql_demo.yaml
jkupferer: Same story here. It isn't a problem with ansible 2.9 and later.

stencell (approved):

PR #3022, opened 2020-10-12
Updating playbook to prepull image for student

christianh814:
##### SUMMARY

Adding task that "prepulls" the Windows base image. This helps by speeding up application deployment since the image is 5GB in size.

This is done by creating a Job that does an `ssh` connection to the Windows node and runs a `docker pull` of the image. This job is run in the same namespace as the operator, so that it may use the same secret needed to ssh into the Windows node.

Right now this is a "best effort" and I only check if the job gets created. The playbook should still succeed even if the job fails. I've accounted for this in my documentation to the student.

##### ISSUE TYPE

- New config Pull Request

##### COMPONENT NAME

- Adding a `job` task.


christianh814:
@wkulhanek PTAL

wkulhanek (approved):

PR #3023, opened 2020-10-13
[WIP] Change ansible-tower, Add dynamic workload support

fridim:
Add dynamic workloads to ansible-tower config.

The workload is a list of dict and thus the target groups can be specified.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
ansible-tower config


tonykay (approved):

tonykay:
Assume we would do "tower" virtualenvs etc the same way in the future

tonykay:
Actually I think the tasks should move into one of the existing standard plays. `software.yml` at some point? This is "software installation and configuration"?

fridim:
> Actually I think the tasks should move into one of the existing standard plays. `software.yml` at some point? This is "software installation and configuration"?

Yes, i'm looking at integrating that into agnosticd core

PR #3061, opened 2020-10-19
Add retry to tasks using mirror.openshift.com

fridim:
##### ISSUE TYPE

- Bugfix Pull Request


NonyNo3lle (approved):
Approving because GC

PR #3070, opened 2020-10-20
Add role ocp4_workload_image_registry

jkupferer:
##### SUMMARY

Add role ocp4_workload_image_registry

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role ocp4_workload_image_registry

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_image_registry/.travis.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_image_registry/.travis.yml
@@ -1,0 +1,29 @@
+---
+language: python
+python: "2.7"
+
+# Use the new container infrastructure
+sudo: false
+
+# Install ansible
+addons:
+  apt:
+    packages:
+    - python-pip
+
+install:
+# Install ansible
+- pip install ansible
+
+# Check ansible version
+- ansible --version
+
+# Create ansible.cfg with correct roles_path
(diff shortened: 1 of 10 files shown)


wkulhanek:
@jkupferer Do we actually need that role? ocp4_workload_opentlc_production already has that capability. You could just use that role and set the other customizations off if a registry route is all you need.

jkupferer:
@wkulhanek It seems to me that `ocp4_workload_opentlc_production` does too much. Following @tonykay 's principle of Roles should be targeted to specific functions?

jkupferer:
Also, a role that does "opentlc_production" stuff just doesn't seem clearly defined. When I saw this role existed I guessed that it was specific to maintaining our OpenTLC shared clusters, that it wasn't meant to be used in other ways?

wkulhanek:
It might be oddly named (probably better would be "cluster settings" or something. But that doesn't take away from the fact that this role would duplicate existing functionality. Which isn't necessary. I really don't subscribe to the two task "let's add more overhead" roles.

jkupferer:
Cluster settings is also too vague of a role concept. A single role should not be configuring the registry and also self-provisioner access. Everything in OCP4 is a cluster setting? The production role seems like a miscellaneous grab-bag?

As @NonyNo3lle said in chat, I think we may need a tie-breaker vote.

NonyNo3lle:
@fridim your feedback would be good here too

PR #3071, opened 2020-10-20
Fix role ocp4_workload_dso registry route detection

jkupferer:
##### SUMMARY

Minor fix-up for ocp4_workload_dso registry route detection. This adds a failure condition to the step when getting the route to make failure more obvious.

Also remove step to expose the registry as this task will have failed here if the registry were not already exposed.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Role ocp4_workload_dso


NonyNo3lle (approved):
LGTM

PR #3072, opened 2020-10-20
Change workload ocp4-workload-quarkus-workshop expose https route for get-a-username

edeandrea:
##### SUMMARY

The route that gets exposed for the `get-a-username` application only exposes an http route. This change also exposes an https route.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`install-username-distribution` task within `ocp4-workload-quarkus-workshop`


edeandrea:
@jamesfalkner / @danieloh30 would either of you like to review?

jamesfalkner:
Nice! I also think this could/should be used in other workshops that use this - DG, CCN, etc.

edeandrea:
The main reason for the change is that I like to use the red.ht URL shortener when i build workshops. The red.ht URL shortener only works with https:// URLs.

danieloh30:
+1, agree with James!

@jamesfalkner on the side note, how is going to maintain this by Hugo and Evan? Should we still maintain this separately?

danieloh30:
> The main reason for the change is that I like to use the red.ht URL shortener when i build workshops. The red.ht URL shortener only works with https:// URLs.

Right. I remembered you raised this issue in the past. Sorry for addressing it earlier :D

jamesfalkner:
> +1, agree with James!
> 
> @jamesfalkner on the side note, how is going to maintain this by Hugo and Evan? Should we still maintain this separately?

This only affects the Quarkus workshop - they could also adopt the same mechanism when deploying get-a-username if they wanted, but this won't break anything for them.

edeandrea:
If you want @jamesfalkner / @danieloh30 I can do a similar PR for the DG8 & CCN roadshow workshops.

danieloh30:
> > +1, agree with James!
> > @jamesfalkner on the side note, how is going to maintain this by Hugo and Evan? Should we still maintain this separately?
> 
> This only affects the Quarkus workshop - they could also adopt the same mechanism when deploying get-a-username if they wanted, but this won't break anything for them.

Makes sense!

jkupferer (approved):

PR #3076, opened 2020-10-20
Add new required variable

makentenza:
##### SUMMARY

Config is failing now as a new variable `ocp4_network_ovn_install_workaround` is required. Adding this default vars.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-integration


wkulhanek:
@makentenza Actually..... would you mind fixing the root cause instead (that's my bad, I thought I had done it right....):

In Role host-ocp4-installer / tasks / main.yml, Line 13 add `| default(false)` before the bool. Then you don't need to add this.

wkulhanek:
@makentenza Actually never mind. Submitting a new PR to fix this at the root.

makentenza:
Thanks @wkulhanek closing this then

PR #3155, opened 2020-11-03
Change role infra-ec2-template-generate to add aws_dns_create_public_zone

jkupferer:
##### SUMMARY

This PR implements functionality for `ocp4-cluster` to match the `ocp4-workshop` flag `use_root_zone`. This is a feature we need for GPTE infrastructure clusters in order to shorten the DNS domain names.

This also makes some whitespace formatting changes to the jinja2 template to make control flow statements easier to see and visually match by moving the opening bracket `{%` to the beginning of the line.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role infra-ec2-template-generate

##### ADDITIONAL INFORMATION

Test deploys with ocp4-cluster performed both with and without this flag to confirm new functionality and that this feature does not break anything.

### DIFF
--- a/ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
+++ b/ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
@@ -123,7 +123,9 @@ Resources:
       HostedZoneConfig:
         Comment: "{{ aws_comment }}"
 
-  {% if secondary_stack is not defined %}
+{% if secondary_stack is not defined
+   and aws_dns_create_public_zone | bool
+%}
   DnsZonePublic:
     Type: "AWS::Route53::HostedZone"
     Properties:
@@ -136,11 +138,11 @@ Resources:
     DependsOn:
       - DnsZonePublic
     Properties:
-    {% if HostedZoneId is defined %}
+{%   if HostedZoneId is defined %}
       HostedZoneId: "{{ HostedZoneId }}"
-    {% else %}
+{%   else %}
       HostedZoneName: "{{ aws_dns_zone_root }}"
-    {% endif %}
+{%   endif %}
       RecordSets:
         - Name: "{{ aws_dns_zone_public }}"
           Type: NS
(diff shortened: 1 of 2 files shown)


ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
```diff
@@ -149,133 +151,135 @@ Resources:
...
+{%     if not aws_dns_create_public_zone | bool %}
+      HostedZoneName: "{{ aws_dns_zone_root }}"
+{%     elif secondary_stack is defined %}
       HostedZoneName: "{{ aws_dns_zone_public }}"
-      {% else %}
+{%     else %}
       HostedZoneId:
         Ref: DnsZonePublic
-      {% endif %}
+{%     endif %}
```
fridim, line 176: i'm not sure to understand the logic here.
What happens if `secondary_stack` is true and `aws_dns_create_public_zone` is true ?

It would try to use the ref DnsZonePublic which should not be present in the current stack (as it is secondary).

I think you want:

```jinja2
{%     if aws_dns_create_public_zone | bool %}
{%       if secondary_stack is defined %}
      HostedZoneName: "{{ aws_dns_zone_public }}"
{%       else %}
      HostedZoneId:
        Ref: DnsZonePublic
{%       endif %}
{%     else %}
{%       if secondary_stack is defined %}
      HostedZoneName: "{{ aws_dns_zone_public }}"
{%       else %}
      HostedZoneName: "{{ aws_dns_zone_root }}"
{%     endif %}
```

ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
```diff
@@ -289,29 +293,31 @@ Resources:
...
+{%       if not aws_dns_create_public_zone | bool %}
+      HostedZoneName: "{{ aws_dns_zone_root }}"
+{%       elif secondary_stack is defined %}
       HostedZoneName: "{{ aws_dns_zone_public }}"
-      {% else %}
+{%       else %}
       HostedZoneId:
         Ref: DnsZonePublic
-      {% endif %}
+{%       endif %}
```
fridim, line 303: same here

ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
jkupferer: Good point. I changed it to, which I believe is logically equivalent to what you suggested:
```
{%     if secondary_stack is defined %}
      HostedZoneName: "{{ aws_dns_zone_public }}"
{%     elif not aws_dns_create_public_zone | bool %}
      HostedZoneName: "{{ aws_dns_zone_root }}"
{%     else %}
      HostedZoneId:
        Ref: DnsZonePublic
{%     endif %}
```

ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
jkupferer: also changed here to give `secondary_stack` more weight than `aws_dns_create_public_zone`

ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
fridim: Not quite.

In your code snippet, if `secondary_stack` is true and `aws_dns_create_public_zone` is false, your template will still use

`aws_dns_zone_public` and not `aws_dns_zone_root`

You want both conditions nested IMO.

fridim (approved):

PR #3188, opened 2020-11-13
[Bug Fix] Adding log to see if updates correctly

weimeilin79:
##### SUMMARY

[Bug Fix] Adding log to see if updates correctly

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


stencell:
I don't think this is needed...the one merge that seems to be a problem is the `provision_shared` tasks. #3189 will merge just those changes as everything else appears to be good.

PR #3193, opened 2020-11-13
Add role host-ocp4-installer ocp4_scsi_device_detection_workaround

jkupferer:
##### SUMMARY

Add work around for OpenShift cinder device id truncation issue documented at https://bugzilla.redhat.com/show_bug.cgi?id=1897603

This PR adds a MachineConfig that configures udev to create symbolic links matching what OpenShift is looking for, truncated at 20 characters.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role host-ocp4-installer


ansible/roles/host-ocp4-installer/tasks/main.yml
```diff
@@ -4,26 +4,35 @@
 - name: Generate install_config.yaml
   import_tasks: generate_install_config.yml
 
+- name: Run openshift-install create manifests
```
stencell, line 7: Shouldn't we have a conditional here? If we're not doing either of the workarounds, we don't need to do this.

jkupferer: Yeah, but it hardly matters. Not running this step complicates the code for very little difference in execution time.

stencell (approved):

PR #3210, opened 2020-11-18
set-repositories role update

wilson-walrus:
##### SUMMARY

When registering a system to rhn directly we could not specify which systems were RHEL 7 or RHEL 8. This change preserves the original function of the role, but adds the option to exclusively define rhel_repos_el7 or rhel_repos_el8 in the case that a developer requires two systems running different major versions of RHEL.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

role, set-repositories

##### ADDITIONAL INFORMATION

Example use:

rhel_repos:
 - not defined, is exclusive to _el7 and _el8, preserving original functionality

rhel_repos_el7:
  - rhel-7-server-rpms
  - rhel-7-server-rh-common-rpms
  - rhel-7-server-extras-rpms

rhel_repos_el8:                           
  - rhel-8-for-x86_64-baseos-rpms
  - rhel-8-for-x86_64-appstream-rpms

```paste below

```

### DIFF
--- a/ansible/roles/set-repositories/tasks/pre_checks_rhn.yml
+++ b/ansible/roles/set-repositories/tasks/pre_checks_rhn.yml
@@ -35,3 +35,34 @@
         - rhsm_pool_ids is defined
         - rhel_subscription_user is not defined
         - rhel_subscription_pass is not defined
+
+- name: Ensure rhel_repo dictionary is exclusively defined
+  when:
+    - rhel_repos is defined
+  assert:
+    that: "{{ check.that }}"
+    fail_msg: "{{ check.msg }}"
+    quiet: true
+  loop_control:
+    loop_var: check
+    label: "{{ check.msg }}"
+  loop:
+    - msg: Ensure no exclusive variables are set.
+      that:
+        - rhel_repos_el7 is not defined
+        - rhel_repos_el8 is not defined
+
+- name: Ensure el7 and el8 dictionaries are exclusively defined
+  when:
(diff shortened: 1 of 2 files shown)


ansible/roles/set-repositories/tasks/pre_checks_rhn.yml
```diff
@@ -35,3 +35,34 @@
...
+    that: "{{ check.that }}"
+    fail_msg: "{{ check.msg }}"
+    quiet: true
+  loop_control:
+    loop_var: check
+    label: "{{ check.msg }}"
+  loop:
+    - msg: Ensure no exclusive variables are set.
+      that:
+        - rhel_repos is not defined
```
jkupferer, line 68: What's going on with these loops with only one loop item? Why not combine all these `assert` tasks with loops with one item into a single assertion?

ansible/roles/set-repositories/tasks/rhn-repos.yml
```diff
@@ -35,9 +35,40 @@
...
+- name: Enable repos with rhsm_repository
+  when: 
+    - rhel_repos is defined 
+  rhsm_repository:
+    name: "{{ __rhel_repos }}"
+    state: enabled
+  loop:
+    - '{{ rhel_repos }}'
+  loop_control:
+    loop_var: __rhel_repos
```
jkupferer, line 48: Why all these loops with only one item?

jkupferer (changes requested):

jkupferer (approved):

PR #3223, opened 2020-11-23
[Don't Merge] custom install-config template

miteshget:
##### SUMMARY
Custom install-config.yaml.j2 template added for IPI installation 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
config: ocs4-external-implementation


bosebc (approved):
Looks good. Thanks

ansible/configs/ocs4-external-implementation/sample_vars/osp_vars.yml
```diff
@@ -25,8 +25,9 @@ platform: labs
 purpose: development
 
 # OpenShift Installer version to use
-#ocp4_installer_version: 4.5.3
-#ocp4_installer_version: 4.6.1
+# ocp4_installer_version: 4.5.3
```
jkupferer, line 28: Why include this commented variable?

ansible/configs/ocs4-external-implementation/pre_software.yml
```diff
@@ -32,6 +32,13 @@
 - name: Step 003.1 - Configuring Bastion Hosts
   hosts: bastions
   become: true
+  pre_tasks:
+    - name: set ocp network subnetid
+      set_fact:
+        ocp_network_subnet_id: "{{ hostvars['localhost']['ocp_network_subnet_id'] }}" 
+
+    - debug: var=ocp_network_subnet_id
+
```
jkupferer, line 41: Please use consistent YAML indentation. No other YAML list has extra indent in this file.

jkupferer (changes requested):

wkulhanek:
My bigger question is: why is this a separate config and not ocp4-cluster? What's so different here? @miteshget

ansible/configs/ocs4-external-implementation/post_infra.yml
```diff
@@ -21,3 +21,11 @@
     - name: OpenShift Floating IPs on OpenStack 
       include_role:
         name: ocp-infra-osp-fip
+    
+    - name: find ocp network subnet id
+      shell: openstack subnet list -f csv |grep {{ guid }}-ocp-subnet |cut -f1 -d,
```
jkupferer, line 26: Rather than use `-f csv` with `cut`, perhaps use `-f json` and use filter `from_json`?

miteshget:
> My bigger question is: why is this a separate config and not ocp4-cluster? What's so different here? @miteshget

Hi @wkulhanek, Thank you for Review. 
The default install-config creates following network  infra as in Figure-1
top1_ocp

We wanted to have network as in Figure-2
top2_ocp

for that I need to include following parameters in install-config.yaml.j2 so I thought to keep within config instead updating role-

  openstack:
       machinesSubnet: {{ ocp_network_subnet_id }}

ansible/configs/ocs4-external-implementation/post_infra.yml
miteshget: Thank you @jkupferer, I will check how I can use that instead cut.

miteshget:
@wkulhanek Please let me know if we can get the result other way.

ritzshah:
> > My bigger question is: why is this a separate config and not ocp4-cluster? What's so different here? @miteshget
> 
> Hi @wkulhanek, Thank you for Review.
> The default install-config creates following network infra as in Figure-1
> top1_ocp
> 
> We wanted to have network as in Figure-2
> top2_ocp
> 
> for that I need to include following parameters in install-config.yaml.j2 so I thought to keep within config instead updating role-
> 
> openstack:
> machinesSubnet: {{ ocp_network_subnet_id }}

@wkulhanek - We wanted to have 3 ceph nodes (external ceph) acting as part of OCS External Storage connected to worker nodes. This means that the worker and the ceph nodes to be talking over network in both direction which was not possible in the default installer configuration file is what I understand. Correct me if there is a better way to do this pls.

ansible/configs/ocs4-external-implementation/sample_vars/osp_vars.yml
miteshget: I kept the commented variables as it is as it was in original config. Should I remove that ?

miteshget:
We are not working on it any more

PR #3234, opened 2020-11-24
Bitwarden PoC

fridim:
##### SUMMARY

This change adds:
* `bitwarden.py` ansible module to communicate and load secrets from bitwarden secret notes
* `agnosticd_bitwarden` role that uses the module

Please see example in the Readme of the doc.

##### ISSUE TYPE

- Feature Pull Request
- New role Pull Request

##### COMPONENT NAME

Bitwarden Role and module

### DIFF
--- a/ansible/include_vars.yml
+++ b/ansible/include_vars.yml
@@ -61,3 +61,8 @@
           item.stat.path
           | default('skipped')
         ) if 'stat' in item else item }}
+
+  - name: Load secret vars from Bitwarden
+    when: agnosticd_bitwarden is defined
+    include_role:
+      name: agnosticd_bitwarden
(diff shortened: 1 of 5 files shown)


jkupferer (commented):
A few thoughts...

I think I am missing the big picture? How do we plan to manage `agnosticd_bitwarden_username` and `agnosticd_bitwarden_password` values? Will each catalog item or account have bitwarden credentials? If these are common secrets then this will be a significant weakness in our security... no worse than our current situation on the admin host, but we were trying to tighten secret security in babylon.

I believe we only would want to have the bitwarden command executed from localhost. Which makes me wonder if perhaps this should be implemented as an action plugin rather than a module? Or perhaps bitwarden should be implemented as a lookup plugin? As a lookup plugin I believe it would be much simpler to reference values from configs.

marcosmamorim:
We try to find a solution for the username and password, encrypting the username and password when connecting, but we still need the secret file stored somewhere, just like we do today.

I'm creating a new way using a bitwarden organization passkey, but the API documentation is not very good, so it's not as simple as using the existing client.

PR #3236, opened 2020-11-24
Update OCP 4 HA Lab Config for RHEL 8.2 and dynamic dns

wkulhanek:
##### SUMMARY

The OpenStack HA Config needed an update to run on the Orange (and Pink etc.) OpenStack clusters.
I also took the opportunity to update the bastion and utilityvms to RHEL 8.2 and Satellite.

Nameservers had to be set to override the ones coming from OpenStack to get acceptable performance.

Validated that the solver still works (except for registry PVC at the end which is no longer necessary).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4-disconnected-osp-lab

### DIFF
--- a/ansible/configs/ocp4-disconnected-osp-lab/default_vars.yml
+++ b/ansible/configs/ocp4-disconnected-osp-lab/default_vars.yml
@@ -92,7 +92,7 @@ install_k8s_modules: false
 # This var will set the version of ftl-injector to use.
 # In addition, it implies that FTL should be installed, eventually
 # removing the need for install_ftl var.
-ftl_injector_tag: v0.16.0
+ftl_injector_tag: v0.17.1
 
 # FTL injector will try to install python-pip and we only have python3-pip available
 # This var will force the ftl-injector role to adapt accordingly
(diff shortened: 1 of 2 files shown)


ansible/configs/ocp4-disconnected-osp-lab/default_vars.yml
```diff
@@ -231,34 +232,36 @@ query_subnet_cidr: "[?name=='{{ ocp_network }}'].subnet_cidr"
...
-    allocation_end: 192.168.47.254
-    dns_nameservers: []
-    create_router: true
+- name: ocp
+  shared: "false"
+  subnet_cidr: 192.168.47.0/24
+  gateway_ip: 192.168.47.1
+  allocation_start: 192.168.47.10
+  allocation_end: 192.168.47.254
+  dns_nameservers:
```
stencell, line 241: I know why you added this based on the chat earlier, but I don't understand why we need it. We should just let the OpenStack DNS servers do this resolution for us. It is how we have been doing it and it has worked fine - not sure what broke. 

I seem to remember not doing this because it broke name resolution inside the tenant. Like you were no longer able to resolve `example.com` because Google DNS has NFC what that is.

stencell (changes requested):
One thing to address...I won't die on my "this isn't right" sword, I just want to make sure it still works if we hard code ourselves to rely on external Internet dns servers.

stencell:
Also this is the OCP4 disconnected config that you changed...the PR says it is the OCP4 HA lab. Not sure if you crossed wires or intended to push more.

wkulhanek:
It’s the config for the ILT. 

I’m fine either way with the name servers. It seems to install ocp just fine with them set. But we can leave the default empty amd set in agnosticv until osp is fixed.

stencell:
Oh, well my main concern was things being able to resolve `utilityvm.example.com`. If you installed OCP in disconnected ILT then I reckon that works. I missed where you said you ran the solver.

stencell (approved):

stencell:
Tag `ocp4-disconnected-osp-lab-prod-0.34` pushed as well.

PR #3245, opened 2020-12-01
Change action_plugin user_info: create output_dir if not existent

guimou:
##### SUMMARY

When testing ocp_workloads independently (like on an existing OCP cluster, from localhost), the user_info script will fail if the output_dir does not exist.
This PR adds basic directory check and creation to the script. This does not change behaviour in a full deployment when the directory has already been created, but prevents failing on workload test.
Added line example:
`os.makedirs(os.path.dirname(os.path.join(output_dir, 'user-info.yaml')), exist_ok=True)`

### DIFF
--- a/ansible/action_plugins/agnosticd_user_info.py
+++ b/ansible/action_plugins/agnosticd_user_info.py
@@ -88,19 +88,19 @@ class ActionModule(ActionBase):
                 )
             )
             if not user and msg != None:
-                os.makedirs(os.path.dirname(os.path.join(output_dir, 'user-info.yaml')), exist_ok=True)
+                os.makedirs(output_dir, exist_ok=True)
                 fh = open(os.path.join(output_dir, 'user-info.yaml'), 'a')
                 fh.write('- ' + json.dumps(msg) + "\n")
                 fh.close()
             if not user and body != None:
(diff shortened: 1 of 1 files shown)


ansible/action_plugins/agnosticd_user_info.py
```diff
@@ -88,16 +88,19 @@ def run(self, tmp=None, task_vars=None):
                 )
             )
             if not user and msg != None:
+                os.makedirs(os.path.dirname(os.path.join(output_dir, 'user-info.yaml')), exist_ok=True)
```
jkupferer, line 91: Why join then strip off dir with dirname? why not just?
```
                os.makedirs(output_dir, exist_ok=True)
```

jkupferer (changes requested):

ansible/action_plugins/agnosticd_user_info.py
guimou: Because I'm lazy and totally dumb, so I just copy/pasted the code for the path that did not work from the other line. 😔
I will correct this.

guimou: Change made, thanks for spotting this Johnathan!

jkupferer (approved):

PR #3249, opened 2020-12-02
machinesSubnet parameter added

miteshget:
##### SUMMARY
For flat network machinesSubnet: {{ ocp_network_subnet_id }} parameter added with condition in install-config.yaml.j2 template

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
role: host-ocp4-installer


wkulhanek (approved):

wkulhanek:
I would suggest adding a comment to the default_vars_osp.yml file as well. Leave it commented out but have documentation there...

jkupferer (commented):
How will we use this? How would someone know how to set the `ocp_network_subnet_id`?

miteshget:
@jkupferer @wkulhanek `ocp_network_subnet_id` needs UUID of subnet. So in our case we have added a task in post_infra.yml which will fetch subnet id and store that value in the variable. E.g

```    - name: find ocp network subnet id
      command: openstack subnet list -f json
      register: __subnet_ids
    - name: set ocp network subnetid
      set_fact:
        ocp_network_subnet_id: "{{ item.Network }}" 
      when: 'item.Name == guid + "-ocp-subnet"'
      loop: "{{ __subnet_ids.stdout }}"```

**Comment by wkulhanek** [comment]
@miteshget That's not a good solution. Because if you only have this in post-infra of that one configs this will break in other configs. I'd suggest moving this to the install role with a block conditional.

**Review by jkupferer — CHANGES_REQUESTED** [review]
We need to figure out how to set `machinesSubnet` based on the uuid created in the heat template. The parameter used to configure this should be named something that is easy to understand, something like `ocp_use_single_network` or something of that sort.

**Comment by miteshget** [comment]
Hi @wkulhanek and @jkupferer,

I have added task in tasks/osp_pre.yml to find subnet id which will run when `ocp_use_single_network` variable as boolean set.

Hope this time it fulfills all the requirements.

Please review and merge or comment for improvements.

**Comment by miteshget** [comment]
JFI- I have run successful test just by using following variables -

osp_new_infra
`ocp_use_single_network: true`
`ocp4_installer_version: 4.6.1`

attach is network topology after deployment

**Comment by prakhar1985** [comment]
@miteshget looks good. I think moving it to the role and not adding anything in post_software is a good recommendation.

**Review by prakhar1985 — APPROVED** [review]

**Comment by miteshget** [comment]
Hello @wkulhanek and @jkupferer, could you please spare some time to review this PR again?

File: ansible/roles/host-ocp4-installer/tasks/main.yml [hunk 1]
```diff
@@ -1,6 +1,11 @@
 - name: Install client and OpenShift Installer binaries
   import_tasks: install_installer.yml
 
+## Open the port for the api
+- name: OpenStack specific requirement - find network port id
+  when: cloud_provider == "osp"
```
wkulhanek, line 6: I think we may want another when here to only do this when ocp_use_single_network is true

wkulhanek (changes requested):
@miteshget This looks good - but I'd like another when condition to only run the osp network stuff when the variable ocp_use_single_network is set to true.

miteshget:
@wkulhanek I have added condition, please review.

wkulhanek (approved):

PR #3256, opened 2020-12-03
Expose internal registry for core cluster

ecwpz91:
##### SUMMARY

Expose internal registry for core cluster.

Without this patch the `Retrieve registry route` task on line 43 fails because the registry route is not exposed by default.

Fixes merged [PR 3071](https://github.com/redhat-cop/agnosticd/pull/3071/files)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Role ocp4_workload_dso


sabre1041 (approved):
LGTM

PR #3297, opened 2020-12-10
Fix role ocp4-workload-dil-streaming route_subdomain var

jkupferer:
##### SUMMARY

Fix setting rout_subdomain in ocp4-workload-dil-streaming.

The role did not specify `api_version` when attempting to fetch the ingress configuration, resulting in getting a result from the wrong api group.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Role ocp4-workload-dil-streaming


rut31337 (approved):
LGTM

PR #3332, opened 2021-01-05
Add support for AWS region me-south-1

paulbarfuss:
##### SUMMARY

Add support for AWS Region Bahrain (me-south-1)

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME

- infra-ec2-template-generate

##### ADDITIONAL INFORMATION

We are looking to add support for the me-south-1 region. Note that the `RHEL81NBDE` image doesn't appear to be available and returned `None` so I left that one out. Other than that one AMI, I was able to get the rest of the images using `tools/find_ami.sh`

Also, I added a DNS mapping for `me-south-1.compute.internal`


stencell (approved):
This looks okay.

@rut31337 It looks like this is a new region that hasn't been enabled in our accounts. FYI if you want to get that enabled. If so, the RHEL NBDE ami needs to be shared manually to it like we did with the other regions.

PR #3333, opened 2021-01-06
Add bash shell init flag to deploy_certs.sh

jkupferer:
##### SUMMARY

Without the "login" shell flag, `-l` bash does not initialize the shell environment which causes problems for ansible execution from cron. This PR adds this flag to the shebang `#!` line of teh deploy_certs.sh in the three roles in which it is used.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Roles:
- ocp4-workload-enable-lets-encrypt-certificates
- ocp4-workload-idm
- ocp4_workload_le_certificates


wkulhanek (approved):
This is weird. It works fine on DEV4 and NA4 without that flag. But I guess it can't hurt...

PR #3350, opened 2021-01-12
Fix Satellite DNS resolution

marcosmamorim:
##### SUMMARY

Fix issue to deploy Satellite when using OSP 16, we have to add the internal hostname and IP address to /etc/hosts file

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-smart


rhjcd (approved):
💯

PR #3351, opened 2021-01-12
Change role host-ocp4-install to implement retry

jkupferer:
##### SUMMARY

- Add retry OpenShift install on failed waiting for Kubernetes API.
- Also add yamllint and cleanup.
- Remove non-functional OpenStack host restart.
- Remove retry wait for bootstrap complete as I could find no evidence of this succeeding in the logs after initial failure, but many cases where it simply made the failure take longer to be reported.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role host-ocp4-install

### DIFF
--- a/ansible/roles/host-ocp4-installer/.yamllint
+++ b/ansible/roles/host-ocp4-installer/.yamllint
@@ -1,0 +1,2 @@
+---
+extends: ../../../tests/static/.yamllint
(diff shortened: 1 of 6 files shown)


wkulhanek (dismissed):

wkulhanek (approved):
I'm not sure that install / destroy / install will fix the problems of some nodes showing up as NotReady on OpenStack.... but the logic should at least not break anything.

PR #3354, opened 2021-01-13
tech-breakfix-shift-004 add binary grading script

ashish-k-s:
##### SUMMARY
Add binary version of grading script for the module

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


miteshget (commented):
Binary file should be kept outside agnosticD as per guidelines.

ashish-k-s:
> Binary file should be kept outside agnosticD as per guidelines.

Hi Mitesh,

This PR is based on our old catalog entries cee-cf-11[0,1,2] which were worked on long time back.
We had followed this practice in the above three catalogs 
I agree your suggestion and will keep that in mind for new catalog entries in the future.
But this request is not the new catalog creation from scratch but is based on existing catalog.
Let me know if this justification is enough for approval of this PR.
If not, I will have to re-work a big task.

miteshget:
Hi @fridim, could you please review comment and PR?

jkupferer:
@ashish-k-s Can you provide some background on what this binary is and where it is coming from? I see it appears to be a compiled executable?

ashish-k-s:
> @ashish-k-s Can you provide some background on what this binary is and where it is coming from? I see it appears to be a compiled executable?

https://github.com/redhat-cop/agnosticd/blob/37d5956bd41ae4bbac0bdf4b7c513afb0b961c5a/ansible/roles_ocp_workloads/ocp4_workload_pntae/files/tech-breakfix-shift-004
Above is the text version of this script.
Associated training is a breakfix activity.
The script is converted to binary to hide the break and grading steps from learner.

jkupferer (approved):

jkupferer:
@miteshget Approving. @ashish-k-s and I discussed. The break/fix script needs to be binary to avoid exposing the break code to lab users and so give away what needs fixed. Normally we would not allow binaries, but the binary is small and the rationale is clear. Making an exception for this case.

PR #3368, opened 2021-01-18
Iterate over SAP hosts only

makentenza:
##### SUMMARY

Iterate over SAP hosts only (HANA and S4HANA) to avoid issues while getting facts

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sap-hana


marcosmamorim (approved):
LGTM

PR #3380, opened 2021-01-27
Use fixed version (8.1) of UBI image so it's compatible with RHEL8.1 repo

ggbecker:
##### SUMMARY
We have to use a specific version of UBI because we use the repo configured in the HOST to install new packages into the built UBI image. in this case the repo configured is RHEL8.1, otherwise it there is package conflicts.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

rhel-custom-security-content

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/configs/rhel-custom-security-content/data/Dockerfiles/test_suite-rhel8
+++ b/ansible/configs/rhel-custom-security-content/data/Dockerfiles/test_suite-rhel8
@@ -1,5 +1,5 @@
 # This Dockerfile is a minimal example for a RHEL 8-based SSG test suite target container.
-FROM registry.access.redhat.com/ubi8
+FROM registry.access.redhat.com/ubi8:8.1
 
 ENV AUTH_KEYS=/root/.ssh/authorized_keys
 


stencell (changes requested):
This should use `registry.redhat.io/ubi8:8.1`

Also, that image is 10 months old and has a health index of D (meaning it probably has lots of vulns). Is it possible to bump to a newer version or does this config have hard dependencies on 8.1?

jkupferer (approved):

ggbecker:
> This should use `registry.redhat.io/ubi8:8.1`
> 
> Also, that image is 10 months old and has a health index of D (meaning it probably has lots of vulns). Is it possible to bump to a newer version or does this config have hard dependencies on 8.1?

We have to install packages into the container and the repo used is the one from the host which is a RHEL8.1 repo, and it's not possible to install the required package in the container because of package conflicts. So it's a hard requirement.

rut31337 (approved):

PR #3403, opened 2021-02-04
ocp4-cluster Improve instance type of master

fridim:
- change to c5d (better perf)
- the sizing depends on number of workers. xlarge is generally enough,
  but can be a problem if the number of workers is high.

### DIFF
--- a/ansible/configs/ocp4-cluster/default_vars_ec2.yml
+++ b/ansible/configs/ocp4-cluster/default_vars_ec2.yml
@@ -76,11 +76,14 @@ bastion_instance_image: RHEL82GOLD
 # Root Filesystem Size
 bastion_rootfs_size: 30
 
-master_instance_type: >-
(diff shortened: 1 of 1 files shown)


ansible/configs/ocp4-cluster/default_vars_ec2.yml
```diff
@@ -76,7 +76,11 @@ bastion_instance_image: RHEL82GOLD
 # Root Filesystem Size
 bastion_rootfs_size: 30
 
-master_instance_type: "m5a.2xlarge"
+master_instance_type: >-
+  {{ 'c5d.xlarge' if worker_instance_count|int <= 10
```
fridim, line 80: make `c5d` another variable, like `master_instance_family`

wkulhanek (approved):
Cool. So that's default if nothing else is set.

PR #3410, opened 2021-02-09
Updated logic to allow for stable releases of OC Client rather than h…

wkulhanek:
##### SUMMARY

Updated logic to allow for stable releases of OC Client rather than having to specify the full release number.

Now the variable **studentvm_ocp4_oc_version** can be either just the major OpenShift release (4.4, 4.5, 4.6, ...) and it will install the latest available **stable** version. Or the variable can be set to a specific release (e.g. 4.6.15).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles_studentvm / studentvm_ocp4


rhjcd (approved):
thank you

PR #3417, opened 2021-02-10
Ocp4 disco scsi symlink

newgoliath:
##### SUMMARY

There's a bug in OCP 4.6 that doesn't account for 20+ character limit in filenames when using cinder on OSP.

Add files for iscsi 20 char symlink limit workaround to $HOME/resources for students to execute.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-disconnected-osp-lab


wkulhanek (approved):
Although rather than doing it in the config we could just add it to the instructions... easier to maintain.

newgoliath:
Copy paste is often a challenge.  And I get points for the easy pr. 😁

PR #3418, opened 2021-02-11
ocp4-cluster config: IOPS setting for master storage

fridim:
Allow user to specify the desired IOPS for master storage.

In case of large clusters the master instance should have provisioned
IOPS for the etcd database to work properly.

This change, if applied, automatically sets the
`controlPlane.platform.aws.rootVolume.type` to `io1` if the number of
workers is greater than 10.

In that case, the IOPS value is also calculated:

* minimum of 2000 IOPS,
* if number of wokers > 20, then it's 100 IOPS per node
* maximum of 32000 IOPS

That IOPS value can be overridden by the user with the `master_storage_iops` variable, directly in the config or agnosticv.

According to OpenShift documentation
https://docs.openshift.com/container-platform/4.6/installing/installing_aws/installing-aws-customizations.html

* type = io1 / 2000 IOPS  is a good setting for large clusters

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
* ocp4-cluster config
* host-ocp4-installer role
##### ADDITIONAL INFORMATION
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html

### DIFF
--- a/ansible/configs/ocp4-cluster/default_vars_ec2.yml
+++ b/ansible/configs/ocp4-cluster/default_vars_ec2.yml
@@ -76,6 +76,8 @@ bastion_instance_image: RHEL82GOLD
 # Root Filesystem Size
 bastion_rootfs_size: 30
 
+# Masters
+
 master_instance_type_family: c5d
 master_instance_type_size: >-
   {{ 'xlarge' if worker_instance_count|int <= 10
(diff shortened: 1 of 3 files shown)


wkulhanek (approved):
/LGTM. Although if we do that as discussed maybe switch masters back to m5a.

wkulhanek (approved):

rut31337 (approved):

PR #3421, opened 2021-02-11
add information to use and configure own AK when no Sat server is available - agd training

amayagil:
##### SUMMARY
add information to use and configure own AK when no Sat server is available

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

06_deploying_a_base_config_osp.adoc
```


wilson-walrus (approved):
Looking good

PR #3428, opened 2021-02-12
updated variables and tasks

miteshget:
##### SUMMARY
1: Renamed ceph_cluster_color to ceph_cluster - For better synchronization with osp-migration config and to get updated.

2: Moved infra-local-create-ssh_key role to pre_software.yml as it was always been there. It also helps in troubleshoot if all configs are prepared in same way.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Config: smart-management-foundation


bosebc (approved):
Looks good. Thanks

PR #3444, opened 2021-02-17
Add agnosticd_save_output_dir

jkupferer:
##### SUMMARY

Add ability to save `output_dir` content to AWS S3 cloud storage.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role agnosticd_save_output_dir and core playbooks.


fridim (changes requested):
As discussed, what if ansible fails somewhere, that will not be executed

We need a tarball of output_dir for troubleshooting because something went wrong. If ansible runs ok, it's very unlikely we ever need to fetch the output_dir tarball.

jkupferer:
I agree, but I think we will just have to make do with strategically adding `rescue` blocks with calls to this role. I don't see any better way possible. This is just meant as a start of the implementation.

jkupferer:
I just investigated if we could use a background `async` Ansible task that watches for failures in the main playbook... the answer is no. It won't work. Ansible sends a SIGKILL to async processes on failure and so there is no opportunity to handle and respond to the signal.

fridim:
That's going to be a lot of invocations to the role in all different `rescue` blocks in agnosticd. 
I don't think it'll scale and the maintenance cost for it doesn't look attractive.

 Especially when all you want is the tarball to narrow the issue, and instead you'll have to update agnosticd with a PR to add the new `rescue` block, push to development, order a new environment, hope to replicate the issue, and finally get the output_dir.
I don't see that happening (especially when you're troubleshooting), and we'll probably end up just finding another route.

jkupferer:
It shouldn't really be all that many if also reform our use of playbooks. We currently have too many plays and playbooks, we can reduce that down significantly... but I agree that it still won't be pretty.

The only reason I still favor this approach is that I don't believe any other approach is possible at this time.

ansible/main.yml
```diff
@@ -97,4 +97,6 @@
     - post_software
     - post_software_tasks
 
+- import_playbook: save_output_dir.yml
+
```
fridim, line 101: That playbook is missing in your PR ?

fridim (changes requested):
I'll approve but let's make sure it doesn't break any development.
we can `include_role` with a condition ensuring all mandatory variables of the role are defined. If not, then just pass.

ansible/main.yml
jkupferer: Ah! Good catch! I did miss it. Adding it now.

fridim (approved):

PR #3549, opened 2021-03-22
Fix missing GUID. Closes #3548

### LINKED ISSUE #3548: Bastion role bash_profile GUID
**Describe the bug**
ansible/roles/bastion/files/bash_profile
export GUID=`hostname | awk -F. '{print $2}'` does not set GUID.

In Novello/OSP env internal DNS sets the hostname as control/bastion or control.example.com/bastion.example.com

**To Reproduce**
Steps to reproduce the behaviour, ex:
1. Deploy this config using those variables
2. connect to the host, and perform echo $GUID.
4. No output.

**Expected behavior**
echo $GUID should return GUID

wkulhanek:
##### SUMMARY

Replace "hostname" logic in .bashrc with just a {{ guid }}. Fixes issue #3548 

FYI: @prakhar1985 (OpenShift environments have mostly switched to bastion-lite instead of bastion - guess that's why nobody noticed until now.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Role: bastion

### DIFF
--- a/ansible/roles/bastion/tasks/main.yml
+++ b/ansible/roles/bastion/tasks/main.yml
@@ -188,8 +188,8 @@
     - install_bash_customization
 
 - name: Install .bash_profile
-  copy:
-    src: ../files/bash_profile
+  template:
+    src: ../templates/bash_profile.j2
     dest: "{{ item.directory }}/.bash_profile"
     mode: 0644
     owner: "{{ item.user }}"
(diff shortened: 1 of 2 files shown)


prakhar1985 (approved):

prakhar1985:
@wkulhanek thanks for the quick fix.

PR #3628, opened 2021-04-07
Nenad post

pericnenad:
##### SUMMARY

Fixing a bug in quoting.
Adding bastion hostnames as variables instead of hard-coded names

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

role, ocp4_workload_migration


pericnenad:
Erroring out. Need to troubleshoot what is going on.

PR #3653, opened 2021-04-14
Changed hard-coded bastion to variable

pericnenad:
##### SUMMARY

Changed a @bastion. hard-coded name into an ansible variable

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp4-cluster post_software.yml

### DIFF
--- a/ansible/configs/ocp4-cluster/post_software.yml
+++ b/ansible/configs/ocp4-cluster/post_software.yml
@@ -186,7 +186,7 @@
         msg: "{{ item }}"
       loop:
       - "You can access your bastion via SSH:"
-      - "ssh {{ student_name }}@bastion.{{ guid }}{{ subdomain_base_suffix if cloud_provider == 'ec2' else '.'+ocp4_base_domain }}"
+      - "ssh {{ student_name }}@{{ hostvars[bastion_hostname]['ansible_hostname'] }}.{{ guid }}{{ subdomain_base_suffix if cloud_provider == 'ec2' else '.'+ocp4_base_domain }}"
       - ""
       - "Make sure you use the username '{{ student_name }}' and the password '{{ hostvars[bastion_hostname]['student_password'] }}' when prompted."
     - name: Print kubeadmin user info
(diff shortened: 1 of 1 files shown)


ansible/configs/ocp4-cluster/post_software.yml
```diff
@@ -186,7 +186,7 @@
         msg: "{{ item }}"
       loop:
       - "You can access your bastion via SSH:"
-      - "ssh {{ student_name }}@bastion.{{ guid }}{{ subdomain_base_suffix if cloud_provider == 'ec2' else '.'+ocp4_base_domain }}"
+      - "ssh {{ student_name }}@{{ hostvars[bastion_hostname]['ansible_hostname'] }}.{{ guid }}{{ subdomain_base_suffix if cloud_provider == 'ec2' else '.'+ocp4_base_domain }}"
```
stencell, line 189: This is stored as a fact with localhost...can't it be more simple? I think just `{{ bastion_hostname }} should be fine since this task is running on localhost ?

pericnenad: Hmm good point, but I haven't tested that. 
Does {{ bastion_hostname }} store only the host part or the full FQDN?

stencell: Ah, nvm that is gonna grab everything. To be consistent with what we use earlier, we could also use `hostvars[bastion_hostname].shortname`.

pericnenad: All-right, will change it into shortname then and have it all consistent.

stencell: This is where we set it earlier

https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-common-ssh-config-generate/tasks/main.yml#L42

pericnenad: Ok. Updated with the next commit.

stencell (approved):

PR #3675, opened 2021-04-16
Fixing pip requirements for venv.

pericnenad:
##### SUMMARY

Something has changed with google authenticator package which is getting included in venv.
It now requires a lower version of rsa to be able to install venv properly, compared to what we had in requirements.txt. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
role, ocp4-workload-migration


NonyNo3lle (approved):
LGTM

PR #3703, opened 2021-04-23
Update config sandbox: enable web console access for AWS sandbox

fridim:
This commit, if applied, adds a new feature to the 'sandbox' config to
enable access to the AWS Web console to students.

- introduce new boolean variable sandbox_enable_console_access (default: true)
- Remove useless tasks in ec2_destroy_env.yml
- Enable LoginProfile for the student user in the cloudformation template
- use include_tasks to dynamically load either ibm  or ec2 tasks:
  when provisioning for ec2, we don't want to display IBM tasks and vice versa.
- Display access in user.info

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- sandbox config

### DIFF
--- a/ansible/configs/sandbox/default_vars_ec2.yml
+++ b/ansible/configs/sandbox/default_vars_ec2.yml
@@ -12,4 +12,4 @@ chomped_zone_internal_dns: "{{ guid }}.internal"
 
 infra_ec2_template_generate_auto_select_availability_zone: false
 
-sandbox_enable_console_access: true
+sandbox_enable_ui: true
--- a/ansible/configs/sandbox/default_vars_ec2.yml
+++ b/ansible/configs/sandbox/default_vars_ec2.yml
@@ -11,3 +11,5 @@ zone_internal_dns: "{{ guid }}.internal."
 chomped_zone_internal_dns: "{{ guid }}.internal"
 
 infra_ec2_template_generate_auto_select_availability_zone: false
+
+sandbox_enable_console_access: true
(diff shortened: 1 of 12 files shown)


ansible/configs/sandbox/default_vars_ec2.yml
```diff
@@ -11,3 +11,5 @@ zone_internal_dns: "{{ guid }}.internal."
 chomped_zone_internal_dns: "{{ guid }}.internal"
 
 infra_ec2_template_generate_auto_select_availability_zone: false
+
+sandbox_enable_console_access: true
```
stencell, line 15: Can we change this to `sandbox_enable_ui`? We already use this to enable console access to IBM Cloud and that is live. If you don't want to change it here, we should change the var we use for IBM Cloud to be the same for consistency.

stencell (approved):
One comment to address if you want, otherwise lgtm. And thanks for cleaning it up to make it more readable 😸

rut31337 (approved):

jgholt:
@fridim Tried to provision on DEV today and got this error:

fatal: [localhost]: FAILED! => {"reason": "no action detected in task. This often indicates a misspelled module name, or incorrect module path.\n\nThe error appears to have been in '/tmp/sandboxes-gpte-SANDBOX-dev-67ba-20791/agnosticd/ansible/configs/sandbox/pre_infra_ec2.yml': line 7, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Get the current caller identity information\n  ^ here\n"}
FAIL sandboxes-gpte-SANDBOX-dev-67ba run_playbook main.yml

fridim:
Hi @jgholt 

It's because the catalog item doesn't have any python virtualenv.

I submitted a change in the agnosticv Repo.

Thanks for the heads up.

PR #3752, opened 2021-05-06
Add ilt final lab

wkulhanek:
##### SUMMARY

Big, beautiful, actually the most big and beautiful PR. :-) Just big and beautiful because it deploys a lot. This is actually reasonably light-weight.

Adding a workload to deploy the environment for the OpenShift Advanced App Deploy (2021 version) Final Lab environment.

Currently deploys to DEV4, NA46 will need updates before it can be deployed there.

##### ISSUE TYPE
- New role Pull Request

##### COMPONENT NAME
ocp4_workload_adv_app_deploy_ilt_final_lab


fridim (approved):
Too big, too beautiful.

stencell (approved):
61 files 🙄 

I think it is time for an update to PR standards that it can't be more than 60 files 😆 

Also, please fix the failed check.

wkulhanek:
@stencell I hear you... but 90% of these files are just templates. It's really just 5 files that need reviewing here. :-)
Also there isn't really a way to split this. This is one role.

PR #3759, opened 2021-05-11
apicurio registry to be installed in each student namespace.

varodrig:
SUMMARY:

This is a workload  to install Apicurio Registry Operator at namespace level per each student. In this case at the schema-registry-userX. For now it'll be used in the Adv. CND VILT, but can be used by any catalog.

ISSUE TYPE

New config Pull Request

COMPONENT NAME
ilt-adv-cloud-native-development-prod

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /defaults/main.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /defaults/main.yml
@@ -31,7 +31,7 @@ ocp4_workload_apicurio_registry_catalogsource_name: community-operators-snapshot
 ocp4_workload_apicurio_registry_catalog_snapshot_image: quay.io/gpte-devops-automation/olm_snapshot_community_catalog
 
 # Catalog snapshot image tag
-ocp4_workload_apicurio_registry_catalog_snapshot_image_tag: "v4.5_2021_05_10"
+ocp4_workload_apicurio_registry_catalog_snapshot_image_tag: "v4.6_2021_05_10"
 
 # for creating users in apicurio registry
 ocp4_workload_apicurio_registry_operator_create_users: false


ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /tasks/remove_workload.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Remove OpenShift Apicurio Project
+  k8s:
```
wkulhanek, line 3: Does this need a when clause to only run when "set up users" is set to true?

ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /tasks/workload.yml
```diff
@@ -0,0 +1,33 @@
...
+    name: install_operator
+  vars:
+    install_operator_action: install
+    install_operator_name: apicurio-registry-operator
+    install_operator_namespace: 'user{{ item }}-schema-registry'
+    install_operator_channel: "{{ ocp4_workload_apicurio-registry_operator_channel }}"
+    install_operator_catalog: ""
+    install_operator_automatic_install_plan_approval: "{{ ocp4_workload_apicurio-registry_operator_automatic_install_plan_approval | default(true) }}"
+    install_operator_starting_csv: "{{ ocp4_workload_apicurio-registry_operator_starting_csv }}"
+    install_operator_use_catalog_snapshot: true
```
wkulhanek, line 23: This should probably come from a variable and not be hardcoded to true.

ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /tasks/workload.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Create OpenShift Apicurio Project
+  k8s:
```
wkulhanek, line 3: Is this related to the variable to "set up users" ? If so it probably needs a when clause.

wkulhanek (changes requested):

ansible/roles_ocp_workloads/ocp4_workload_apicurio_registry_operator /tasks/workload.yml
varodrig: I'm not setting up users. I'm creating a project per users.

wkulhanek (approved):

PR #3777, opened 2021-05-18
Update config ibm-rhoic

vincepower:
##### SUMMARY

Completely rearchitected approach based on feedback from IBM Cloud to use IBM Schematics with a terraform template to handle provisioning instead directly working with all the individual APIs. It has improved stability.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
config/ibm-rhoic

##### ADDITIONAL INFORMATION
@stencell  was involved with the discussions with IBM which drove the change


ansible/configs/ibm-rhoic/infra.yml
```diff
@@ -20,307 +20,148 @@
     - post_flight_check
   tasks:
     ## Authenticating
-    - name: Authorize to get bearer token for REST calls
+    - name: Authorize to get bearer access and refresh tokens
       uri:
         url: "{{ ibm_cloud_api_iam_url }}"
+        headers:
+          Authorization: Basic Yng6Yng=
```
vincepower, line 27: This is from IBM Cloud's documentation, not a secret we need to protect. This is username bx and password bx which is some kind of a system account in IBM Cloud. If I don't pass this header then it won't generate a refresh token which is needed for IBM Schematics plan and apply.

ansible/configs/ibm-rhoic/destroy_env.yml
```diff
@@ -58,212 +58,151 @@
       set_fact:
         sandbox_master_api_key: "{{ r_sandbox_account['json']['master_api_key']['S'] }}"
 
-    - name: Get bearer token
+    ## Authenticating
+    - name: Authorize to get bearer access and refresh tokens
       uri:
         url: "{{ ibm_cloud_api_iam_url }}"
+        headers:
+          Authorization: Basic Yng6Yng=
```
vincepower, line 66: This is from IBM Cloud's documentation, not a secret we need to protect. This is username bx and password bx which is some kind of a system account in IBM Cloud. If I don't pass this header then it won't generate a refresh token which is needed for IBM Schematics plan and apply.

ansible/configs/ibm-rhoic/software.yml
```diff
@@ -18,7 +18,34 @@
...
+      uri:
+        url: "{{ ibm_cloud_api_iam_url }}"
+        headers:
+          Authorization: Basic Yng6Yng=
+        method: POST
+        body_format: form-urlencoded
+        body:
+          grant_type: "urn:ibm:params:oauth:grant-type:apikey"
+          response_type: "cloud_iam"
+          apikey: "{{ sandbox_master_api_key }}"
```
stencell, line 32: We need to transition this to use the SID API key that is created vs the master key. The master key should really only be used to configure access to the account and for administrators to get in if necessary.

ansible/configs/ibm-rhoic/software.yml
```diff
@@ -52,88 +79,166 @@
...
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  become: false
+  tags:
+    - post_flight_check
+  tasks:
+    ## Authenticated from last section in this playbook
+    # Create Workspace to support RHOIC
+    - name: Create a new workspace
```
stencell, line 92: I haven't looked at this in a while as you know, but can't we create multiple schematics in the same workspace rather than having different workspaces for each thing?

stencell (changes requested):
Looks good overall, just a couple of comments to address and then this can move forward with some testing.

ansible/configs/ibm-rhoic/software.yml
vincepower: Commit a change right now to fix which key is used.

ansible/configs/ibm-rhoic/software.yml
vincepower: From everything I can tell it is one terraform repository per workspace. IBM recommended two workspaces as well.
I also tried to only use one set of terraform scripts for creating the VPC and RHOIC cluster and it works for the build, but I had failures testing the destroy. The failure to destroy with the combined scripts would always be related to not being able to remove the VPC when it has network interfaces associated with it.

stencell (approved):

PR #3782, opened 2021-05-21
Update config/ibm-rhoic and roles-infra/sandbox-ibm

vincepower:
##### SUMMARY

Correcting an until that was wrong in the destroy, also adding extra checks to the until in destroy so they won't fail when the API misfires.

Added a default variable just to make sure it has a value, and also a set_fact for email so it will always have a value even when testing locally.

Lastly renaming the files in the sandbox-ibm to match the name change from roks to rhoic. The content of the files was already fixed in a previous PR.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
configs/ibm-rhoic
roles-infra/sandbox-ibm

##### ADDITIONAL INFORMATION
N/A


ansible/configs/ibm-rhoic/software.yml
```diff
@@ -255,6 +255,10 @@
         var: r_existing_rhoic
         verbosity: 2
 
+    - set_fact:
+        email: "no-valid-email-set"
```
stencell, line 259: you can get a valid email address from `{{ email }}`

vincepower: Added a name that describes when this runs as per our conversation.

tonykay (approved):
Extra credit for adding those ```name:```!

PR #3833, opened 2021-06-07
Add HA support for satellite subscription

rut31337:
##### SUMMARY

Add HA support for satellite subscription

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

set-repositories-satellite

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/set-repositories/tasks/satellite-repos.yml
+++ b/ansible/roles/set-repositories/tasks/satellite-repos.yml
@@ -108,6 +108,16 @@
     server_hostname: "{{ set_repositories_satellite_hostname }}"
     activationkey: "{{ set_repositories_satellite_activationkey }}"
     org_id: "{{ set_repositories_satellite_org | default(satellite_org) }}"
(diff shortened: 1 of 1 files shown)


evgeni:
The only difference to the non-HA case is that you set `rhsm_baseurl` and provide a port/path in `server_hostname`?

This will only work for capsules (and that's the only "HA" scenario that is officially supported today).
And it wouldn't hurt (besides the port) to set on "normal" registration either (in fact, that's what `bootstrap.py` does anyways).

So instead of duplicating the whole block, I'd suggest to *always* pass all those extra params, and only differentiate the port based on sat/capsule (not non-HA/HA).

Or even better, you could get the port dynamically from rhsm.conf:
https://github.com/Katello/katello-client-bootstrap/blob/e6ce46b62fbb3414a2065facf5cb420801ac3fe5/bootstrap.py#L1008-L1015

marcosmamorim (approved):
LGTM

PR #3843, opened 2021-06-08
Convert sealed secrets from Operator to Helm Chart

wkulhanek:
##### SUMMARY

The Sealed Secret Community operator is crap... So replacing the Operator with the Helm Chart - which is also better maintained.

Need to use the helm binary because AgnosticD doesn't support using the k8s collection.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_sealed_secrets


gmontalvoy (approved):

gmontalvoy:
Tested on OpenShift 4.6.22 w/ Ansible 2.11

PR #3864, opened 2021-06-14
Add ability to not lookup public subnet for OpenStack provider

fdupont-redhat:
##### SUMMARY

When deploying to OpenStack, the router external IP address floating IP addresses are created on the provider network specified by the user via `provider_network`. If the provider network has more than one subnet, the logic is to pick the one with the least used IP addresses. This is done with `openstack ip availability` but it lookups a hard coded `external` network. This pull request replaces it with the `provider_network`.

And some clusters have a policy that forbids to specify the subnet when creating a router or floating IP. The pull request adds a boolean variable named `osp_find_public_subnet` that will enable lookup when set to true. Its default value is `true` to be backward compatible with the current behavior.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
roles-infra/infra-osp-project-create
roles-infra/infra-osp-template-generate

### DIFF
--- a/ansible/roles-infra/infra-osp-project-create/tasks/main.yml
+++ b/ansible/roles-infra/infra-osp-project-create/tasks/main.yml
@@ -32,7 +32,8 @@
     block:
     - name: Get public subnet to use
       shell: |
-        openstack ip availability show {{ provider_network }} -f json | jq -r '.subnet_ip_availability | min_by(.used_ips) | .subnet_id'
+        openstack ip availability show {{ provider_network }} -f json \
+        | jq -r '.subnet_ip_availability | min_by(.used_ips) | .subnet_id'
       register: r_floating_subnet
     - name: Set fact for public subnet
       set_fact:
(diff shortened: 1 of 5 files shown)


fridim (approved):
Took me a while to find out the version without the `shell` module + `jq`:

```yaml
    - name: Get public subnet to use
      command: >-
        openstack ip availability show {{ provider_network }} -f json
      register: r_floating_subnet

    - name: Set fact for public subnet
      set_fact:
        osp_public_subnet: >-
          {{ r_floating_subnet.stdout
          | from_json
          | json_query('subnet_ip_availability')
          | min(attribute='used_ips')
          | json_query('subnet_id')
          }}
```

The problem is that it doesn't work with all versions of ansible, `min` filter is new to 2.11.

So i guess it's fine to keep the PR as is.

ansible/roles-infra/infra-osp-project-create/tasks/main.yml
```diff
@@ -28,13 +28,16 @@
...
-  - name: Set fact for public subnet
-    set_fact:
-      osp_public_subnet: "{{ r_floating_subnet.stdout }}"
+  - when: osp_find_public_subnet | bool
+    block:
+    - name: Get public subnet to use
+      shell: |
+        openstack ip availability show {{ provider_network }} -f json \
+        | jq -r '.subnet_ip_availability | min_by(.used_ips) | .subnet_id'
+      register: r_floating_subnet
```
fridim, line 37: Please add `changed_when: false`

fdupont-redhat: Done.

fridim (approved):

PR #3880, opened 2021-06-18
Install operator updates

wkulhanek:
##### SUMMARY

Update install_operator role to support CatalogSource secrets and global operators in namespaces other than openshift-operators. 

Update rhods_olm workload to use that role (and implement deletion)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
install_operator
ocp4_workload_rhods_olm

@thoraxe FYI. That operator was tricky... it also wants to be a global operator in a namespace that's not `openshift-operators`...


newgoliath (approved):
Woah, that's a lot.  I'll need to review again.  Unblocking you...

PR #3883, opened 2021-06-21
Add manifest feature to the tower-license-injector role

ericzolf:
##### SUMMARY

Manifest can be loaded as URL, file or base64-variable

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

roles/tower-license-injector

##### ADDITIONAL INFORMATION

The code works with Tower 3.8 and the collection awx.awx version 19.2.0

### DIFF
--- a/ansible/roles/tower-license-injector/README.md
+++ b/ansible/roles/tower-license-injector/README.md
@@ -1,0 +1,1 @@
+CAUTION: the manifest approach works only with versions of awx.awx older than 19.2.0 (i.e. 19.1.0 or below), because the `eula_accepted` value has been removed but is required for versions of Tower 3.8 or less.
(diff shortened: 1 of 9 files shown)


ericzolf:
I think there is a glitch, I need to check again.

prakhar1985 (approved):
That's so cool. We were looking for something like this. Thanks

ericzolf:
Thanks. BTW, it doesn't _fail_ with version 19.2.0 of the collection but it doesn't work either, you need version 19.1.0 or the eula_accepted parameter is missing. It _seems_ that Tower 3.8.4 might work with 19.2.0 (19.2.1 again will require many changes to the code, or Ansible 2.11, to work properly).

PR #3884, opened 2021-06-21
updates odh olm role to use install_operator role

thoraxe:


wkulhanek:
I did that in the PR that just got merged... @thoraxe

wkulhanek (approved):

PR #3903, opened 2021-06-28
removing forced subnet option for external net to default false for O…

9strands:
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- FIP subnet picker for OSP heat templates

##### ADDITIONAL INFORMATION
Users are receiving the following errors in production openstack, which means something is using a full subnet (which should not happen).  The recent OSP hotfixes for OVN from Feb/Mar 2021 should remove the requirement to force a subnet.

See user error here - this is currently common for OSP provisions:
```
CREATE_FAILED  Resource CREATE failed: IpAddressGenerationFailureClient: resources.fip_bastion-4fnnv: No more IP addresses available on network e0ddc3ba-1707-44e1-a341-7c7006d60f45.",
```


rhjcd (approved):
👍

PR #3940, opened 2021-07-08
Fix check for number of periods in installer version

wkulhanek:
##### SUMMARY

We were checking for 3 parts (2 dots) - which fails for 4.8.0-rc.3. Need to check for >=3

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
host-ocp4-installer


newgoliath (approved):
LGTM

PR #3972, opened 2021-07-23
Add config ansible-workshops-test - migration from ansible/workshops

sstanovnik:
##### SUMMARY

Migrate the RHEL workshop from [ansible/workshops](https://github.com/ansible/workshops/).

This config provisions, but is not "finished", since we need to validate it works.

/cc @IPvSean, we can use this as the base for further planning

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME

`ansible-workshops-test`

##### ADDITIONAL INFORMATION

This is `ansible-workshops-test` as it includes _all_ workshops. I've only tested RHEL, which needs to be split out. Most code will need to be shared though, since all workshops have the same structure.

The workshop provisions in 10 minutes, which is an improvement. The launch command is in the readme.

I've had to very liberally use agnosticd's framework since there are some things the existing workshops do differently, mainly concerning how the inventory is built up. Most things are identical to what worked before, since I tried to break as little as possible. Machine provisioning is obviously completely rewritten into CloudFormation. EC2 is the only provider supported because of extremely tight coupling.

The code is a mess, I know this very well. Cleanups will happen when I know there won't be any major functionality changes necessary to avoid premature optimization.

I initially wanted to do a staged approach to migrating the workshops, but the logic that needed to be ported was too complex to be split up into nice, logical parts. In the end, I figured the workflow downsides outweighed any the benefits of any smaller, but still >2kLOC PRs.

Major non-standard components, also the main things one needs to pay attention to here, are:

* `default_vars.yml`, where I dumped, in order, all variable files from ansible/workshops. Duplicates aplenty, this is 800LOC with gratuitous whitespace.
* `ami_find/*` which are the existing AMI filtering playbooks.
* `autoincluded_vars_workshop_instances.yml`, which is a Jinja2 template of variables templated out and included in `pre_infra.yml` featuring multiple levels of templating and lazy evaluation.
    * This is where instances for each workshop are configured.
    * 1.3 kLOC
    * The schema is documented inside the file, but never validated.
    * This file is very ugly, but it works. Any improvements I could make would need to be propagated through the whole ansible/workshops structure and sometimes require knowledge of why workshops are set up the way they are.
* `ec2_cloud_template.j2` is not the "standard" agnosticd template, it is completely rewritten.
    * VPCs, subnets, secgroups and the like are mostly hardcoded.
    * Instances have a loop which is controlled by what is in `autoincluded_vars_workshop_instances.yml`.
* There is no bastion host and the SSH keypair is handled differently.
* Likely more I just don't remember off the top of my head.

TODOs:

* Split out the RHEL workshop (then others).
* Verify whether the RHEL workshop actually works (not just that it provisions).
* Need to manually convert any relevant changes in ansible/workshops since the commit I based this migration on.


abenokraitis:
@tonykay would appreciate a review if possible, thanks!

tonykay:
Taking a look right now @abenokraitis - thanks for all the work here

tonykay:
Adding some comments inline, because it is a config, and effectively sandboxed, we can merge with no impact to others. One potential issue I see is there are a number of hard coded names and filters which we would expect to be variables so we can update without a PR each time. I will add some comments to the code

ansible/configs/ansible-workshops-test/ami_find/ami_find_network.yml
```diff
@@ -0,0 +1,63 @@
+---
+#### CISCO AMI
+- name: BLOCK FOR CISCO AMI
+  block:
+    - name: find ami for cisco (NETWORKING MODE)
+      ec2_ami_info:
+        owners: "679593333241"
+        filters:
+          name: "cisco-CSR-.16.09.02-BYOL*"
```
tonykay, line 9: Should be a var, as this is likely to change over time.

ansible/configs/ansible-workshops-test/files/cloud_providers/terraform_ec2_cloud_template.tf.j2
```diff
@@ -0,0 +1,164 @@
+#----------------------------------------------
```
tonykay, line 1: Generally, we don't use the terraform cloud_provider

tonykay (commented):
Added several comments, happy to have a call or connect via chat if this helps. We can be a resource if you ahve any questions.

sstanovnik:
I've resolved most of the things we've talked about. Changes are in separate commits to make reviewing easier. Let me know if I need to squash anything.

AMIs are now completely parametrized and resolved in a loop, so variable overrides will work. Variables are also split into the "defaults", which normally wouldn't be changed, and the "sample vars", which are things you'd likely want to or need to customize. I think that's the way they were intended to be used.

The config was also renamed from `ansible-workshops-test` into `ansible-workshops`, with the other existing `ansible-workshops*` configs removed, since they'd be duplicates.

I couldn't manage to get things working with bastions hosts, i.e. with the built-in inventory builder. I saw no way of making host connections be direct, without proxying through the bastion host. This means the `ansible-playbook` command still needs to include `--skip-tags create_inventory,create_ssh_config,wait_ssh,set_hostname` to skip those steps. I don't know whether this is even possible in agnosticd, or whether the workshops can be adapted to function with a bastion host.

tonykay:
> I couldn't manage to get things working with bastions hosts, i.e. with the built-in inventory builder. I saw no way of making host connections be direct, without proxying through the bastion host. This means the `ansible-playbook` command still needs to include `--skip-tags create_inventory,create_ssh_config,wait_ssh,set_hostname` to skip those steps. I don't know whether this is even possible in agnosticd, or whether the workshops can be adapted to function with a bastion host.

The workshops should be able to function with a bastion host and the `--skip-tags` listed remove that capability. This is core functionality of AgnosticD and in effect is a large part of how it is cloud agnostic. 

(We can, and do, run agnosticd with `--skip-tags` but the above tag mix removes core functionality which would then have to be reimplemented per cloud provider)

Each cloud_provider has its own `create-Inventory` and `create_ssh_config` relies on that. This allows multi cloud deployment and the `create_ssh_config` is what enables configuration via bastion. It is simple configuring ansible to know how to contact each host via a jumpbox. 

You could effectively use your first control node for this bastion/jumpbox task. All it needs, and the bastions effectively get this "for free", is the correct ssh keys **and** IP connectivity to all hosts.

IPvSean:
call at 9AM ET between XLAB & GPTE

@tonykay will deploy workshop and talk to GPTE leadership to verify a list of what else is missing to merge PR

cc @abenokraitis

tonykay:
@sstanovnik can you paste in the exact command you are using to deploy, and confirm you are deploying with `2.9.x`? Thanks. If you have time adding a `requirements.txt` to the config would be useful or confirming which one you are using. They are in `tools/virtualenvs` The README's typically should contain the command to deploy so other users can basically just copy and paste to test. Thanks.

tonykay:
The README is important for ops. Needs to have a description so they know this is the port of the workshops as there will be 2 entries with `ansible-workshop` in their name. Plus detailed instructions on how to run. For example either a long `ansible-playbook cmd with all the `-e` options they need to set or simply a sample of what they would over-ride via `-e @my_vars.yml`. We typically split it into 2 files and have the secret one last. Happy to contribute once we have it added. Basically Ops need to know how to deploy/destroy more or less via copy and paste.

tonykay:
Re the README it would be good to call out the `offline_token` and how to get it (simply a link to the RHN docs will do). We would prefer it to have a more meaningful name as we use lots of tokens to pull from repos and repositories. A `rhn_` prefix will help ops if they see log errors etc.

sstanovnik:
The exact command was in the readme, but I missed changing the config path when I renamed it to `*-agnosticd`. I've now extended the readme with detailed instructions.

`offline_token` can't be renamed since it's a variable that ansible/workshops uses directly. Unless you want to duplicate it under a different name in the vars file, but that would break overriding.

I've added a `requirements.txt`. I wasn't using 2.9, but I've now verified this works under these two versions:

* `ansible 2.9.25`
* `ansible [core 2.11.2]`

tonykay:
Thanks @sstanovnik testing with 2.11. Note your point re the var name

tonykay:
I notice it looks up all amis for all workshops? Shouldn't this be conditional so that only the current workshop amis are looked up. For example if a network ami task breaks then the workshop will break for everyone plus it is a additional work. Can we wrap them in `when`?

tonykay:
Need to document all vars to supply as over-rides e.g `workshop_dns_zone`

tonykay:
@sstanovnik when tested on an environment running other workshops an issue arises here https://github.com/xlab-steampunk/agnosticd/blob/a38dae6689a97b35a691ec0d1d7c343a380741e4/ansible/configs/ansible-workshops-agnosticd/post_infra.yml#L23

Basically the filter picks up every instance in every running workshop. I suggest the filter is changed from a simple ` - name: Grab all workshop hosts
      ec2_instance_info:
        filters:
          instance-state-name: running
          "tag:ansible-workshops": "true"
      register: workshop_hosts
`
To using `guid` as well. Otherwise we will potentially pick up hundreds of instances before proceeding. It is already picking up the existing instances from the `ansible-workshop` config

tonykay:
@sstanovnik re the above filter picking up all the instances, this is how we normally do it and as long as you are creating the project tag including the value of `guid` all should be good as you'll only pick up the guid instances https://github.com/redhat-cop/agnosticd/blob/d2dd93ad960d5ae1e486eac3f9a7828b649bd9ef/ansible/roles-infra/infra-ec2-create-inventory/tasks/main.yml#L7

tonykay:
Hi @sstanovnik we need the filter issue fixed before we can carry out further tests. I'll check your repo morning Wednesday US time. This is blocking us from deploying safely.

sstanovnik:
I've limited instance lookups to the `guid`.

AMIs are all fetched because that was easiest. I could either

* mark each image with the workshops it's used in, which needs to be modified properly when images are added or workshops start using different ones, or
* parse the necessary images from the workshop instances, but this is a bit more work initially.

By documenting variables, do you mean documenting their meaning along their sample values, or documenting all of them, perhaps including the meanings, in the readme?

sstanovnik:
This is actually a red herring. the `*_node_facts` variables are dynamically created in `post_infra.py#36`. I did this to avoid porting the inventory/hosts templates because creating the necessary variables was much easier.

What went wrong is that the `Student` tag was set to `N` instead of `studentN`, which I must have missed when transcribing the tags. There is a filter in the templates that silently does nothing when anything other than the expected format is given. I've fixed this now and can confirm that the inventory files are populated with IPs.

IPvSean:
let me rebase and try! thanks Saso

IPvSean:
OK IP addresses are populating!  Great.  

Two issues that are hopefully minor->

1. can you **remove** the IBM community grid stuff, that is from an older project and I think GPTE has pushed back on this.
e.g. just remove the role calls here->
`/agnosticd/ansible/configs/ansible-workshops-agnosticd/pre_software.yml`

just delete these entirely 
```
- name: IBM community grid managed nodes
  hosts: "managed_nodes"
  become: true
  gather_facts: true
  environment: *aws_environment

  tasks:
    - name: install boinc-client and register
      include_role:
        name: ansible.workshops.community_grid
      when:
        - ibm_community_grid is defined
        - ibm_community_grid

- name: IBM community grid control node
  hosts: "control_nodes"
  become: true
  gather_facts: true
  environment: *aws_environment

  tasks:
    - name: install boinc-client and register
      include_role:
        name: ansible.workshops.community_grid
        tasks_from: auto_shutoff
      when:
        - ibm_community_grid is defined
        - ibm_community_grid
        
```

2. something is up with the roles being called in... pre_software

e.g.
```
- name: add dns entires for all student control nodes
  hosts: '*ansible-1'
  become: true
  gather_facts: false
  environment: *aws_environment
  tasks:
    - include_role:
        name: ansible.workshops.aws_dns
```

I am getting warnings->
```
[WARNING]: Could not match supplied host pattern, ignoring: *ansible-1

PLAY [configure ansible control node] *************************************************************************************************************************************************************************
skipping: no hosts matched
[WARNING]: Could not match supplied host pattern, ignoring: *ansible-2
[WARNING]: Could not match supplied host pattern, ignoring: *ansible-3
```
and skips that don't make sense to me->
```
PLAY [add dns entires for all student control nodes] **********************************************************************************************************************************************************
skipping: no hosts matched
```

so the DNS is not working right now, and its skipping a bunch of roles...
it seems like the group ansible-1,2,3 are being created

IPvSean:
actually i don't think clusters are supported, i think that is OK, so the easiest fix much be switching ansible-1, etc to `control_nodes` and i think this will just work

sstanovnik:
This seems to be my fault again for missing a double nested loop that created the control nodes in ansible/workshops. I've fixed the provisioning to be able to create multiple replicas of nodes (and thus support clustering), adding the `-N` suffix which was missing from the control node names, resulting in the above failures. Non-clustered deployment now executes the tasks that weren't executed before.

Clustering definitely does not work though, even in ansible/workshops. See the following section in `ansible/workshops:provisioner/provision_lab.yml#115`

```yaml
    - include_role:
        name: ansible.workshops.control_node
        tasks_from: package_dependencies
      when: create_cluster|bool
```

where `package_dependencies` should be `15_package_dependencies`. I've fixed this in this repo since it's one of the changes I need to sync manually anyway.

IPvSean:
ok @sstanovnik I got this to pass with some minor edits upstream in the `ansible.workshops` to support the output_dir for the control_node role and the populate_controller role.  I did a new release, tested it and created two small PRs to your fork/branch.  Can you accept those and I can re-test and that should finish out this project!

sstanovnik:
Great, thanks for the fixes! I've merged the PRs. Not for the (hopefully) final test!

9strands:
Cleaning old PR entries at end of 2022. Please reopen if this is still required.

PR #3986, opened 2021-07-27
Add UPDATE_COMPLETE as a valid Cloudformation status for ocp4-workshop destroy

paulbarfuss:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ocp4-workshop

##### ADDITIONAL INFORMATION
There are various reasons why ocp4-workshop's underlying Cloudformation stack might end up in `UPDATE_COMPLETE` rather than simply `CREATE_COMPLETE` and is a common reason to need to manually clean up resources created by the Openshift installer.

This will help a common situation where the Cloudformation resources get cleaned up but the Openshift environment gets left behind.

### DIFF
--- a/ansible/configs/ocp4-workshop/destroy_env.yml
+++ b/ansible/configs/ocp4-workshop/destroy_env.yml
@@ -390,4 +390,5 @@
       when:
         # delete stack if stack creation failed
         - stack_status is defined
-        - stack_status != 'CREATE_COMPLETE' or stack_status != 'UPDATE_COMPLETE'
+        - stack_status != 'CREATE_COMPLETE'
+        - stack_status != 'UPDATE_COMPLETE'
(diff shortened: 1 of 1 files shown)


ansible/configs/ocp4-workshop/destroy_env.yml
```diff
@@ -389,4 +390,4 @@
       when:
         # delete stack if stack creation failed
         - stack_status is defined
-        - stack_status != 'CREATE_COMPLETE'
+        - stack_status != 'CREATE_COMPLETE' or stack_status != 'UPDATE_COMPLETE'
```
oybed, line 393: this should probably be an `and`?

paulbarfuss: Agreed, dropped into another list item to make `and` as this makes more sense.

stencell:
FYI, there is very little maintenance done on this config. Things should move to ocp4-cluster, which has been the replacement for some time. I know some people still use it and that is fine, but unless you need something specific you might benefit from switching to the newer config.

paulbarfuss:
Closing due to deprecation of ocp4-workshop.

PR #3989, opened 2021-07-27
Added if logic for osp_public_subnet for sap smart mgmt

thomas-crowe:
Added if logic to jinja templates for osp_public_subnet


marcosmamorim (approved):
LGTM

PR #4020, opened 2021-08-04
CHANGE: roles-infra/infra-azure-ssh-key

vincepower:
##### SUMMARY

Occasional errors with key vaults and names containing multiple dashes in a row, this is the easiest way to fix that.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
roles-infra/infra-azure-ssh-key/tasks/main.yml

##### ADDITIONAL INFORMATION
There could be more elegant ways to do this, but this will work.


vincepower:
One of the checks failed and its just a very messy PR... new PR has been opened https://github.com/redhat-cop/agnosticd/pull/4029

PR #4074, opened 2021-09-03
ocp4-odf4-aiml-worshop workload 

guimou:
Two workloads created for the XRay Pipeline Lab:
- The XRay lab itself
- OCS deployment workload

Note: apparently some change in the `ocp4_workload_bookbag_user` workload, but I don't know where they come from... I did a rebase on latest `deployment`branch before submitting the PR, but it should not have done that...


d-jana (approved):
Looks good

PR #4145, opened 2021-10-06
Install ansible with disabled EPEL

ericzolf:
##### SUMMARY

In order to not use EPEL software, we install ansible beforehand from RHEL while disabling EPEL

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
config ansible-automation-platform

##### ADDITIONAL INFORMATION
only the config is impacted, nothing else.


d-jana (approved):
Looks good. Approving this as it is reducing duration.

PR #4177, opened 2021-10-20
Ansible Execution Environment support

fridim:
##### SUMMARY

It's becoming harder and harder to support all users' different plateforms for agnosticd development and onboarding.

For example, on some MacOS agnosticd doesn't work because of some security settings:

https://github.com/ansible/ansible/issues/32499

Plus, sometimes, our python virtualenvs just doesn't install on MacOS or other OSes.

Using images seems a good way to work around all the different variants of OS.

 
### Presentation as a user story

As an agnosticd user/dev, i want to be able to access a variety of supported agnosticd images (one for each cloud provider most likely) to facilitate onboarding.

This should be the recommended and supported way.

### In this PR

* Update Readme: Add instructions howto run Execution Environments
* Update Readme: Add instructions howto build Execution Environments
* Create new multi-cloud EE ansible2.9-python3.6

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request


tonykay:
Do you think that dates are the ideal way to name EEs? Would semantic versioning make more sense in the long run. Also, I assume in the longer term that collections could be baked into the Ees themselves? When we move towards 2.11 and beyond perhaps we should re-visit how we build these etc? Being close to the RH toolchain might pay dividends as the platform evolves.

fridim:
Yes, I have thought of using semantic versioning for the image tags but preferred the date in the end. The reason I think is because of the naming convention we have. (Ansible version + python version).

Semantic versioning is MAJOR. MINOR . PATCH.
For those image names, I don't think it makes sense?

Maintenance of those images will probably only be an upgrade of the python modules or ansible version. It's more a "snapshot" than a version of a software with an API or something.
I couldn't find a use-case where semantic made sense in this context.
- when do you update MAJOR?
- when do you update MINOR?
- when do you update PATCH?

Plus, in the registry, Quay.io, we could use labels if we want to specify which agnosticd version is supported. `agnosticd-v1` for example. wdyt?

We have the ansible version and python version in the name of the image. Maybe we can drop that with EE? Ex:

- a simpler name  like  `redhat-gpte/agnosticd-images:multi-cloud-v0.0.1`
- use labels to add additional information like `python = 3.6`, `ansible = 2.9.27`

But again, I can't think of a use-case when you "break the API" and change MAJOR and not MINOR?

rut31337 (approved):

tonykay (commented):
Structure, location, dir naming makes complete sense to me. I do think we want to discuss if there is going to be a 1:1 relationship with existing venvs and EEs? It might be equally, or more, valid to have a single EE ( @jkupferer ? ) or per cloud if forced to. Otherwise developers will be faced by choices they don't understand e.g. python 3.6 v 3.8.

One other option is say a legacy ee for 2.9 and a "AAP2" all cloud ee.

Finally, I think any naming should make sense to someone looking at say `podman image ls` so would propose something like `ee-agnosticd-<NAME>`. Then people have context to what these images are embedded in the name.

jkupferer (approved):

fridim:
> Finally, I think any naming should make sense to someone looking at say `podman image ls` so would propose something like `ee-agnosticd-<NAME>`. Then people have context to what these images are embedded in the name.

I just tested, we don't need to keep agnosticd in the name. By default this is the output of `podman image list`

```
$ podman image list
REPOSITORY                    TAG                                               IMAGE ID      CREATED      SIZE
quay.io/redhat-pfe/agnosticd  ee-equinix_metal-ansible2.9-python3.6-2021-11-03  705d0fe6fd18  2 weeks ago  1.48 GB

# haven't tested with private repo, but it should look like:
REPOSITORY                                                 TAG     IMAGE         ID  CREATED  SIZE        
image-registry.apps.open.redhat.com/agnosticd/multi-cloud  v0.0.1  705d0fe6fd18  2   weeks    ago   1.48  GB

```
The name of the repo is displayed so i think we can drop the `agnosticd` part in `ee-agnosticd-<NAME>` and make it just `ee-<NAME>`

wkulhanek (approved):

jkupferer (approved):

PR #4194, opened 2021-10-23
Set sensible defaults for test-empty-config on ec2 for simpler deploys

tonykay:
##### SUMMARY

config test-empty-config lacks sensible ec2 defaults leading to constructs like

`-e env_type=test-empty-config -e cloud_provider=ec2 -e guid=1234`

It is normal for the defaults file to contain these vars already set.

Given that this is often used by people getting started it makes sense to clean this up, hence this PR

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

config: test-empty-config

### DIFF
--- a/ansible/configs/test-empty-config/default_vars_ec2.yml
+++ b/ansible/configs/test-empty-config/default_vars_ec2.yml
@@ -1,10 +1,6 @@
 ---
 # mandatory to run ansible/destroy.yml playbook
-
 aws_region: us-east-1
-
 env_type: test-empty-config
 cloud_provider: ec2
-
 guid: test-config-ec2-00
-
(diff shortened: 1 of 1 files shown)


fridim:
default vars cannot set those variables, they need to be passed as extra vars, see https://github.com/redhat-cop/agnosticd/blob/development/docs/Variables.adoc
If you don't, `main.yml` playbook fails at `import_playbook`, see here: https://github.com/redhat-cop/agnosticd/blob/development/ansible/main.yml#L22


You can define them in sample_vars_* though, as those are passed with `-e @sample_vars_ec2.yml` .
It's already the case here: https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/test-empty-config/sample_vars.yml

PR #4243, opened 2021-11-17
Fix permission error

wkulhanek:
##### SUMMARY

Oops. Changing /root requires become: True

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
host-ocp4-installer


tonykay (approved):
You could put it in a `block:` with one become, but no big deal

PR #4253, opened 2021-11-23
sap-smart

rickgcv:
##### SUMMARY

Added fix so that there are no Pulp errors when trying to apply packages with the lab's Satellite

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

sap-smart
##### ADDITIONAL INFORMATION

```paste below

```


ansible/configs/sap-smart/post_software.yml
```diff
@@ -133,6 +133,28 @@
...
 
+- name: Prevent Pulp 500 error
+  hosts: sats
+  become: yes
+  tasks:
+
+    - name: Make sure MongoDB will be able to start
+      shell: >
+        echo "-vv" > /.autorelabel &&
+        foreman-maintain service stop &&
```
jkupferer, line 144: This seems not very ansible-y? Particularly echoing to a file with `shell`. It is also not clear what this shell command has to do with MongoDB? Maybe something like:

```
  - name: Create /.autorelabel to trigger SELinux relabeling
    copy:
      content: -vv
      dest: /.autorelabel
      
  - name: Run this command for some reason...
    command: foreman-maintain service stop
```

ansible/configs/sap-smart/post_software.yml
```diff
@@ -133,6 +133,28 @@
...
+
+    - name: Reboot Satellite server
+      shell: "sleep 5 && reboot"
+      async: 1
+      poll: 0
```
jkupferer, line 149: Perhaps use `pause` instead of sleeping in a shell command? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pause_module.html

And rather than using the `reboot` command, use the `reboot` module?
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/reboot_module.html

jkupferer (approved):

PR #4257, opened 2021-11-25
Create a new role ocp4-workload-oadp for a new CI

pericnenad:
##### SUMMARY

This is a PR which enables a new role for a workload on top of OCP4. 
It is connected to a CI OCP4_OADP
It was tested and it deploys successfully in test clusters outside of AgD and it deploys in AgD DEV environment without errors. 

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME

role name: ocp4-workload-oadp

##### ADDITIONAL INFORMATION

Thank you :)

### DIFF
--- a/ansible/roles/ocp4-workload-oadp/tasks/post_workload.yml
+++ b/ansible/roles/ocp4-workload-oadp/tasks/post_workload.yml
@@ -101,26 +101,26 @@
   #               name: quay.io/openshifthomeroom/workshop-dashboard:5.0.1
   #           type: Source
   #         triggers: []
-  - name: "Building bookbag image"
-    shell: "{{ item }}"
-    args:
-      chdir: "{{ bookbag_dir }}/{{ bookbag_build_dir }}"
-    loop:
-    - "oc project lab-instructions"
-    - "oc process -f build-template.yaml -p GIT_REPO='{{ bookbag_repo }}' | oc apply -f -"
-    - "oc start-build bookbag --follow --from-dir={{ bookbag_dir }}"
-  - name: "Deploying bookbag image"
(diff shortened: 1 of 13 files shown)


ansible/roles/ocp4-workload-oadp/tasks/post_workload.yml
```diff
@@ -0,0 +1,158 @@
...
+  - name: create .aws dir
+    file:
+      owner: "{{ ansible_user }}"
+      state: directory
+      path: "{{ bookbag_dir }}/{{ bookbag_build_dir }}/.aws"
+  - name: create .aws/credentials file to noobaa s3
+    copy:
+      owner: "{{ ansible_user }}"
+      content: "{{ cloud_credentials_data_string }}"
+      dest: "{{ bookbag_dir }}/{{ bookbag_build_dir }}/.aws/credentials"
```
fridim, line 60: does it mean  the secret is exposed to the student, and this workload can only be executed in an AWS sandbox?

pericnenad: Hi,  yes this will be exposed to the student because that API and storage will be used in the lab as it currently is.
And you are right, it will need to be limited to AWS sandbox. 
If possible we can maybe change that in the future (but that would be on a TODO at any rate now).

pericnenad: Ok, I misunderstood the question at first. 

I removed the .aws dir and the credentials save. This will cause only a minor change in the lab docs. 
This change will not influence the operation of the lab and showcasing of capabilities as such. 

Thanks for bringing this to my attention!

fridim (approved):

PR #4260, opened 2021-11-29
Add new config osp-example-workflow-images

agonzalezrh:
This new ansible config is used to demonstrate the new OSP image workflow


9strands:
Cleaning old PR entries at end of 2022 which have not had work done in a year and which are out of date with `development` Please reopen if this is still required and refresh against the current primary branch.

PR #4261, opened 2021-11-29
Adding configs osp-generate-undercloud and osp-generate-ipmihost

agonzalezrh:
These configurations uses the image workflow to generate images to be used for OSP labs


9strands:
Cleaning old PR entries at end of 2022 which have not had work done in a year and which are out of date with `development` Please reopen if this is still required and refresh against the current primary branch.

PR #4267, opened 2021-11-30
Introducing dynamic images

fridim:
see https://issues.redhat.com/browse/GPTEINFRA-726

##### SUMMARY

As an agnosticd user/dev, i want to specify the images i want to use for each host using a combination of:

* filter of image name
* tags
* owner
* ...

Then agnosticd would dynamically get the latest image(s) in pre_infra stage, grab the id of the images and then use that in the next stages.

That feature should be *cloud agnostic*, and should work the same regardless of the cloud provider.

In AWS, Currently the images are all Hardcoded in agnosticd:  https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles-infra/infra-ec2-template-generate/defaults/main.yml#L98

That is cumbersome and not really easily maintainable.

Instead, for each hosts of an agnosticd config, i want to be able to define the images to be used with a combination of:
 
 * owner
 * tags
 * name filter

This change, if applied, adds the ability to pass the filters using the `instances[].image` dictionary. Example playbook:

```yaml
- hosts: localhost
  gather_facts: false
  vars:
    cloud_provider: ec2
    aws_region: us-west-2
    image_owner_id_redhat: 309956199498
    instances:
    - name: "bastion"
      image:
        # Latest RHEL 8.4 GOLD image
        name: RHEL-8.4*Access*
        architecture: x86_64
        owner: "{{ image_owner_id_redhat }}"
        aws_filters:
          is-public: false # <1>

    - name: "node"
      image:
        # Latest RHEL 9 GOLD image
        name: RHEL-9*Access*
        architecture: x86_64
        owner: "{{ image_owner_id_redhat }}"
        aws_filters:
          is-public: false

    - name: "worker"
      image:
        # Latest RHEL 8 GOLD image
        name: RHEL-8*Access*
        architecture: x86_64
        owner: "{{ image_owner_id_redhat }}"
        aws_filters:
          is-public: false

    - name: "frontend"
      image:
        # latest RHEL 8 public
        name: RHEL-8*
        architecture: x86_64
        owner: "{{ image_owner_id_redhat }}"
        aws_filters:
          is-public: true
    
    # You can also use predefined commonly used dict by providing string
    - name: "bastion-rhel84-gold"
      image: RHEL84GOLD-latest # <2>

    - name: "bastion-rhel83-gold"
      image: RHEL83GOLD-latest

    - name: "bastion-rhel7-gold-latest"
      image: RHEL7GOLD-latest

    - name: "bastion-rhel77-gold-latest"
      image: RHEL77GOLD-latest

    - name: "bastion-rhel8-gold-latest"
      image: RHEL8GOLD-latest

    - name: "bastion-rhel9-gold-latest"
      image: RHEL9GOLD-latest

    - name: "bastion-rhel9-latest"
      image: RHEL9-latest

    - name: "bastion-rhel8-latest"
      image: RHEL8-latest

    - name: "bastion-rhel7-latest"
      image: RHEL7-latest


  tasks:
    - include_role:
        name: infra-images
```
note: `instances[].image.architecture` defaults to `x86_64`

<1> see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html for filters
<2> You can also pass a string. See commonly used predefined dicts in the link:defaults/main.yaml[defaults/main.yaml]. If it's not a predefined dict, then image fallbacks to the mapping in the cloudformation template.

##### ISSUE TYPE

- Feature Pull Request
- New role Pull Request

##### COMPONENT NAME

* infra-images
* just-a-bunch-of-nodes config

##### ADDITIONAL INFORMATION
tested with our python virtualenv `ansible2.9-python3.6-2021-01-22`
with the provided playbook in readme.adoc
Also tested with just-a-bunch-of-nodes config


ansible/configs/just-a-bunch-of-nodes/default_vars.yml
```diff
@@ -7,14 +7,30 @@
...
-  ec2: "t2.medium"
+  ec2: "t3.medium"
   azure: Standard_A2_V2
   osp: 2c2g30d
 
-bastion_instance_image: RHEL75GOLD
+# Lookup images dynamically
+# see infra-images role
+node_instance_image:
+  owner: 309956199498
```
marcosmamorim, line 17: I recommend you to create a variable for the owner

ansible/roles-infra/infra-images/defaults/main.yaml
```diff
@@ -1,3 +1,86 @@
 ---
 # by default use the agnosticd 'instances' variable
 infra_images_instances: "{{ instances | default([]) }}"
+
+
+# Predefined search dicts for commonly used images
+infra_images_redhat_owner_id: 309956199498
```
tonykay, line 7: Should this also be a var, perhaps defaulted to `309956199498`?
And does it need to be extended in some way as it is very AWS/RedHat centric or is the assumption this is only a Red Hat thing? The role is generic or should it be an AWS-specific role?

`infra_images_redhat_owner_id: "{{ infra_images_owner | default('309956199498' }}"`

fridim: that variable is local to the default.yaml file, i don't expect it to be used outside of it.

It's just a convenience var, but i could have use the id all over the place when defining the infra_images_predefined mapping

tonykay (approved):
Looks good and I think will make the lad developer experience simpler

PR #4325, opened 2021-12-23
New parameter to force ODCR to a single AZ

fridim:
jira: GPTEINFRA-2334

New feature for the agnosticd_odcr module: force all AZs to be a single AZ.

Sometimes we want to provision all the instances into a single AZ (for saving costs for example).

In order to do that, one would have to redefine the reservations defined in the var `agnosticd_aws_capacity_reservations` to make it single-zone. Here is an example with ocp4-cluster.

From:

```yaml
agnosticd_aws_capacity_reservations:
  # Bastion can have its own AZ
  az1:
  - instance_type: "{{ bastion_instance_type }}"
    instance_count: 1
    instance_platform: "{{ bastion_instance_platform }}"

  masters:
  - instance_type: "{{ master_instance_type }}"
    instance_count: 3
    instance_platform: Linux/UNIX

  # Split workers in 2 AZs if possible.  Could be the same zone.
  workers1:
  # Workers: half of workers
  - instance_type: "{{ worker_instance_type }}"
    instance_count: >-
      {{ ( worker_instance_count | int / 2 )
      | round(0, 'ceil')
      | int }}
    instance_platform: Linux/UNIX
  workers2:
  - instance_type: "{{ worker_instance_type }}"
    instance_count: >-
      {{ ( worker_instance_count | int / 2 )
      | round(0, 'ceil')
      | int }}
    instance_platform: Linux/UNIX
```

to:

```yaml
agnosticd_aws_capacity_reservations:
  # Bastion can have its own AZ
  az1:
  - instance_type: "{{ bastion_instance_type }}"
    instance_count: 1
    instance_platform: "{{ bastion_instance_platform }}"

  - instance_type: "{{ master_instance_type }}"
    instance_count: 3
    instance_platform: Linux/UNIX

  - instance_type: "{{ worker_instance_type }}"
    instance_count: "{{ worker_instance_count }}"
    instance_platform: Linux/UNIX
```

To avoid the work of redefining the reservations in all the configs depending on whether the user wants a single AZ or spread across multiple AZ, we want a simple ON|OFF flag.

This PR, if applied, adds a new parameter to the `agnosticd_odcr` module to automatically group the reservations into one or keep the default behavior. That parameter can be controlled via the global variable `agnosticd_aws_capacity_reservation_single_zone` (false by default).
The advantage of doing so inside the module is that it should be transparent for the config (ocp4-cluster). No logic needed there.

If the parameter `single_zone` is set to true, all the reservation will target the same AZ.
All the reservations are merged into one AZ. Reservations that are the same `instance_type` and `instance_platform` are merged together and `instance_count` is incremented accordingly.

The result will include availability_zone value for the virtual zones, but they will all be the same.

- Add unit tests to helper functions to regroup all the reservation into
a single AZ
- test with ocp4-cluster
  - patch install-config.yml.j2 template to actually use odcr AZs
- update Readme.adoc

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION
Tested with:

* ansible2.9-python3.6-2021-11-30 python virtualenv
* ansible 2.9.27
* python 3.10.1

### DIFF
--- a/ansible/action_plugins/agnosticd_odcr.py
+++ b/ansible/action_plugins/agnosticd_odcr.py
@@ -42,7 +42,7 @@ DOCUMENTATION = """
             required: False
             type: bool
             default: False
-          all_in_one:
+          single_zone:
             description: If true, force all the zones to be the same. If false, do best-effort.
             required: False
             type: bool
@@ -195,7 +195,7 @@ def inject_reservation(l, *reservations):
 
 
 def regroup_reservations(reservations):
-    """Used when all_in_one is True. Regroup all the zones into one, named 'az1'."""
+    """Used when single_zone is True. Regroup all the zones into one, named 'az1'."""
 
     reservations = deepcopy(reservations)
     result = []
@@ -536,7 +536,7 @@ class ActionModule(ActionBase):
         'aws_secret_access_key',
         'ttl',
         'distinct',
-        'all_in_one',
+        'single_zone',
         'state'
     ))
     def run(self, tmp=None, task_vars=None):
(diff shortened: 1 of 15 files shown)


ansible/action_plugins/agnosticd_odcr.py
```diff
@@ -133,6 +139,72 @@ def check_response(response):
...
+    l = deepcopy(l)
+
+    for reservation in deepcopy(reservations):
+        for r in l:
+            if reservation_match(r, reservation):
+                r['instance_count'] = int(r['instance_count']) + int(reservation['instance_count'])
+                break
+            else:
+                continue
+            break
```
jkupferer, line 188: This break statement will never be hit I think? Can this logic be simplified to:
```
        for r in l:
            if reservation_match(r, reservation):
                r['instance_count'] = int(r['instance_count']) + int(reservation['instance_count'])
                break
```

jkupferer (approved):

PR #4334, opened 2022-01-06
Add alternative DNS names for instances

fridim:
instances[].dns_records   list of alt names for instances.

```
- name: node
  count: 3
  public_dns: true
  dns_records:
    - machine
    - server
```

Will create:

- node{1,2,3}.guid A records, using the public IP of each node
- machine{1,2,3}.guid CNAME records pointing to the corresponding A records
- server{1,2,3}.guid CNAME records pointing to the corresponding A records

##### ISSUE TYPE

- Feature Pull Request


tonykay (approved):
BTW have we established a best practice. I agree role vars should start `<role_name>_`

PR #4383, opened 2022-01-20
Fix to install collections into collections paths

jkupferer:
##### SUMMARY

Use collections path to determine where to install collections.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Automatic ansible-galaxy requirements resolution.

### DIFF
--- a/ansible/install_galaxy_roles.yml
+++ b/ansible/install_galaxy_roles.yml
@@ -33,6 +33,10 @@
              or (r_requirements_content is mapping
                  and 'roles' in r_requirements_content)
             )
+      register: r_ansible_galaxy_install_roles
+      until: r_ansible_galaxy_install_roles is successful
+      retries: 10
+      delay: 30
 
     - name: Install collections from requirements.yml
       command: >-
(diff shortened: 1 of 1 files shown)


tonykay (approved):
Elegant

ansible/install_galaxy_roles.yml
```diff
@@ -38,7 +38,7 @@
       command: >-
         ansible-galaxy collection install
         -r "{{ requirements_path }}"
-        -p collections
+        -p {{ lookup('config', 'COLLECTIONS_PATHS')[0] | quote }}
```
fridim, line 41: Is first collection paths always in user land?
No risk of being system path?

tonykay: Ah we saw concurrent colelction issues on admin recently. Ansible Workshops was installing into default `~/.ansible/collections` with 2 different versions. We saw lots of issues, race conditions. That has been resolved but a simple `ansible-galaxy install` without a `-p` should be avoided!

rut31337 (approved):

PR #4400, opened 2022-01-31
fix workload output for storage

rut31337:
##### SUMMARY
fix workload output for storage

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
roles/ocp4-workload-workshop-admin-storage

##### ADDITIONAL INFORMATION
workshop output was using debug not agnosticd_user_info module which does not populate babylon

```paste below

```


bbethell-1 (approved):

jkupferer (approved):
This doesn't need to be a loop, but will definitely work as written.

PR #4401, opened 2022-01-31
fix all user.info debug in live agnosticV CIs

rut31337:
##### SUMMARY
Change user.info output to use agnosticd_user_info module and not debug.  this is to make the info get sent to babylon which is the correct way.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp4-workload-ccnrd-cuttingedge
ocp4-workload-ccnrd-stable
ocp4-workload-debugging-workshop
ocp4-workload-jdg-workshop
ocp4-workload-openbanking-sandbox
ocp-workload-bxms-pam
ocp-workload-bxms-pam-foundations
ocp-workload-developer-environment
ocp-workload-rhte19-shadowman
ocp-workload-terminal
rosa
##### ADDITIONAL INFORMATION

```paste below

```


bbethell-1:
Please don't merge until a plan has been created: 

 really you can only test that PR after merging it. you would have to deploy those CIs affected in dev and have the owners do a quick test.
i wouldn't expect that they fail but since its several CIs we should be careful. then after they are all tested you would need to update the test/prod tags for each one

PR #4434, opened 2022-02-08
Add update entry point playbook

jkupferer:
##### SUMMARY

The update entry point playbook allows for updating existing
deployments. The update functionality must be implemented within the
config. An example for test-empty-config is provided.

Update functionality includes the new `agnosticd_user_data` lookup
plugin and `agnosticd_restore_output_dir` role so that updates can
fetch data items set during the initial provision.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

- Core update.yml entry-point playbook
- Config test-empty-config
- `agnosticd_user_data` lookup plugin
- `agnosticd_restore_output_dir` role

### DIFF
--- a/ansible/update.yml
+++ b/ansible/update.yml
@@ -1,0 +1,38 @@
+---
+# Entry point for update action.
+
+- name: Setup AgnosticD Runtime
+  import_playbook: setup_runtime.yml
+
+- name: Start environment
+  import_playbook: >-
+    {{ lookup('first_found', {
+         'files': [ 'start.yml', 'lifecycle.yml' ],
+         'paths': ['configs/' + env_type, playbook_dir]
+       })
+    }}
+  vars:
+    ACTION: start
+
+- name: Process Lifecycle Post Start Hook
+  vars:
+    __find_lifecycle_hook_params:
+      files:
+        - configs/{{ env_type }}/lifecycle_hook_post_start.yml
+        - configs/{{ env_type }}/lifecycle_hook.yml
+        - lifecycle_hook.yml
+  import_playbook: >-
+    {{ lookup('first_found', __find_lifecycle_hook_params) }}
+
+- name: Update deployment
+  vars:
+    __find_update_params:
+      files:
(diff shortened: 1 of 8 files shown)


bbethell-1 (approved):

ansible/update.yml
```diff
@@ -0,0 +1,38 @@
...
+
+- name: Start environment
+  import_playbook: >-
+    {{ lookup('first_found', {
+         'files': [ 'start.yml', 'lifecycle.yml' ],
+         'paths': ['configs/' + env_type, playbook_dir]
+       })
+    }}
+  vars:
+    ACTION: start
```
fridim, line 15: Should we make the update action atomic?
(Managing chaining or dependency of state outside of the action itself).
I would tend to say yes.

In the context of babylon, anarchy would make sure the state is "started" before running the update.
In the context of developing or running from the laptop, you don't update a stopped environment.

Also, is there a use-case where we don't need to "start" the environment to update it? Like a shared cluster which would always be on?

jkupferer: I would agree, but Ansible is slow and inefficient. If we schedule a start before update then this would likely add minutes to every user waiting for an environment and double the workload on Anarchy. So, while it would be more elegant to separate update from start, it is far more efficient to combine them.

fridim: I see.

Then i would argue the place to add the start would be inside the config, wdyt?

fridim: I think "started" is the final state we want to be after an "update"

But how we get there could depend on the config. Maybe the environment needs to be stopped in order to change the image of the instances, and then it's started.
So instead of:

1.  start the env
2. stop the env
3. change the AMI
4. start the AMI

it could just be 2,3,4. That's why i'm suggesting it's a config thing.

jkupferer: Ah, that's a fair point. I'll update the PR.

jkupferer: Updated

fridim (approved):

PR #4469, opened 2022-02-22
Update ocp4-cluster to optionally deploy an OCP on ARM cluster

wkulhanek:
##### SUMMARY

Updated ocp4-cluster config to allow for deployment of OpenShift on ARM64 (on AWS).

Additional environment variable: openshift_architecture with default x86_64. When set to arm64 an ARM cluster is being installed. Default instance types: m6g.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4-cluster
host-ocp4-installer

### DIFF
--- a/ansible/configs/ocp4-cluster/default_vars.yml
+++ b/ansible/configs/ocp4-cluster/default_vars.yml
@@ -91,7 +91,11 @@ worker_instance_count: 2
 
 # Install OpenShift 4 - and which version
 install_ocp4: true
-ocp4_installer_version: "4.5.15"
+
+# Use 4.9 for latest available release in the `stable-4.9` directory.
+# Use 4.9.9 for exactly the specified release
+ocp4_installer_version: "4.9"
+
 # Run logic to enable cluster shutdown before 24h initial certificate rotation
 # Only works for OCP 4.1 and 4.2. OCP 4.4.8 and later no longer require this.
 ocp4_enable_cluster_shutdown: false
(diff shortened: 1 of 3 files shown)


ansible/roles/host-ocp4-installer/templates/install-config.yaml.j2
```diff
@@ -7,6 +7,9 @@ fips: {{ ocp4_fips_enable | bool | to_json }}
 controlPlane:
   name: master
   hyperthreading: Enabled
+{% if openshift_architecture | default('x86_64') is match('arm64') %}
```
jkupferer, line 10: Let's put in the role defaults:
```
openshift_architecture: x86_64
```

Then cleanup all the repeated `| default(x86_64` ?

ansible/configs/ocp4-cluster/default_vars.yml
```diff
@@ -91,19 +91,31 @@ worker_instance_count: 2
...
+#   {{ 'https://mirror.openshift.com/pub/openshift-v4/aarch64/clients' if openshift_architecture | default('x86_64') is match('arm64')
+#   else 'https://mirror.openshift.com/pub/openshift-v4/clients'
+#   }}
+
 # CloudFront Mirror
-ocp4_installer_root_url: http://d3s3zqyaz8cp2d.cloudfront.net/pub/openshift-v4/clients
+ocp4_installer_root_url: >-
+  {{ 'http://d3s3zqyaz8cp2d.cloudfront.net/pub/openshift-v4/aarch64/clients' if openshift_architecture | default('x86_64') is match('arm64')
+  else 'http://d3s3zqyaz8cp2d.cloudfront.net/pub/openshift-v4/clients'
+  }}
```
jkupferer, line 115: Why is the default for `ocp4_installer_root_url` in the config defaults? Shouldn't this value be in the `host-ocp4-installer` role defaults?

wkulhanek: It's always been that way...  maybe something to tackle on AD 2.0?

jkupferer (approved):

ansible/roles/host-ocp4-installer/templates/install-config.yaml.j2
wkulhanek: Actually changed the variable to ocp4_architecture_cluster. Seems more consistent with variable naming. Will also allow us to add an ocp4_architecture_bastion in the future should we want to support ARM bastions.

PR #4575, opened 2022-03-31
Allow symmetric encryption of output_dir tarball

fridim:
Allow user to define a password to protect the output_dir in s3 in case
of a shared s3 bucket.

> agnosticd_save_output_dir_archive_password

Append '.gpg' extension in that case.

Use GNUPG tool to perform the symmetric encryption (by default it's
AES256), as it's the most available, even in images (it's more
available in images by default than openssl).

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

* agnosticd_save_output_dir
* agnosticd_restore_output_dir

### DIFF
--- a/ansible/roles/agnosticd_save_output_dir/tasks/create-output-dir-archive.yml
+++ b/ansible/roles/agnosticd_save_output_dir/tasks/create-output-dir-archive.yml
@@ -18,4 +18,16 @@
     # Disable warning for using tar command rather than archive module.
     # archive module does not support chdir type option.
     warn: false
+
+- when: agnosticd_save_output_dir_archive_password is defined
+  block:
+    - name: Encrypt tarball using password
+      command: >-
+        gpg --symmetric --batch --passphrase-fd 0
+        --output {{ (agnosticd_save_output_dir_archive_tempfile ~ '.gpg') | quote }}
+        {{ agnosticd_save_output_dir_archive_tempfile | quote }}
+      args:
+        chdir: "{{ output_dir }}"
+        stdin: "{{ agnosticd_save_output_dir_archive_password }}"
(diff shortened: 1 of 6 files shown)


clopnis:
great feature :+1:

jkupferer (approved):

ansible/roles/agnosticd_save_output_dir/tasks/create-output-dir-archive.yml
```diff
@@ -18,4 +18,16 @@
...
+  block:
+    - name: Encrypt tarball using password
+      command: >-
+        gpg --symmetric --batch --passphrase-fd 0
+        --output {{ (agnosticd_save_output_dir_archive_tempfile ~ '.gpg') | quote }}
+        {{ agnosticd_save_output_dir_archive_tempfile | quote }}
+      args:
+        chdir: "{{ output_dir }}"
+        stdin: "{{ agnosticd_save_output_dir_archive_password }}"
+        creates: "{{ agnosticd_save_output_dir_archive_tempfile }}.gpg"
```
jkupferer, line 32: Why a block with only one task?
Also, I don't think you want the command `args` `creates` here as that would prevent updating a `.gpg` file if somehow one happened to already exist.

fridim: Good point for the block, it used to be another task that i deleted since, leaving only one. Fixed.

fridim: for the `command.args.creates`,  i added them mainly because gnupg would fail if the file exist, and also because the files should always be absent anyway.

I removed them and added `--yes` to gpg commands to force overwrite.

jkupferer (approved):

PR #4657, opened 2022-04-20
ocp4-cluster config, support multiple zone-layout for ODCR

fridim:
##### SUMMARY
This change, if applied, allows more ways to arrange the clusters in different zones.

###### Before change

By default:
* 2 zones for workers
* 1 zone for masters
* 1 zone for bastion

Supported names for logical zones:
* workers1, workers2
* masters
* az1 for bastion

We can force zones to be distinct with `agnosticd_aws_capacity_reservation_distinct: true`.
It's possible to force everything into a single zone (costs effective, less HA) with `agnosticd_aws_capacity_reservation_single_zone: true`.

No flexibility to change the number of zone as the configs expects `masters` `workers1` and `workers2` logical zones.

It's always possible to change the layout by also changing the variables to specify how to use the new layout:

```yaml
openshift_controlplane_aws_zones_odcr
openshift_machineset_aws_zones_odcr
aws_availability_zone
```

But it's cumbersome for common use-cases.

###### After change

By default, same:
* 2 zones for workers
* 1 zone for masters
* 1 zone for bastion

It's now possible to change the layout a bit. Supported names for logical zones are now:
* workers, workers1, workers2
* masters, masters1, masters2, masters3
* az1 for bastion

 For example, this will now work:

* masters in the same zone   A
* workers in the same zone   B

Same as before: we can force zones to be distinct with `agnosticd_aws_capacity_reservation_distinct: true`.
Same as before: it's possible to force everything into a single zone (costs effective, less HA) with `agnosticd_aws_capacity_reservation_single_zone: true`.

###### Examples

This change also adds  examples to the readme of the reservation role, including examples that need to change the layout entirely. This can be done by overriding the variables that use the result of the role:

```yaml
openshift_controlplane_aws_zones_odcr
openshift_machineset_aws_zones_odcr
aws_availability_zone
```

##### ISSUE TYPE

- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

* infra-aws-capacity-reservation role (doc)
* ocp4-cluster config

### DIFF
--- a/ansible/configs/ocp4-cluster/pre_infra.yml
+++ b/ansible/configs/ocp4-cluster/pre_infra.yml
@@ -63,16 +63,22 @@
           aws_availability_zone: "{{ _r.az1.availability_zone }}"
           # masters
           # Support several logical mapping (masters, masters1-3)
-          openshift_controlplane_aws_zones_odcr:
-          - "{{ _r.masters.availability_zone  | default(None) }}"
-          - "{{ _r.masters1.availability_zone | default(None) }}"
-          - "{{ _r.masters2.availability_zone | default(None) }}"
-          - "{{ _r.masters3.availability_zone | default(None) }}"
+          openshift_controlplane_aws_zones_odcr: >-
+            {{
+              [ {{ _r.masters.availability_zone  | default(None) }},
+                {{ _r.masters1.availability_zone | default(None) }},
+                {{ _r.masters2.availability_zone | default(None) }},
+                {{ _r.masters3.availability_zone | default(None) }},
+              ] | select() | list
+            }}
           # workers
           # Support up to 4 zones
-          openshift_machineset_aws_zones_odcr:
-          - "{{ _r.workers.availability_zone  | default(None) }}"
-          - "{{ _r.workers1.availability_zone | default(None) }}"
-          - "{{ _r.workers2.availability_zone | default(None) }}"
-          - "{{ _r.workers3.availability_zone | default(None) }}"
-          - "{{ _r.workers4.availability_zone | default(None) }}"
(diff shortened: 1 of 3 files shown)


ansible/configs/ocp4-cluster/pre_infra.yml
```diff
@@ -56,13 +56,23 @@
...
-          aws_availability_zone: "{{ agnosticd_aws_capacity_reservation_results.reservations.az1.availability_zone }}"
+          aws_availability_zone: "{{ _r.az1.availability_zone }}"
           # masters
+          # Support several logical mapping (masters, masters1-3)
           openshift_controlplane_aws_zones_odcr:
-          - "{{ agnosticd_aws_capacity_reservation_results.reservations.masters.availability_zone }}"
+          - "{{ _r.masters.availability_zone  | default(None) }}"
+          - "{{ _r.masters1.availability_zone | default(None) }}"
+          - "{{ _r.masters2.availability_zone | default(None) }}"
+          - "{{ _r.masters3.availability_zone | default(None) }}"
```
jkupferer, line 70: It seems a bit awkward to have these `None` values here? Perhaps instead moving the `select` here?
```
      openshift_controlplane_aws_zones_odcr: >-
        {{ [
          _r.masters.availability_zone | default(None),
          _r.masters1.availability_zone | default(None),
          _r.masters2.availability_zone | default(None),
          _r.masters3.availability_zone | default(None),
        ] | select() | list }}
```

fridim: yeah i thought of that but i thought it was even more weird?

jkupferer (approved):

PR #4661, opened 2022-04-20
Adding missing libraries to copy/download images from/to osp/ibmcloud

agonzalezrh:


ansible/library/osp_copy_vm_disk_to_image.py
```diff
@@ -0,0 +1,195 @@
...
+             dest: myvm1-image
+           - src: myvm2
+             dest: myvm2-image
+'''
+RETURN = '''
+'''
+
+from ansible.module_utils.basic import *
+from ansible.module_utils.openstack import openstack_full_argument_spec, openstack_module_kwargs, \
+    openstack_cloud_from_module
```
jkupferer, line 61: I wonder if it is appropriate to use the openstacak module utils in Ansible core now that this has moved to a collection?
Should we develop against the collection style implementation, for example:
https://github.com/openstack/ansible-collections-openstack/blob/master/plugins/modules/image.py#L158

agonzalezrh: changed, thanks

jkupferer (approved):

PR #4718, opened 2022-05-11
Add collections_paths to root ansible.cfg as well (ansible/ansible.cf…

wkulhanek:
…g already had it correct)

##### SUMMARY

In order for dynamic installation of collections to work the root ansible.cfg needs to match the one in ansible/ansible.cfg. Both need a collections_path=/tmp/ansible-collections.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
AgnosticD Core

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -2,6 +2,7 @@
 nocows                  = 1
 # Don't forget to update other ansible.cfg files like:  ansible/configs/linklight/ansible.cfg
 roles_path              = dynamic-roles:ansible/dynamic-roles:roles-infra:ansible/roles-infra:roles:ansible/roles:ansible/roles_studentvm:roles_studentvm:ansible/roles_ocp_workloads:roles_ocp_workloads
+collections_paths       = /tmp/ansible-collections
 
 forks                   = 50
 become                  = False


rhjcd (approved):

tonykay (commented):
So on admin and admin-dev I know at least 1 config (aap2-ansible-workshops) relies on collections in `~opentlc-mgr/.ansible/collections` so concerned this will break stuff. Also has this been reconciled with what at least one call to `ansible-galaxy install` does [here](https://github.com/redhat-cop/agnosticd/blob/bfb6942dcec8fe13a31ca784b95c9d622199f65f/ansible/install_galaxy_roles.yml#L65)?

wkulhanek:
This is basically catching the /ansible.cfg up to /ansible/ansible.cfg. Pretty sure AD uses /ansible/ansible.cfg anyway.

fridim:
ansible/ansible.cfg is a link to ansible.cfg, maybe that's why you saw it after changing ansible.cfg?

I would be extra careful with that sort of change as this could impact a lot of provisions.
i'm not entirely sure of the impact for this one.

PR #4807, opened 2022-06-04
Fix package errors

wkulhanek:
##### SUMMARY

Revert use of package module to install things for vscode. Also fix one variable omission in ama demo.

##### COMPONENT NAME
ocp4_workload_vscode_bastion
ocp4_workload_ama_demo


wkulhanek:
@rut31337 @d-jana

d-jana (approved):

PR #4952, opened 2022-06-23
Add option to use own ssh keys in bastion-lite role

jfilipcz:
##### SUMMARY
This PR is a  partial revert of https://github.com/redhat-cop/agnosticd/pull/4534, with the aim to add back the option to use our own key pair in bastion-lite role. Only real difference from the previous approach is that now `use_own_key` is by default set to `false` .

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

bastion-lite role

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles/bastion-lite/tasks/main.yml
+++ b/ansible/roles/bastion-lite/tasks/main.yml
@@ -1,4 +1,16 @@
 ---
+- name: copy the user's SSH private key
+  become: true
+  copy:
+    src: "~/.ssh/{{key_name}}.pem"
+    dest: "/root/.ssh/{{key_name}}.pem"
+    owner: root
+    group: root
+    mode: 0400
+  when: not use_own_key|bool
+  tags:
+    - copy_env_private_key
+
 - name: Generate host .ssh/config Template
   become: false
   template:
--- a/ansible/roles/bastion-lite/templates/bastion_ssh_config.j2
+++ b/ansible/roles/bastion-lite/templates/bastion_ssh_config.j2
@@ -4,7 +4,11 @@ Host ec2* *.internal
 Host *.example.com
 {% endif %}	
   User {{remote_user | default(hostvars.localhost.remote_user) }}
-  IdentityFile ~/.ssh/{{ env_authorized_key }}.pem
+{% if use_own_key|bool %}
+  IdentityFile ~/.ssh/{{env_authorized_key}}.pem
+{% else %}
+  IdentityFile ~/.ssh/{{key_name}}.pem
+{% endif %}
   ForwardAgent yes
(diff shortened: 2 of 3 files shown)


ansible/roles/bastion-lite/templates/bastion_ssh_config.j2
```diff
@@ -4,7 +4,11 @@ Host ec2* *.internal
 Host *.example.com
 {% endif %}	
   User {{remote_user | default(hostvars.localhost.remote_user) }}
-  IdentityFile ~/.ssh/{{ env_authorized_key }}.pem
+{% if use_own_key|bool %}
+  IdentityFile ~/.ssh/{{env_authorized_key}}.pem
+{% else %}
+  IdentityFile ~/.ssh/{{key_name}}.pem
```
oybed, line 10: This logic seems flipped -no? The default should be to keep the existing behavior - so need to move `{{ key_name }}` to line 8 above (i.e.: if `use_own_key == true`)

ansible/roles/bastion-lite/tasks/main.yml
```diff
@@ -1,4 +1,16 @@
 ---
+- name: copy the user's SSH private key
+  become: true
+  copy:
+    src: "~/.ssh/{{key_name}}.pem"
+    dest: "/root/.ssh/{{key_name}}.pem"
+    owner: root
+    group: root
+    mode: 0400
+  when: not use_own_key|bool
```
oybed, line 10: should remove the `not` here, right?

oybed:
@wkulhanek the key name come from the AgnosticV side (i.e.: configured to use key xvz), and the "client" (for example Tower or any other automation) knows to use the correct key to connect (key kept in a secure place for client to fetch/use).

As @jfilipcz points out, this is mostly just to bring back previous behavior but with the default config to use new behavior, so mostly just a change to be able to continue the use of existing configurations and surrounding tooling that otherwise is broken with the latest version of AgnosticD.

wkulhanek:
@oybed In AgnosticV the following should be used these days:

ssh_authorized_keys:
- key: https://github.com/wkulhanek.keys

Of course that public key must exist on Github.

If I'm misunderstanding what you're trying to do I'll have to delegate to @fridim on approving.

oybed:
Yeah, so this is independent of having the SSH keys in github ... just bringing back some former functionality. 

Having said that, we had a chat on our side and there may be a better approach to this, so leaving this PR in `draft` for now and we'll come back with an update very soon (and potentially some other PRs).

oybed:
Replaced by PR https://github.com/redhat-cop/agnosticd/pull/5060

PR #5048, opened 2022-07-07
Updating with module name

edeandrea:
Updating module name `k8s` & `k8s_info` with full module names.

Also fixing `from_yaml_all` to `from_yaml_all | list`


rhjcd (approved):
lgtm

PR #5060, opened 2022-07-08
Add config to externally provide private SSH key used for provision and destroy at EC2

makirill:
##### SUMMARY

The `create_ssh_provision_key` infra role generates and store new SSH
key pair for each provisioning job. Plus setting following environment
variables:
- ssh_provision_pubkey_content
- ssh_provision_pubkey_path
- ssh_provision_key_path
- ssh_provision_key_name

In some cases this algorithm is not working well (e.g., the key pair can be lost
if not stored at persistent storage and destroy job will fail). This
change allow to bring your own permanent key by setting following
variables:
- ssh_provision_key_path
- ssh_provision_key_name
- ssh_provision_pubkey_content (optional)

And this provided key will be reused by provisioning and destroy jobs.

For example set following variables at AgnosticV:
```
ssh_provision_key_name: "my_private_ssh_key.pem"
ssh_provision_key_path: "/home/account/.ssh/{{ ssh_provision_key_name }}"
ssh_provision_pubkey_content: ssh-rsa AAAAB3NzaC1 ...rest of the key... JjQ==
```

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

- cloud_providers/ec2_destroy_env
- cloud_providers/ec2_infrastructure_deployment

- configs/hybrid-cloud/destroy_env_ec2
- configs/hybrid-cloud/destroy_env_ec2
- configs/ocp4-cluster/destroy_env_ec2

- roles-infra/infra-ec2-ssh-key

- roles/control-user
- roles/locate_env_authorized_key
- roles/set_env_authorized_key

##### ADDITIONAL INFORMATION
Environment provisioning and destroy was tested at AWS EC2 with externally provided SSH key and without it. Unfortunately I don't have environment to test the `control-user` role but it has exactly same change to save public SSH key form the `ssh_provision_pubkey_content` variable that was done for `set_env_authorized_key` role and this change works. But it still better to test it if there is such opportunity.

### DIFF
--- a/ansible/roles/locate_env_authorized_key/tasks/main.yaml
+++ b/ansible/roles/locate_env_authorized_key/tasks/main.yaml
@@ -19,8 +19,16 @@
   delegate_facts: true
   set_fact:
     env_authorized_key_path_pub: >-
-      {{ hostvars.localhost.env_authorized_key_path
-      | regex_replace('\.pem$', '') }}.pub
+      {%- if hostvars.localhost.ssh_provision_pubkey_path is defined -%}
+      {{ hostvars.localhost.ssh_provision_pubkey_path }}
+      {%- else -%}
+      {{ output_dir }}/{{ hostvars.localhost.ssh_provision_key_name | regex_replace('\.pem$', '') }}.pub
+      {%- endif -%}
+
+- name: Check if env_authorized_key_pub exists
+  stat:
+    path: "{{ env_authorized_key_path_pub }}"
+  register: env_authorized_key_pub_stats
 
 - name: Generate SSH pub key content if it doesn't exist
   shell: >-
@@ -30,6 +38,8 @@
     creates: "{{ hostvars.localhost.env_authorized_key_path_pub }}"
   delegate_to: localhost
   become: false
+  when:
+    - not env_authorized_key_pub_stats.stat.exists
 
 - name: Save SSH pub key content as fact
   delegate_to: localhost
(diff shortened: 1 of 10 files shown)


ansible/roles/locate_env_authorized_key/tasks/main.yaml
```diff
@@ -30,11 +38,17 @@
     creates: "{{ hostvars.localhost.env_authorized_key_path_pub }}"
   delegate_to: localhost
   become: false
+  when:
+    - not env_authorized_key_pub_stats.stat.exists
```
fridim, line 42: Isn't the `args.creates` parameter enough here?

fridim (commented):
can you update `docs/SSH_keys_and_access.adoc` too?
Like add a section about "providing and manage our own SSH key"?

ansible/roles/locate_env_authorized_key/tasks/main.yaml
makirill: Yes, `args.creates` is enough, fixed with 7aa99f3950649176357bc49008992c3de431fcee

makirill: Documentation updated with 394dadf3393b4ae03eefc8d27f4e5c8adc8b628e

fridim (approved):

fridim:
Can you fix the checks?

```
FAIL ........ yamllint ansible/configs/hybrid-cloud/destroy_env_ec2.yml
/home/runner/work/agnosticd/agnosticd/ansible/configs/hybrid-cloud/destroy_env_ec2.yml
  37:7      error    wrong indentation: expected 4 but found 6  (indentation)
```

PR #5068, opened 2022-07-11
Updated setup gitops file to add more edits to git repos

ritzshah:
##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


fridim (changes requested):
using lineinfile to work with yaml files seems wrong to me.

Can't you use a template instead to generate the files?

ritzshah:
The templates are already created but these are the ones I need to update the templates post cluster and gitea operator creation on the fly.

fridim (commented):
delegating review to Wolfgang

wkulhanek:
@fridim I'm afraid that's my pattern. :-( We have a Gitea repo with YAML Files that need to be updated for each individual user.

Maybe it _is_ time to look into Helm Charts instead...

wkulhanek (approved):

PR #5139, opened 2022-07-25
Convert Code Server Workload to GitOps

wkulhanek:
##### SUMMARY

Convert codeserver workload to deploy via Application/ApplicationSet using OpenShift Gitops.

This change *should* be transparent to any existing configurations as long as they have the gitops operator installed.

Speed increase for 10 instance of code server for 10 users (laptop deploy from the US to an existing cluster in eu-central-1):
- Old Workload: 3 minutes 40 seconds
- New Workload: 21.4 seconds

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4_workload_codeserver


ansible/roles_ocp_workloads/ocp4_workload_codeserver/tasks/workload.yml
```diff
@@ -25,55 +21,113 @@
...
+  #     owner: ec2-user
+  #     mode: 0664
+
+  - name: Install code server (multi user)
+    kubernetes.core.k8s:
+      state: present
+      definition: "{{ lookup('template', 'applicationset.yaml.j2') }}"
 
   - name: Save code server information for each user
     agnosticd_user_info:
```
newgoliath, line 126: JK wants switches on agonsticd_user_info to allow it to be shut off.

newgoliath (approved):

ansible/roles_ocp_workloads/ocp4_workload_codeserver/tasks/workload.yml
wkulhanek: Happy to add - once I understand what the point of that would be. I mean all information that we provide is necessary...

PR #5203, opened 2022-08-03
re-add ssh config to ocp on gcp

rut31337:
##### SUMMARY
re-add ssh config to ocp on gcp

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

gcp cloud provider

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/cloud_providers/gcp_ssh_config_setup.yml
+++ b/ansible/cloud_providers/gcp_ssh_config_setup.yml
@@ -78,4 +78,4 @@
   tasks:
     - name: add -F option ansible_ssh_extra_args
       set_fact:
-        ansible_ssh_extra_args: "{{ ansible_ssh_extra_args|d() }} -F {{ hostvars['localhost'].ansible_ssh_config }}"
+        ansible_ssh_extra_args: "{{ ansible_ssh_extra_args|default() }} -F {{ hostvars['localhost'].ansible_ssh_config }}"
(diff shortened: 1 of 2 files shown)


ansible/cloud_providers/gcp_ssh_config_setup.yml
```diff
@@ -0,0 +1,81 @@
...
+  gather_facts: false
+  any_errors_fatal: true
+  ignore_errors: false
+  tags:
+    - step001
+    - ssh_args
+  tasks:
+    - name: add -F option ansible_ssh_extra_args
+      set_fact:
+        ansible_ssh_extra_args: "{{ ansible_ssh_extra_args|d() }} -F {{ hostvars['localhost'].ansible_ssh_config }}"
```
jkupferer, line 81: Style nit pick:

We've been trying to get away from the `d()` syntax in favor of spelling out `default()`. Also I'm surprised to see that default can be called without an argument? The effect would be to empty string, right?

```
        ansible_ssh_extra_args: >-
          {{ ansible_ssh_extra_args | default('') }} -F {{ hostvars['localhost'].ansible_ssh_config }}
```

jkupferer (approved):

ansible/cloud_providers/gcp_ssh_config_setup.yml
rut31337: changed to default as requested.  I copied this from the existing common_ssh which already exists.  so that is where this all came from.   yeah i guess there would just be no extra args added if it was blank like -vvv etc.

PR #5297, opened 2022-08-23
Move OpenStack lifecycle to infra role

jkupferer:
##### SUMMARY

Change OpenStack lifecycle to use infra role.

This allows us to support use of the `openstack.cloud` collection.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

OpenStack lifecycle

### DIFF
--- a/ansible/configs/tech-desktop-901/requirements.yml
+++ b/ansible/configs/tech-desktop-901/requirements.yml
@@ -1,0 +1,4 @@
+---
+collections:
+- name: openstack.cloud
+  version: 1.7.2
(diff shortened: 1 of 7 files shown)


ansible/lifecycle.yml
```diff
@@ -34,7 +34,8 @@
       include_tasks: lifecycle_ec2.yml
 
     - when: cloud_provider == 'osp'
-      include_tasks: lifecycle_osp.yml
+      include_role:
+        name: lifecycle_osp.yml
```
fridim, line 38: wrong name

fridim (changes requested):

ansible/lifecycle.yml
jkupferer: Fixed, good catch

fridim (approved):

PR #5305, opened 2022-08-24
Switch contexts for bookbag deploy

treddy08:


ansible/roles_ocp_workloads/ocp4_workload_bookbag_ohc_demo/tasks/workload.yml
```diff
@@ -28,7 +31,6 @@
     - ./templates/bookbag-base-bc.yml.j2
 
 - name: Wait for workshop image {{ ocp4_workload_bookbag_ohc_demo_name }}:latest
-  ignore_errors: true
```
newgoliath, line 31: I needed this because the build was failing because of timeout sometimes.  Just leave it.  The builds continue and if they do fail, we'll see it later.  But mostly I found they do not fail, just timeout here.  No sense waiting.

newgoliath (approved):

PR #5318, opened 2022-08-25
fix lifecycle in ocp4-cluster

fridim:
Remove block
https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/ocp4-cluster/lifecycle_hook_post_start.yml#L25-L28

introduced by 4e1482ee829ee93abf2ebeed07d003cb983792a8

##### ISSUE TYPE

- Bugfix Pull Request


newgoliath (approved):

tonykay (commented):
Is there some code to replace this logic somewhere or just using default keys? Ignore, I should have realized where this code is, why would we generate keys on a start? Am I missing something. Isn't this a one time deploy thing

PR #5483, opened 2022-09-26
Increase ocp4_workload_pipelines wait for tkn-cli-serve

jkupferer:
##### SUMMARY

Increase wait for tkn-cli-serve in ocp4_workload_pipelines from 6m40s to 30m. Investigation of a recent failure showed that the deployment became ready around 20m after the failure. Not sure why it takes so long, but better to wait a bit longer than to fail.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Role ocp4_workload_pipelines


bosebc (approved):
Looks good. Thanks

PR #5555, opened 2022-10-18
Save output dir to s3 when OCP4 installer fails

fridim:
When the OCP4 installer fails, there is a `always` ansible block to fetch the installer log.

This change, if applied, adds a step to run the `agnosticd_save_output_dir` role so the installer logs will be found in the outputdir archive for our production environments when this happens.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
- host-ocp4-installer role


yvarbanov (approved):

d-jana (approved):
Good one

PR #5588, opened 2022-10-25
Tigera

gejames:
##### SUMMARY

TIgera changed the URL for their manifest files from a list of single URLs for each manifest to a tgz file containing all the manifest.  This caused the existing deployments to fail.  The unarchive module does not support gz files that are not a tar file, so I had to use the gtar command in a shell to unarchive the files.

Updted to set file permissions correctly

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_tigera_calico
agnosticd/ansible/roles/host-ocp4-installer/create-manifests.yml

```


gejames (commented):
fixed extra spaces

PR #5657, opened 2022-11-22
Fix kubevirt operator install

juliaaano:


wkulhanek (approved):
Weird. I have never seen a problem with that.

PR #5883, opened 2023-01-31
Use cluster TLS and add Argo CD RBAC for workshop user to each instance

gnunn1:
##### SUMMARY

Modifies the configuration of the user Argo CD instances for this workshop as follows:
- Uses edge termination instead of passthrough so it will use the wildcard TLS certificate, avoids problem of users having to accept a private certificate
- Adds RBAC to enable logging into Argo CD using the OpenShift credential instead of requiring users to lookup the Argo CD certificate
- Adds RBAC to all instances for the opentlc-mgr user so the workshop owner can easily assist users with problems.
- Disables login via the secret to avoid confusion

##### COMPONENT NAME

Changes to ocp4_workload_gitops_workshop role to support the OpenShift GitOps workshop

##### ADDITIONAL INFORMATION

Additional changes required in Workshop material, only merge into development for now (i.e. no test or prod environments)


christianh814 (approved):
lgtm

ahsen-shah (approved):

PR #5926, opened 2023-02-08
Add workflow to build ee-multicloud

fridim:
##### SUMMARY
Add a reusable workflow to build ee-multicloud public EE and push it to Quay.

Add a pull_request workflow to automatically build an image with tag  pr-NUM  which expires after 1d.
PR workflow is triggered with path `tools/execution_environments/ee-multicloud-public/**`

use https://github.com/redhat-actions/buildah-build
and https://github.com/redhat-actions/push-to-registry  Actions.

Convert the ee-multicloud-public EE as a simple Dockerfile using ubi8 image.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


bbethell-1 (approved):

tonykay (approved):
There are some optimizations to the build process we could add, but its probably best to focus on getting it working and then getting it optimized. Then we can add multi-arch like `--platform linux/arm64/v8,linux/amd64` again later

PR #5928, opened 2023-02-09
Update main.yml

dwojciec:
I changed 5 to 50 to have 50 users minimum. https://issues.redhat.com/browse/GPTEINFRA-5367

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New config Pull Request
- New role Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


fridim (changes requested):
I may miss some context here but this is generally done in agnosticV at the catalog item level.
I would suggest to update that there instead.

PR #5944, opened 2023-02-13
[infra-vmc-resources] Delete project even if is empty.

agonzalezrh:
When a provision fails, the cleanup was not deleting the project because the folder was empty.


bbethell-1 (approved):
LGTM

fridim (approved):

PR #5994, opened 2023-02-22
EE build: Add retries to galaxy commands

fridim:
##### SUMMARY
Add retry to tasks that fail regularly because galaxy is flapping.

##### ISSUE TYPE

- Bugfix Pull Request


rut31337 (approved):

fridim:
Not implementing this for now as galaxy seems back online.
I would rather keep things simple in the containerfile.

PR #6027, opened 2023-03-01
bugfix for sshpass

YogiSoni-lazy:
Bugfix Pull Request

##### COMPONENT NAME
osp17-director-deployment fixed the sshpass plaintext password in test.yaml


bbethell-1 (approved):

bbethell-1:
@fridim  is this secure?

fridim:
While it would be "secure", it would still be a static password, even if protected by gpg on this repo.

Best is to generate the password using ansible, when provisioning.

```yaml
-   set_fact:
        undercloud_ssh_password: "{{ lookup('password', '/dev/null length=18 chars=ascii_letters,digits') }}"
```

and return it to user if needed using `agnosticd_user_info`


```  agnosticd_user_info:
    data:
      undercloud_ssh_password: "{{ student_password }}"
```

Print it back to the user or not depends on the lab/demo.

bbethell-1:
@YogiSoni-lazy  
Please can you adjust it to fit into Guillaume's suggestion
Thank you @fridim

YogiSoni-lazy:
@fridim in this specific lab, there are total 6 vms provided with static passwords i.e r3dh4t1! for internal connectivity. I would like to highlight that learner is not going to take access of these vms and login using workstation vm that would be having password generated dynamically. 

So, do I need to generate all passwords dynamically? and if yes, then any reason why?

fridim:
If the VMs are accessed internally, the usual pattern to provide access to them is to automate the creation of an SSH key on the workstation (or bastion) and to add it as authorized_key on the other instances.
That way, the student will not have to type passwords when SSH'ing from the workstation or bastion.

You can also re-use the provision SSH key for that purpose.

PR #6032, opened 2023-03-02
Adds sushy restart, firewalld disabled and new commit id for kcli gh

alosadagrande:
SUMMARY
Adds extra checks:
- When installing sushy-tools a restart is required if it is going to be executed twice.
- Disables firewalld since it is mandatory to be disabled. If I am right, firewalld, it is not enabled by default in RHPDS but anyway just in case it is enabled later on.
- Updating the commid id for an external repo we are using.
- Mounting an extra disk on /var/lib/libvirt since the installation by default does not allow selecting the proper disk size for the root filesystem. Since we are running multiple VMs the /var/lib/libvirt folder it gets exhausted easily. The task selects a disk > 500GB which is not in use.
- Adds a wait for the registry to be running before trying to log into it.

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
ocp4_workload_5gran_deployments_lab

ADDITIONAL INFORMATION

cc/ @mvazquezc

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
@@ -143,6 +143,12 @@
   ansible.builtin.shell:
     cmd: /tmp/deploy-sushy-tools.sh
 
+- name: Ensure sushy tools are running
+  ansible.builtin.systemd:
+    state: restarted
+    enabled: true
+    name: sushy-tools
+
 - name: Ensure disconnected registry dependencies are installed
   ansible.builtin.dnf:
     name: podman, httpd-tools
@@ -197,6 +203,13 @@
 - name: Ensure trusted certs are updated
   ansible.builtin.shell:
     cmd: update-ca-trust
+
+- name: Ensure firewalld is disabled
+  ansible.builtin.systemd:
+    state: stopped 
+    masked: true
+    enabled: false
+    name: firewalld
 
 - name: Ensure registry is running by login into it
   containers.podman.podman_login:
--- a/ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
@@ -13,6 +13,49 @@
   ansible.builtin.pip:
     name: kubernetes, passlib
 
+- name: Find an extra disk for libvirt folder
+  set_fact:
+    _disks: "{{ _disks|default([]) + [ item ] }}"
+    _diskname: "{{ '/dev/' + item.key }}"
+  when:
+    - not item.value.partitions
+    - item.key is search ("nvme*")
+    - min_value|int <= item.value.sectors|int * item.value.sectorsize|int
+    - extra_disk_libvirt_images is true
+  vars:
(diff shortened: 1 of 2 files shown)


d-jana (approved):

ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
```diff
@@ -198,6 +243,13 @@
   ansible.builtin.shell:
     cmd: update-ca-trust
 
+- name: Ensure firewalld is disabled
```
d-jana, line 246: May I why this workload need disable firewall?

d-jana (dismissed):

d-jana:
@fridim @yvarbanov Is this look right? Looking for review on this.

ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
alosadagrande: We are creating KVM VMs and if it is enabled it does not allow the VMs to access the registry, DNS and IPMI interfaces (sushy) running in the hypervisor.

fridim (dismissed):

ansible/roles_ocp_workloads/ocp4_workload_5gran_deployments_lab/tasks/pre_workload.yml
fridim: for what instances are you disabling firewalld?

Could you allow the traffic instead of disabling everything? Are those VM exposed in a way?

alosadagrande: > for what instances are you disabling firewalld?

The baremetal server hosting the VMs 

> Could you allow the traffic instead of disabling everything? Are those VM exposed in a way?

I can take a look, but I think by default it is already stopped in equinox. I include that task just in case in the future is enabled and for my testing on my lab. No, VMs are not exposed are under a virtual NATed network.

fridim: @hgeaydem  any opinion on this?

mvazquezc: Just FYI, the equinix base environment we were using didn't have firewalld/iptables installed. So this change does not have any impact on rhpds environments. This is targeting other envs outside RHPDS since we plan to use the playbook outside rhpds as well...

fridim: sorry we don't mean to block your work, but this catches the attention.

If the firewalld is not there, then don't disable it.

If it''s there in another context, the firewall is probably there for a reason and you should consider allowing the specific rules instead of disabling it entirely?

alosadagrande: Ok, I commented out the firewalld task so we can continue testing the lab. Meanwhile we are going to take a look at the network flows that need to be enabled. Makes sense? @fridim

fridim (approved):

agonzalezrh (approved):

ahsen-shah (approved):

ahsen-shah (approved):

ahsen-shah (approved):

PR #6055, opened 2023-03-07
ee-multicloud: Use latest for collections and rm receptor

fridim:
Since we have a detailed changelog now, use latest versions of Galaxy Collections. The versions of the collecions for each version of the Execution Environment will be properly documented.

Remove receptor as it should not be needed.
See this comment: https://github.com/ansible/awx-ee/pull/33#pullrequestreview-582446299 Removing the x86_64 binary will:
- facilitate the creation of an image for ARM.
- reduce size

##### ISSUE TYPE

- EE update


fridim:
* [ee-report diff with v0.0.11](https://gist.github.com/ee07c4e272022c9df37e53326fdfb0cd)

tonykay (approved):

PR #6275, opened 2023-04-19
Fix acc app dev vars and add gitea to user page

juliaaano:


juliaaano:
@bbethell-1 can you please also merge this? Thank you.

bbethell-1 (approved):

PR #6317, opened 2023-04-25
Add loginctl command to vscode-server role to avoid podman moaning

ericzolf:
##### SUMMARY

Because the VScode terminal isn't a login session, it can be that podman complains about no systemd user session being available. loginctl enable-linger solves the issue

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
vscode_server

##### ADDITIONAL INFORMATION
tested with advanced automation controller


bbethell-1 (approved):
LGTM

PR #6487, opened 2023-05-18
implement bookbag in ocp4-cluster

newgoliath:


jkupferer (changes requested):
This doesn't look right...

PR #6508, opened 2023-06-09
Add support for `common_extra_packages`

wkulhanek:
##### SUMMARY

Specifying `common_extra_packages` allows installation of extra packages without having to override the common_packages_el9 or common_packages_el8 variables.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
common

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -75,6 +75,7 @@
 - name: Install extra packages
   when:
   - not hostvars.localhost.skip_packer_tasks | default(false)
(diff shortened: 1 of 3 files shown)


ansible/roles/common/tasks/main.yml
```diff
@@ -66,8 +66,17 @@
 - name: Install common packages for RHEL 9
   when:
   - ansible_distribution_major_version == "9"
```
fridim, line 68: Should we add a condition for RHEL?
Like "debian 9" would trigger the execution of those tasks too?

fridim (approved):

PR #6675, opened 2023-07-14
Fix azure destroy

rut31337:
##### SUMMARY
Fix azure destroy to only run subscription based destroy functions when running the subscription based env type.
The destroy functions for subscription based destroy are not compatible with resource group based env types.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

./cloud_providers/azure_destroy_env.yml

##### ADDITIONAL INFORMATION

```paste below

```


bbethell-1 (approved):
Thanks Patrick

tonykay (approved):

PR #6990, opened 2023-09-06
add GroupName property

### LINKED ISSUE #6947: Request - Add Security Group Name Property
https://github.com/redhat-cop/agnosticd/blob/e7a1321e70807f039e5a800cb8131e00dd523fda/ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2#L78C12-L78C35

Add the GroupName Property to match the Name tag otherwise the actual name includes a random hash

willtome:
##### SUMMARY
Fixes #6947

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
infra-ec2-template-generate role

### DIFF
--- a/ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
+++ b/ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
@@ -1,4 +1,5 @@
 #jinja2: lstrip_blocks: "True"
+#touch
 ---
 AWSTemplateFormatVersion: "2010-09-09"
 Mappings:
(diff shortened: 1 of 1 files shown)


ansible/roles-infra/infra-ec2-template-generate/templates/cloud_template.j2
```diff
@@ -78,6 +78,7 @@ Resources:
     Type: "AWS::EC2::SecurityGroup"
     Properties:
       GroupDescription: Host
+      GroupName: "{{security_group['name']}}"
```
fridim, line 81: Commented on the issue 

> Can you elaborate why having a random hash for GroupName is an issue?

willtome: @fridim since the hash is not a derivative of the GUID or any other identifier, there is no way to publish in the user details the user the name or id of the security group to use for subsequent automation in the open environment. Unlike other AWS resources such as an instance, keypair, or subnet, the "Name" tag does not actually represent the Name of the object so it cannot be referenced with only the tag.

fridim (approved):

PR #7037, opened 2023-09-18
Rhel91 summit 2023

xtophd:
##### ISSUE TYPE

Feature enhancement which will allow podman desktop to be deployed on on node1
Save some build info in /root/RHEL9-Workshop/config so hostnames, ip addresses can be pasted into config files (ie: novnc)


klewis0928 (approved):
LGTM

PR #7049, opened 2023-09-21
New config: rosa-consolidated. Deploys Rosa in all ways

wkulhanek:
##### SUMMARY

This new config is supposed to replace the two existing configs rosa and rosa-manual.

Lots of cleanup and error management got added.

Capabilities:
- Set up a bastion VM that is ready to install ROSA (e.g. for an ILT or MOBB workshop)
- Set up a bastion VM and use it to install ROSA

ROSA installation has been changed from classic to STS
This config also supports deploying using hosted control planes - although given that HCP is tech preview this is a bit ... random if it succeeds or not

This config should be able to replace existing ones without modifications.

##### ISSUE TYPE
- New config Pull Request

##### COMPONENT NAME
rosa-consolidated


klewis0928 (approved):
LGTM

PR #7059, opened 2023-09-22
SHOWROOM: remove unnec password; make generated_password option

newgoliath:


bbethell-1 (approved):

tonykay (commented):
So, IMO, this will break non Azure use case or cases where `generated_password` is not set ie on ec2 there will be no known password for the `ansible_user` so the become will fail?

PR #7099, opened 2023-10-02
Work around ansible bug on length of vaulted string

jkupferer:
##### SUMMARY

- The `length` filter applied to a vaulted value produces an error "object of type 'AnsibleVaultEncryptedUnicode' has no len()"
- Use equivalent test for empty string.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Role host-ocp4-provisioner


klewis0928 (approved):
lgtm

PR #7111, opened 2023-10-03
[ee-multicloud-public] Update entrypoint script

agonzalezrh:
To fix issues with the dynamic collection installation and set variable /home/runner

### DIFF
--- a/tools/execution_environments/ee-multicloud-public/Containerfile
+++ b/tools/execution_environments/ee-multicloud-public/Containerfile
@@ -86,7 +86,7 @@ RUN for dir in \
 
 ENV HOME=/home/runner
 
-ADD https://raw.githubusercontent.com/ansible/ansible-builder/release_3.0/src/ansible_builder/_target_scripts/entrypoint /usr/local/bin/entrypoint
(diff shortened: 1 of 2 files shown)


tools/execution_environments/ee-multicloud-public/Containerfile
```diff
@@ -86,7 +86,7 @@ RUN for dir in \
 
 ENV HOME=/home/runner
 
-ADD https://raw.githubusercontent.com/ansible/ansible-builder/release_3.0/src/ansible_builder/_target_scripts/entrypoint /usr/local/bin/entrypoint
+ADD https://gpte-public-documents.s3.amazonaws.com/ansible_ansible-builder_release_3.0_src_ansible_builder__target_scripts_entrypoint.txt /usr/local/bin/entrypoint
```
fridim, line 89: Can you please add the file directly here and use `COPY`?

fridim (changes requested):
Let's add comments explaining exactly why we need a separate entrypoint.
Could you add it?

fridim (approved):

PR #7152, opened 2023-10-10
OCP4 AIO Equinix - Setting version to main until we can get a new tag

bbethell-1:
Will need to change to a new tag one @agonzalezrh  is back tomorrow


d-jana (changes requested):
can you create a tag and use that, rather than using a branch here to make PROD more stable

bbethell-1:
> can you please create a tag and use that, rather than using a branch here to make PROD more stable?
> 

I don't have permission to create the branch as its on Albertos repo

bbethell-1:
Aim to create one once Alberto returns tomorrow

d-jana:
@agonzalezrh Can you please create tag on this repo so that we can use it in AgnosticD?

d-jana (approved):

PR #7224, opened 2023-10-20
Fix bookbag create project bug

jkupferer:
##### SUMMARY

Ansible attempts to patch the ProjectRequest after it is created and gets a 403 error. The namespace was already created by project request so this error can be safely ignored.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

bookbag role


ashokjammula1 (approved):
looks fine

PR #7254, opened 2023-10-26
Update workload.yml create gitops sub task

rhjcd:
##### SUMMARY

Update workload.yml create gitops sub task  because it isn't found when deployed due to syntax

##### ISSUE TYPE
- Bugfix Pull Request


ashokjammula1 (approved):
looks fine

PR #7278, opened 2023-10-31
[host-ocp4-assisted-installer] Reduce DV size + dont remove DV

agonzalezrh:
It will help for troubleshooting in case of issues with the installation ISO


klewis0928 (approved):
LGTM

PR #7374, opened 2023-11-15
remove firewall rules before running ocp uninstall

rut31337:
##### SUMMARY
For some reason OCP creates firewall rules in GCP that the OCP installer cannot destroy causing the destroy to fail.  This will fix that problem.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ocp4-cluster

##### ADDITIONAL INFORMATION

```paste below

```


agonzalezrh (approved):
lgtm

PR #7443, opened 2023-11-27
[host-lets-encrypt-certs-certbot] Improve path for credentials file

agonzalezrh:
##### SUMMARY

Currently is searching for /home/user/credential instead the HOMEUSER/credential, for example for /root/

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Role host-lets-encrypt-certs-certbot

### DIFF
--- a/ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
+++ b/ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
@@ -9,7 +9,7 @@
   block:
     - name: Verify if AWS Credentials exist on the host
       stat:
-        path: "/home/{{ _certbot_user }}/.aws/credentials"
+        path: "{{ ('~' +_certbot_user + '/.aws/credentials') | expanduser }}"
       register: aws_credentials_result
 
     - name: Fail if AWS Credentials are not on the host
@@ -22,7 +22,7 @@
   block:
     - name: Verify if Azure Credentials exist on the host
       stat:
-        path: "/home/{{ _certbot_user }}/.azure.ini"
+        path: "{{ ('~' +_certbot_user + '/.azure.ini') | expanduser }}"
       register: azure_credentials_result
 
     - name: Fail if Azure Credentials are not on the host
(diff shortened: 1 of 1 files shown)


bbethell-1 (approved):

ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
```diff
@@ -9,7 +9,7 @@
   block:
     - name: Verify if AWS Credentials exist on the host
       stat:
-        path: "/home/{{ _certbot_user }}/.aws/credentials"
+        path: "{{ ('~' +_certbot_user + '/.aws/credentials') | expanduser }}"
```
jkupferer, line 12: I think `expanduser` is a mistake here. I believe it will evaluate the current user, `~` , on localhost because all filters run on localhost.

jkupferer (changes requested):

ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
agonzalezrh: you are right, let me solve it

ansible/roles/host-lets-encrypt-certs-certbot/tasks/main.yml
```diff
@@ -4,12 +4,17 @@
...
+    database: passwd
+    key: "{{ _certbot_user }}"
+
 - name: Check on AWS credentials
   when: _certbot_dns_provider is match('route53')
   block:
     - name: Verify if AWS Credentials exist on the host
       stat:
-        path: "/home/{{ _certbot_user }}/.aws/credentials"
+        path: "{{ ansible_facts.getent_passwd[_certbot_user][-2] + '/.aws/credentials' }}"
```
fridim, line 17: Probably only philosophical but i would do instead: 

`ansible_facts.getent_passwd[_certbot_user][4] `

That's how i think backward compatibility could work. `getent` will never change the order but they might add more values

fridim (approved):

PR #7525, opened 2023-12-18
Update MAD with a new module 7 that includes RHDH

varodrig:
##### SUMMARY

A new module is added to the MAD workshop to include Red Hat Developer Hub. This new module is module 7.
This PR includes the RHDH setup that's connected to the Helm charts repository and additional configurations to make it available.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4_workload_mad_roadshow

##### ADDITIONAL INFORMATION

```


ansible/roles_ocp_workloads/ocp4_workload_mad_roadshow/tasks/rhdh/rhdh-main.yml
danieloh30, line 1: Any specific reason to create the 2nd level directory (rhdh)? For structure consistency, those 2 YAML might be better to exist at the same level. Or you can create a separate role to deploy the RHDH then the MAD role just includes the roles like MTA, Gitea, etc.

treddy08 (commented):
I see you're using the prefix ocp4_workload_redhat_developer_hub for variables in your AgV config without actually using the ocp4_workload_redhat_developer_hub role in your infra_workloads.  Standard practice is to use the prefix along with the role that it relates to.

ansible/roles_ocp_workloads/ocp4_workload_mad_roadshow/tasks/rhdh/rhdh-main.yml
varodrig: Yes, the reason is to group everything on RHDH since this is not an operator like MTA. RHDH will be include many operators/configs/plugins that are part of the setup/configuration.

danieloh30: That's exactly what I meant. If RHDH installation is required a lot of stuff, it should be managed by separated roles. MTA and Gitea are also separated roles that MAD workload just invokes. Btw, I see only 2 yaml files under the RHDH directory so I think it would be better not to create a directory though.

danieloh30: @varodrig @wkulhanek

danieloh30: https://github.com/redhat-cop/agnosticd/tree/development/ansible/roles_ocp_workloads/ocp4_workload_mta6
Take a look at the MTA role which is not just an operator installation :)

treddy08 (approved):

bbethell-1 (approved):

ansible/roles_ocp_workloads/ocp4_workload_mad_roadshow/tasks/rhdh/rhdh-main.yml
varodrig: discussed over email.

danieloh30: As discussed, make sure to remove the sub folder before we start the beta test :)

PR #7581, opened 2024-01-17
Upgrade kubernetes.core rosa config

juliaaano:


wkulhanek (changes requested):
This config is defunct. Please use `rosa-consolidated` instead.

PR #7601, opened 2024-01-24
remove emitting generated_password when there's no azure bastion

newgoliath:
Azure bastions require a "generated_password".
But when there's no bastion in the env_type, then there's no generated password, and a failure.


klewis0928 (approved):
LGTM

PR #7610, opened 2024-02-07
[roadshow-ocpvirt] Update post_infra.yml

agonzalezrh:
##### SUMMARY
Using Route53 throws the following error when limit is reached:
botocore.exceptions.ClientError: An error occurred (Throttling) when calling the ListResourceRecordSets operation (reached max retries: 4): Rate exceeded

This code it will retry every 60 seconds, 10 times

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
roadshow-ocpvirt config


ashokjammula1 (approved):
looks good to me

PR #7633, opened 2024-02-20
Revert "Update workload.yml to upgrade IPA server"

d-jana:
Reverts redhat-cop/agnosticd#7631

As actual code updated on helm-charts: https://github.com/redhat-cop/helm-charts/pull/463, we don't need this one.


bosebc (approved):
Looks good. Thanks

bbethell-1 (approved):

PR #7638, opened 2024-02-21
zero-touch-base-rhel Update pre_software.yml to configure ssh for rhel user

wilson-walrus:
##### SUMMARY

Adds ssh configuration for the "rhel" user

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

zero-touch-base-rhel

##### ADDITIONAL INFORMATION

This change was enabled to fix issues with showroom not being able to ssh as the rhel user

```paste below

```


tonykay (approved):
I think we should move this into the role (in fact this used to work), but in the interests of getting this working and in prod approved.

klewis0928 (commented):
LGTM

PR #7749, opened 2024-03-28
Add showroom destroy and deploy tasks to ansible

tonykay:
##### SUMMARY

Added the when: protected option to use Shared Showroom OCP in config ansible-lightspeed

- Allows it to be used with Showroom while simplifying port issues around CNV

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

configs/ansible-lightspeed


newgoliath (approved):
Shared cluster!

PR #7807, opened 2024-04-08
[WIP] object convert to template

miteshget:
##### SUMMARY
RHACS Certral custom resource task converted to template

Reason: In RHACS 4.4 there is new feature  scannerV4 is available and it is disabled by default. So to enable it we have to add option with condition which doesn't effect existing or old CIs.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Role: ocp4_workload_rhacs


wkulhanek (approved):
@miteshget Just in time - I need that for another env as well. :-)

PR #7851, opened 2024-04-15
Adding ocp4-disconnected configs (try 2)

kincl:
##### SUMMARY

Adds:
- feature: ocp4-disconnected config
- feature: not running default cloudformation in EC2
- bugfix: remove double reporting showroom variables

##### ISSUE TYPE

- Feature Pull Request
- New config Pull Request

##### COMPONENT NAME

ocp4-disconnected


kincl:
Tried again with the deveopment branch as the source to see if we can fix the labeler

newgoliath (approved):

kincl:
I also changed the default branch back to `development` in my fork

bbethell-1 (approved):

ansible/cloud_providers/ec2_infrastructure_deployment.yml
```diff
@@ -58,6 +58,7 @@
         name: infra-ec2-template-create
       vars:
         aws_region_loop: "{{aws_region}}"
+      when: ec2_default_cloudformation | default(true) | bool
```
fridim, line 61: that variable name looks wrong.
The role takes the cloudformation template from the *config* if it exists

The variable name suggests it would not do that if such a template exists.

fridim (changes requested):

ansible/cloud_providers/ec2_infrastructure_deployment.yml
fridim: We can think of a variable to enable/disable the infra-ec2-template  roles all together, something more generic.

maybe  `infra_cloudformation_enabled` with a  default=true?

kincl: I am trying to disable the entire deployment of cloudformation for an ec2 environment and use a cloudformation template that is located in my lab guide repo in `ansible/configs/ocp4-disconnected/post_infra.yml`. Is there a better variable name for "disable all default cloudformation in the sandbox"?

kincl: Gotcha, I changed it to `cloudformation_enabled`

fridim (commented):
I suggest we change the variable name to `cloudformation_enabled`
I'm changing my review to "comment" so it's not blocking

PR #7854, opened 2024-04-15
New role to import userdata dictionary from agnosticV definitions

wilson-walrus:
##### SUMMARY

A new role allowing definitions of user data in agnosticV to be imported at any point during the deployment. The passthrough data is run during the setup_runtime.yml before any variables in other roles are defined. This role will set the data at any point defined in the config.

##### ISSUE TYPE

- New role Pull Request

##### COMPONENT NAME

agnosticv-userdata-import

##### ADDITIONAL INFORMATION

```paste below

```


klewis0928 (approved):
LGTM

PR #7909, opened 2024-04-18
final changes for LB1374

newgoliath:
- **put a when condition on deploying cnv lab**
- **add custom MOTD**

I picked up lots of commits in my rebase, but things are fine.


wilson-walrus (approved):
Looking good

PR #8000, opened 2024-05-02
Fix CUDA compile time issues for llama-cpp

tonykay:
##### SUMMARY

RHEL has compilation issues building llama-cpp with CUDA support, which Fedora does not
Versions?

This brings it more in line with how to build for CUDA

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

roles/ai_setup_ilab


ashokjammula1 (approved):
LGTM

PR #8013, opened 2024-05-06
Add sleep to experience-ocpvirt shell script

jkupferer:
##### SUMMARY

This shell script really needs to be rewritten to do all this in ansible so we can do normal retry logic.

Right now this is just a quick fix to allow apis to come up for webhooks to give us a better chance of success.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Config experience-ocpvirt


ashokjammula1 (approved):
LGTM

PR #8247, opened 2024-07-02
convert-to-rhel add role and update pre_software to pull in correct r…

wilson-walrus:
…epos

##### SUMMARY

Adding a role to resolve CentOS7 repos. The base repos configured on the system are no longer valid and must point directly to the new location. This role modifies the repo source definitions to point directly to the new location instead of using mirrors.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

convert-to-rhel

##### ADDITIONAL INFORMATION

```paste below

```


klewis0928 (approved):
LGTM

PR #8329, opened 2024-07-19
ocp showroom: implementation of stacked/tabbed terminals

newgoliath:
stacked and tabbed terminals working with showroom terminal and wetty


wkulhanek (approved):
/LGTM

PR #8355, opened 2024-07-25
Cloud front to mirror http://mirror.openshift.com/pub/openshift-v4/cl…

bbethell-1:
Change the broken cloudfront URL to point to mirror openshift instead


bosebc (approved):
Looks good. Thanks

PR #8376, opened 2024-08-01
Add logic to use pvcs instead of snapshots for os images, add support to pre-delete VMs on CNV to ocp4-cluster

wkulhanek:
##### SUMMARY

To fix the issue with dangling VolumeSnapshots:

- Add logic to ocp4_workload_kubevirt to use pvcs instead of snapshots for OS Disk Images
- Add support to ocp4-cluster for CNV to pre-delete all VMs on the cluster to also clean up snapshots that may have been created by users.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4-cluster
ocp4_workload_kubevirt

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_kubevirt/templates/hyperconverged.yaml.j2
+++ b/ansible/roles_ocp_workloads/ocp4_workload_kubevirt/templates/hyperconverged.yaml.j2
@@ -11,7 +11,7 @@ spec:
       tolerations:
         {{ ocp4_workload_kubevirt_workload_tolerations | to_yaml }}
 {% endif %}
-{% if ocp4_workload_kubevirt_import_boot_sources | bool %}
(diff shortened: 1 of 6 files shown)


bbethell-1 (approved):

ansible/roles_ocp_workloads/ocp4_workload_kubevirt/templates/hyperconverged.yaml.j2
```diff
@@ -11,7 +11,7 @@ spec:
       tolerations:
         {{ ocp4_workload_kubevirt_workload_tolerations | to_yaml }}
 {% endif %}
-{% if ocp4_workload_kubevirt_import_boot_sources | bool %}
+{% if ocp4_workload_kubevirt_boot_sources_shapshot | default(true) | bool %}
```
jkupferer, line 14: No need for `default` here because it is defined in `defaults/main.yml`

jkupferer (approved):

PR #8412, opened 2024-08-12
Testing new GCP module

klewis0928:
- Bugfix Pull Request

##### COMPONENT NAME
open-environment-gcp

##### ADDITIONAL INFORMATION

```paste below

```


wkulhanek (commented):
I don't think this does anything other than print a warning if the EE doesn't have a matching version. If you need newer GCP you probably want a newer EE.

The EE tagged 2024-08-08 already has 1.3.0 in it.

PR #8500, opened 2024-08-30
Allow bastion dnf package install to be turned off, on by default

tonykay:
##### SUMMARY

Allows bastion-base role to NOT install packages if set

- Packages will install by default, unless `bastion_install_basic_packages: true`

Motivated by immutable OS images like RHEL AI, however, a useful feature independent of motivation

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

roles/bastion-base

### DIFF
--- a/ansible/roles/bastion-base/tasks/main.yml
+++ b/ansible/roles/bastion-base/tasks/main.yml
@@ -1,6 +1,6 @@
 ---
 - name: Setup Red Hat packages for bastion
-  when: not bastion_install_basic_packages | default(false)
+  when: bastion_install_basic_packages | bool
   ansible.builtin.package:
     state: present
     name: "{{ agd_bastion_packages }}"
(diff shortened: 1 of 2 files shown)


rut31337 (approved):

ansible/roles/bastion-base/tasks/main.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: Setup Red Hat packages for bastion
+  when: not bastion_install_basic_packages | default(false)
```
jkupferer, line 3: iirc, should have a `bool` filter?
```
  when: not bastion_install_basic_packages | default(false) | bool
```

jkupferer (approved):

ansible/roles/bastion-base/tasks/main.yml
jkupferer: Actually... please set the default in the role defaults. Should not rely on an inline default and putting it in `defaults/main.yml` documents the variable.

jkupferer (changes requested):

ansible/roles/bastion-base/tasks/main.yml
tonykay: Good points both - I was surprised by the lack of a defaults, the good old days...

jkupferer (approved):

PR #8684, opened 2024-10-17
[ocp4_workload_virt_roadshow_vmware] Increase retries

agonzalezrh:
In case to dont be in the same datacenter, some calls can be slow and we need to retry more times

##### SUMMARY

In case to dont be in the same datacenter, some calls can be slow and we need to retry more times

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Role ocp4_workload_virt_roadshow_vmware


ashokjammula1 (approved):
lgtm

PR #8740, opened 2024-10-30
Adding NovaNoVNC config

ianwatsonrh:
Adds a httpd virtualhost enabling the Nova-NoVNC console to work from within the Openstack deployment. 

This requires a websocket upgrade that isn't in the existing wildcard virtual host.


klewis0928 (approved):
LGTM

PR #8752, opened 2024-11-04
Preserve user-data across run

fridim:
- Add action name prefix to `user-data.yaml`, `user-info.yaml`, `user-body.yaml`.
- Default `agnosticd_user_info` lookup to look in `provision-user-data.yaml`

That allows agnosticd to use user-data from provision, duing destroy or stop/start when:

- developping locally
- agnosticd_restore/save_output_dir  role is used in production
- [x] Provision using JBON and micro instance to validate multiregion. Use a small TTL to reduce costs.
- [X] Validate stop / start / destroy is finding the right region. without specifying the region.
- [X] Patch agnosticd base-infra config too
- [X] Patch cloudformation to not run if it was executed already (idempotency)
- [X] Provide a sample var for multiple region fallback

### DIFF
--- a/ansible/destroy.yml
+++ b/ansible/destroy.yml
@@ -4,6 +4,20 @@
 ############ Step 006 Destroy Workshop using workshop_prefix
 ################################################################################
 ################################################################################
+- name: Step 0000 Set Action
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  tags:
+    - must
+    - step0000
+    - setup_runtime
+  become: false
+  tasks:
+    - name: Set ACTION to provision
+      when: ACTION is undefined
+      set_fact:
+        ACTION: destroy
 
 - import_playbook: setup_runtime.yml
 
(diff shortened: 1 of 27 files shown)


jkupferer (changes requested):
discussion in slack...

bbethell-1 (approved):

ansible/update.yml
```diff
@@ -1,5 +1,19 @@
...
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  tags:
+    - must
+    - step0000
+    - setup_runtime
+  become: false
+  tasks:
+    - name: Set ACTION to provision
```
jkupferer, line 13: Task name...

ansible/destroy.yml
```diff
@@ -4,6 +4,20 @@
...
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  tags:
+    - must
+    - step0000
+    - setup_runtime
+  become: false
+  tasks:
+    - name: Set ACTION to provision
```
jkupferer, line 17: Task name here too.

tonykay (approved):

bbethell-1 (approved):

PR #8782, opened 2024-11-08
[infra-ec2-template-create] Set stack_deployed at the end of the block

agonzalezrh:
##### SUMMARY

Setting stack_deployed to true, it skips the tasks after the set_fact

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
infra-ec2-template-create role


newgoliath (approved):
I wonder how many other important variables we'll discover don't make the round trip

PR #8864, opened 2024-11-25
[install_operator] Add option install_operator_skip_operatorgroup to skip create operatorgroup

agonzalezrh:
##### SUMMARY

In some cases we want to still an operator inside a namespace where OperatorGroup exists, this PR is to avoid create the OperatorGroup as if it has different name causes issues.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
install_operator role


wkulhanek (approved):

ashokjammula1 (approved):
LGTM

PR #8893, opened 2024-11-28
Parasol: NFD workload fixes + Deployment end check

guimou:
##### SUMMARY

Two changes.
1) In the `ocp4_workload_nvidia_gpu_setup`, modify NFD deployment. The operand image version becomes a variable, with the default set at `registry.redhat.io/openshift4/ose-node-feature-discovery-rhel9:v4.17`
2) Parasol deployment workload modified to add an ArgoCD wait/check at the end. That way, the deployment is not finished until everything is properly deployed (no more sync wait time after you receive the go from the demo system).


ritzshah (approved):
Since Engineering added -rhel9 to the image and every OCP revision have a different value of image, this change is required for the NFD CR.

PR #8951, opened 2025-01-06
Add multi-platform support for execution environment

marcosmamorim:
##### SUMMARY

- Updated `install_binaries.sh` to dynamically detect system architecture (`ARCH`)
- Modified OpenShift CLI (`oc`) installation to use RHEL 8 version with architecture-specific tarballs.
- Adjusted AWS CLI download URL to include architecture detection.
- Added placeholder for Bitwarden CLI installation, as it does not support ARM64 yet.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ee-multicloud-public

### DIFF
--- a/.github/workflows/build-ee.yml
+++ b/.github/workflows/build-ee.yml
@@ -32,6 +32,16 @@ jobs:
 
       - uses: actions/setup-python@v4
 
+      - name: Setup QEMU for multi-platform builds
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu binfmt-support
+
       - name: Buildah Action
         uses: redhat-actions/buildah-build@v2
         id: build-image
--- a/.github/workflows/build-ee.yml
+++ b/.github/workflows/build-ee.yml
@@ -32,15 +32,15 @@ jobs:
 
       - uses: actions/setup-python@v4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ee.yml
```diff
@@ -32,16 +32,36 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu binfmt-support
+
+      - name: Setup QEMU for multi-platform builds
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
+      - name: Cache Buildah layers
```
marcosmamorim, line 45: @fridim I added cache layer here, it will improve the performance to rebuild the images

tools/execution_environments/ee-multicloud-public/install_binaries.sh
```diff
@@ -24,9 +37,8 @@ rm bw bw.zip
 
 
 # AWS CLI
-
 aws_version=2.4.23
-curl -s -L "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-${aws_version}.zip" \
+curl -s -L "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m)-${aws_version}.zip" \
```
fridim, line 41: shouldn't you be using $ARCH that you set earlier?

marcosmamorim: @fridim didn't work with arm, that's why I'm using uname there

PR #8960, opened 2025-01-11
workload_gitops_bootstrap - add health check and wait for all gitops apps

newgoliath:
- Feature Pull Request


ansible/roles_ocp_workloads/ocp4_workload_gitops_bootstrap/defaults/main.yml
```diff
@@ -13,3 +13,9 @@ ocp4_workload_gitops_bootstrap_helm_values: []
 
 ocp4_workload_gitops_bootstrap_health_retries: 60
 ocp4_workload_gitops_bootstrap_health_ignore: true
+
+# wait 15 mninutes for apps to roll out
+ocp4_workload_gitops_bootstrap_application_wait: true
+ocp4_workload_gitops_bootstrap_application_health_retries: 90
+ocp4_workload_gitops_bootstrap_application_health_ignore: true
+ocp4_workload_gitops_bootstrap_application_label: "demo.redhat.com/catalogItem = true"
```
juliaaano, line 21: This label will be our interface with the external world (GitOps repositories), so I want to be thoughtful about its name.

I believe naming it **demo.redhat.com/application** makes it less presumptive and more flexible.

juliaaano (changes requested):
I love this PR. Thank you so much for creating it.

ansible/roles_ocp_workloads/ocp4_workload_gitops_bootstrap/defaults/main.yml
juliaaano: I also don't think "= true" is necessary.

juliaaano: Finally, I see this is defined as a variable, but I wouldn’t encourage changing it in agV. The reason is the potential inconsistency across different CIs and GitOps repositories. This opens it up for changes without providing any clear benefit, as far as I know.

newgoliath: > I also don't think "= true" is necessary.

I like that it's specific.  This allows us to set it to "false" temporarily, if it's failing but we want to allow it.

newgoliath: > Finally, I see this is defined as a variable, but I wouldn’t encourage changing it in agV. The reason is the potential inconsistency across different CIs and GitOps repositories. This opens it up for changes without providing any clear benefit, as far as I know.

Removing the variable makes the contract stronger. 🦾

newgoliath: It's a binary label now.  "demo.redhat.com/application"
It's no longer a variable.

newgoliath:
@juliaaano I added code to copy data from labeled configmaps to agnosticd_user_info.

in a configmap:  `.data.users_json` is a json string of agnosticd_user_info formatted user data.
All other data in a configmap is expected to by yaml.

* first it copies all the non-`users_json` data into agnosticd_user_info.
* then it creates a variable `data_users_json` which is a recursive json merge of all configmaps' `data.user_json`
* finally it adds all this merged data into agnosticd_user_info

juliaaano:
Splitting this PR in two:

https://github.com/redhat-cop/agnosticd/pull/8962
https://github.com/redhat-cop/agnosticd/pull/8964

PR #8972, opened 2025-01-14
ocp4_workload_nginxplus - Update workload.yml

YoNoSoyVictor:
This item is failing as there is no pod defined as `my-nginx-ingress` [1]. The right reference is probably the NginxIngres here [2].

Changing and testing.

[1] https://github.com/redhat-cop/agnosticd/blob/2fd579f61a4adeeb3301f02ddfad0a93f4bb68c6/ansible/roles_ocp_workloads/ocp4_workload_nginxplus/defaults/main.yml#L22

[2] https://github.com/redhat-cop/agnosticd/blob/2fd579f61a4adeeb3301f02ddfad0a93f4bb68c6/ansible/roles_ocp_workloads/ocp4_workload_nginxplus/templates/nginx-ingress-controller.j2#L4


bbethell-1 (approved):
LGTM

PR #9025, opened 2025-02-03
ARO workshops: Fix dependency error for Rhel 9

YoNoSoyVictor:
This fix should fix the ongoing error with all ARO workshops.

For rhel 9 the correct repository is `https://packages.microsoft.com/config/rhel/9.0/packages-microsoft-prod.rpm` instead of the `https://packages.microsoft.com/config/rhel/9/packages-microsoft-prod.rpm` currently beig used which is causing a dependency error (Notice 9 vs 9.0)

This is explained in the Azure CLI installation documentation (Step 2 in [1])

[1] https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=dnf#install-azure-cli

### DIFF
--- a/ansible/configs/open-environment-azure/pre_software.yml
+++ b/ansible/configs/open-environment-azure/pre_software.yml
@@ -78,7 +78,7 @@
 
         - name: Add Azure CLI repository
           ansible.builtin.yum:
-            name: https://packages.microsoft.com/config/rhel/9/packages-microsoft-prod.rpm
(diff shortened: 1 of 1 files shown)


YoNoSoyVictor:
The sha256sum of both 9.0 and 9 packages is the same so this is unlikely to fix it. I opened an issue with the Azure CLI team: https://github.com/Azure/azure-cli/issues/30748

Will test anyways just in case and close the PR once it fails.

YoNoSoyVictor:
Test failed - Closing PR: 

image

PR #9028, opened 2025-02-04
config: sap-hana-rhel9  update download

rhmk:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
sap-hana-rhel9

##### ADDITIONAL INFORMATION

- Change the way the download was made to be more reliable, leading to less fixing in the future by using an official collection instead of the included role 
- Changed the way the cerbot certificate was created, because AAP installer requires RSA certificates now. Installation failed with the default ECDSA certificates
- fully tested in dev environment


klewis0928 (approved):

d-jana:
https://redhat-internal.slack.com/archives/C07V9CYU15G/p1738848808794259

PR #9031, opened 2025-02-06
enable multi-az in rosa classic sts

newgoliath:
- Bugfix Pull Request

I somehow missed getting this committed before RH1.

Might alleviate lack of baremetal machines


ansible/configs/rosa-consolidated/install_rosa_sts.yml
```diff
@@ -6,6 +6,12 @@
     --mode auto
     --yes
 
+- name: Create list of AZs
+  when: rosa_classic_multi_az | bool
+  ansible.builtin.set_fact:
+    _rosa_sts_azs: |
```
wkulhanek, line 12: What about us-east-1 which has a-f?

wkulhanek (approved):

ansible/configs/rosa-consolidated/install_rosa_sts.yml
newgoliath: Ends up I don't need any of them.  `--multi-az` should figure it out itself.

When creating machinepools in a multi-az rosa, machines should be automatically spread between AZ.

🤞

newgoliath:
Works OK as long as machinepool_replicas is a mulitple of 3

PR #9089, opened 2025-03-06
[infra-openshift-cnv-resources : Wait till VM is running -> Increase …

YoNoSoyVictor:
…retries

This task is failing sporadically in items such as `openshift-cnv.sap-e2e-demo-rhel9-cnv ` due to the fact that the VM does not have enough time to run.

Increasing retries to fix it.


bbethell-1 (approved):

agonzalezrh (changes requested):
Not needed

PR #9096, opened 2025-03-11
Fix role ocp4_workload_le_certificates

juremedvesek:
##### SUMMARY

Fixed:
 - vars were defined as a list of dicts with one item instead of dict with multiple keys
 - FQCN was missing on one instance of module ansible.builtin.setup. Made a change since in all other occurrences FQCN was used

New code should work on all Ansible versions that it worked before. But now also on Ansible 2.18.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
role  ocp4_workload_le_certificates

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_le_certificates/files/deploy_certs.yaml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_le_certificates/files/deploy_certs.yaml
@@ -8,9 +8,9 @@
   gather_facts: false
   become: false
   vars:
-  - _certbot_install_dir: "~/certificates"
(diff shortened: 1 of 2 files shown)


wkulhanek:
Do you need that workload? Could you switch to the ocp4_workload_cert_manager instead? We kind of switched over to that one in most catalog items.

wkulhanek (approved):

juremedvesek:
No I am not blocked by it by any kind. But was annoyed by deprecation warnings. Also it looked like nice entry point for a first contribution.

PR #9115, opened 2025-03-13
common: Add a conditional for basic packages

fridim:
##### SUMMARY
Add `common_update_packages`, `common_install_basic_packages` and `common_install_extra_packages`  booleans so the tasks can be easily skipped. By default make them true to keep current behavior.

##### ISSUE TYPE

- Feature Pull Request


agonzalezrh (approved):
lgtm and thanks! i really needed this and always forget to do it

PR #9154, opened 2025-03-23
ocp4_workload_odf: make label selector more robust

newgoliath:
If common.yaml sets this, but dev.yaml wants to override it, allow use of selector ="".

don't append the value to the list when the variable is ""

- Bugfix Pull Request


wkulhanek (approved):
How about
| default("") | length > 0 
For a one liner?

agonzalezrh (approved):

newgoliath:
> How about | default("") | length > 0 For a one liner?

Well, "" is a failure condition.  It must either be undef or "something"

PR #9159, opened 2025-03-24
Add role collect_openshift_must_gather

jkupferer:
##### SUMMARY

- Add role collect_openshift_must_gather to run `oc adm must-gather`.
- Add calling this role to rosa-consolidated config.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Role `collect_openshift_must_gather`
Config `rosa-consolidated`

### DIFF
--- a/ansible/roles/collect_openshift_must_gather/tasks/main.yml
+++ b/ansible/roles/collect_openshift_must_gather/tasks/main.yml
@@ -38,5 +38,4 @@
     --kubeconfig={{ collect_openshift_must_gather_kubeconfig_path }}
     {% endif %}
   ignore_errors: true
-
 ...
--- a/ansible/roles/collect_openshift_must_gather/tasks/main.yml
+++ b/ansible/roles/collect_openshift_must_gather/tasks/main.yml
@@ -1,0 +1,42 @@
+---
+- name: Write kubeconfig if not exists and have api token and url
+  when:
+  - collect_openshift_must_gather_kubeconfig_path is not exists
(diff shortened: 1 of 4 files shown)


ansible/roles/collect_openshift_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+
+- name: Run oc adm must-gather
+  ansible.builtin.command: >-
+    oc adm must-gather
+    --dest-dir="{{ output_dir }}/oc-adm-must-gather"
+    {% if collect_openshift_must_gather_kubeconfig_path is exists %}
+    --kubeconfig={{ collect_openshift_must_gather_kubeconfig_path }}
+    {% endif %}
+  ignore_errors: true
+...
```
newgoliath, line 41: Accessing the must-gather post-destroy via s3 bucket might be challenging.  Just speculating: Is there any automation for investigators to grab it?  Maybe there's an even better place for them.

newgoliath: Perhaps put a note here about how it fits into the larger solution.

newgoliath (approved):

ansible/roles/collect_openshift_must_gather/tasks/main.yml
jkupferer: It would be good in the future to have a way to download the output dir from the RHDP ui. We don't have it on the roadmap yet, but we should add it.

PR #9204, opened 2025-04-07
replaced instance filter

miteshget:
##### SUMMARY

Replace the instance **name based** filter with a **label based filter,** consistent with the filtering method used for other cloud providers in infra related roles.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Infra-Role: infra-gcp-create-inventory

### DIFF
--- a/ansible/roles-infra/infra-gcp-create-inventory/tasks/main.yml
+++ b/ansible/roles-infra/infra-gcp-create-inventory/tasks/main.yml
@@ -16,7 +16,7 @@
   # command: "gcloud compute instances list --filter=\"name~'^{{ project_tag }}-vm$'\" --format json"
   ansible.builtin.command: >-
     gcloud compute instances list
-    --filter="labels.ansiblegroup=bastions"
+    --filter="labels.ansiblegroup:*"
     --format json
   register: r_vm_info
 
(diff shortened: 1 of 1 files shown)


bbethell-1 (approved):

bbethell-1 (approved):

agonzalezrh (changes requested):
doesn't look a good idea to filter only for bastions, what about other VMs?

rut31337 (approved):

miteshget:
> doesn't look a good idea to filter only for bastions, what about other VMs?

Could please check entire role not just this task?

agonzalezrh:
> > doesn't look a good idea to filter only for bastions, what about other VMs?
> 
> Could please check entire role not just this task?

it has only one task, or im missing something?

agonzalezrh:
> > doesn't look a good idea to filter only for bastions, what about other VMs?
> 
> Could please check entire role not just this task?

so instance_list is to use to create the inventory and r_vm_info is used for what? just to get the gcp_zone?

bbethell-1 (approved):

miteshget:
@agonzalezrh I have made the changes and tested. Everything is working as expected.

can you please re review ?

agonzalezrh (approved):

agonzalezrh:
> @agonzalezrh I have made the changes and tested. Everything is working as expected.
> 
> can you please re review ?

thanks! probably we need to update the whole logic in the future @rut31337 

approved

PR #9284, opened 2025-04-29
install_operator role - Add retry logic to kubernetes.core.k8s calls

YoNoSoyVictor:


agonzalezrh (approved):
lgtm to me, adjusted to 6/5 instead 30/10

PR #9363, opened 2025-06-16
[ADVANCED_GITOPS] Label namespaces for Argo CD

gnunn1:
##### SUMMARY

This PR labels user workspaces to be managed by their corresponding Argo CD instance using the managed-by label. Not sure if there is a better way to do this in AgnosticV where the namespaces are originally created, it's complicated given the label changes based on the user.  

It also tweaks Argo CD to add support OOTB for cluster-admins group which will be added later.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
Advanced GitOps Workshop (In Development only)


stencell (approved):
There may be slightly more efficient ways to do this, but this is perfectly fine and how it is done across a number of other roles.

PR #9347, opened 2025-06-10
OCP4_WORKLOAD_QUARKUS_SUPER_HEROES_DEMO  - Update to RHBQ 3.20

edeandrea:
##### SUMMARY

This is to move the `OCP4_WORKLOAD_QUARKUS_SUPER_HEROES_DEMO` to the red hat build of Quarkus 3.20 along with moving to OpenShift 4.18.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

OCP4_WORKLOAD_QUARKUS_SUPER_HEROES_DEMO

PR #9365, opened 2025-06-17
OCP4_WORKLOAD_QUARKUS_SUPER_HEROES_DEMO - Update to RHBQ 3.20

edeandrea:
This is in addition to #9347

This makes the Kafka instance ephemeral. It was working in dev but not test. The brokers being ephemeral is fine.


d-jana (approved):

edeandrea:
@d-jana did test get re-tagged in agV with these changes?

PR #9372, opened 2025-06-23
IBM Cloud: do not add guid to instance name and fix instance count

fridim:
##### SUMMARY
adding GUID to instance name is an implementation detail, underlying
logic to work at the cloud provider level.
At the agnosticV and agnosticD level, allow names to be simple.

- Allow simply `bastion.GUID.DOMAIN.TLD`
  instead of  `bastion-GUID.GUID.DOMAIN.TLD`

- Add an helper 'instance_name_with_guid' to add the guid for names of
instance resources (network card, disk, ...) which should be unique
across the account.

- Add mandatory tags: guid, service_uuid to instance - Cost tracking etc.

- Fix find_ip_query when building the inventory

- Fix broken logic in the loops of inventory (double looping)

- reuse the equinix_metal filter to translate tags from list to dict

- Implement logic for `instances[].count`: introduce a new filter `agnosticd_expand_instances` to facilitate and simplify the code.
- delete all floating IPs in the RG (using ibmcloud CLI) instead of deleting only {instance-name}-fip
- Fix volumes key as it was using the ibmcloud resource name

- Use ibmcloud CLI for SSH keys (faster, **save 20-40s** compared to ansible module)
- Use ssh_provision_key_type: ed25519  for IBM Cloud  env key
- fix start/stop
- fix lifecycle status

```
Instance             State      Health
----------------------------------------------------------------
bastion-6s6lx        running    ok
node1-6s6lx          running    ok
node2-6s6lx          running    ok
```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

IBM Cloud provider - core

### DIFF
--- a/ansible/roles-infra/infra-ibm-resource-group-resources/tasks/delete_instance.yaml
+++ b/ansible/roles-infra/infra-ibm-resource-group-resources/tasks/delete_instance.yaml
@@ -4,7 +4,6 @@
   ibm.cloudcollection.ibm_is_instance:
     region: "{{ ibmcloud_region }}"
     state: absent
-    #name: "{{ instance_name_with_guid }}"
     id: "{{ instance.id }}"
     image: "{{ instance.image }}"
     vpc: "{{ instance.vpc }}"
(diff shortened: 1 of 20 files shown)


agonzalezrh (approved):
thanks!

ansible/roles-infra/infra-ibm-resource-group-resources/tasks/delete_instance.yaml
```diff
@@ -4,7 +4,7 @@
   ibm.cloudcollection.ibm_is_instance:
     region: "{{ ibmcloud_region }}"
     state: absent
-    name: "{{ instance.name }}"
+    #name: "{{ instance_name_with_guid }}"
```
agonzalezrh: commented?

fridim: yes, not needed as this is a deletion, the instance.id is enough.

agonzalezrh (approved):

agonzalezrh (approved):

PR #9407, opened 2025-07-22
Add option to leave nfd_operator_operand_image empty

rbo:
##### SUMMARY

With newer version, you don't need and want to set `nfd_operator_operand_image`

Now you can simply add 
```
nfd_operator_operand_image: ""
```
and image is not set.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

 - gpu operator & nfd operator

##### ADDITIONAL INFORMATION

n/a


wkulhanek:
@rbo Thank you. I'm looking at this as well and frankly nfd operator needs to be its own workload.

wkulhanek (approved):

wkulhanek:
Do you need the other settings (under image) if you don't set an operand image?

rbo:
> Do you need the other settings (under image) if you don't set an operand image?

I don't know, Looks like that's the defaults from the operator.

wkulhanek:
Ok. Probably won't hurt.

PR #9413, opened 2025-07-24
Write out external IP to a file when running on CNV

stencell:
##### SUMMARY

The external IP is available as an agnosticd variable, but it is not available to someone inside the environment. When someone uses one of these environments to run with their own automation or lab guides, they potentially need access to this IP without having to go to the RHDP UI to find it.

This will write out the IP to a file on the bastion. It will also display the external IP as agnosticd_user_info data, not just in the message.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles/bastion-base
configs/base-infra

### DIFF
--- a/ansible/configs/base-infra/post_software.yml
+++ b/ansible/configs/base-infra/post_software.yml
@@ -64,8 +64,8 @@
               targetport: "{{ hostvars[groups['bastions'][0]].bastion_ssh_port }}"
               subdomain: "{{ guid }}{{ subdomain_base_suffix }}"
               subdomain_internal: "{{ chomped_zone_internal_dns | default('') }}"
-              external_ip: "{{ hostvars[groups['bastions'][0]]['external_ip'] }}"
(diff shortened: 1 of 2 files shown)


ansible/configs/base-infra/post_software.yml
```diff
@@ -64,6 +64,8 @@
               targetport: "{{ hostvars[groups['bastions'][0]].bastion_ssh_port }}"
               subdomain: "{{ guid }}{{ subdomain_base_suffix }}"
               subdomain_internal: "{{ chomped_zone_internal_dns | default('') }}"
+              external_ip: "{{ hostvars[groups['bastions'][0]]['external_ip'] }}"
```
agonzalezrh: Add | default('') so wont fail on other CIs

and we can add a when to the task to write the IP

agonzalezrh (changes requested):

bbethell-1 (approved):

agonzalezrh (approved):

PR #9449, opened 2025-08-06
assisted installer - ad hoc ignition manifests - ai_machineconfig

newgoliath:
- Feature Pull Request

One lab I'm working on needs kernel parameters applied at install time.  It doesn't work post-install, because node reboots locks up a busy cluster that's resource constrained.

agV example

```
ai_machineconfigs:
- apiVersion: machineconfiguration.openshift.io/v1
  kind: MachineConfig
  metadata:
    labels:
      machineconfiguration.openshift.io/role: worker
    name: 99-openshift-machineconfig-worker-psi-karg
  spec:
    kernelArguments:
    - psi=1
```


agonzalezrh (approved):
lgtm

wkulhanek:
@newgoliath As I'm moving this to V2 two things:
1. the when check should check for an empty array
2. Your example from the PR should be in defaults/main.yml

PR #9490, opened 2025-08-20
ocp4_workload_advanced_gitops_workshop | Fix retry logic for "Wait fo…

YoNoSoyVictor:
…r ArgoCD instances to successfully deploy""

---

This is causing some items to fail. When checking AAP logs I can see that no retries are performed despite the task having retry logic. Example [1].

The until condition is evaluated on the very first run of the task. If the status key is missing, the condition fails immediately and the task fails without retrying.

the `| default({})` filter added provides a default empty dictionary for the status key. This ensures the expression won't fail even if the `status` key is not present.

[1] https://aap2-prod-us-east-2.aap.infra.demo.redhat.com/#/jobs/playbook/1496274/output


bbethell-1 (approved):

bbethell-1:
Cheers

PR #9493, opened 2025-08-21
[ADVANCED_GITOPS] Checks that status and phase are defined before checking Available

gnunn1:
##### SUMMARY

This checks if the status and phase are defined in the ArgoCD CR before checking for the Available variable in case this is checked too quickly before the operator starts provisioning and setting these variables.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ADVANCED_GITOPS


perryrivera1 (approved):
LGTM

PR #9498, opened 2025-08-25
Move ODCR role destroy to top-level destroy

fridim:
In case a config has its own destroy, the infra-aws-capacity-reservation role doesn't run at destroy. Thus, aws_region var is undefined. That can cause issues in some config destroy if they don't explicitely call the role or set the variables back from agnosticd_user_data.


bbethell-1 (approved):

agonzalezrh (approved):
**lgtm**

jkupferer (approved):

PR #9511, opened 2025-08-29
Update Microshift workload to use s3 with credential

stencell:
##### SUMMARY

The current workload downloads the microshift image from a public URL, which shouldn't be allowed. This will gives the option of providing s3 credentials to download from an S3 bucket.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles_ocp_workloads/ocp4_workload_microshift


agonzalezrh (approved):
lgtm

PR #9519, opened 2025-09-04
Enable defining number of OpenShift AI Dashboard replicas

stencell:
##### SUMMARY

Adds functionality to define the number of replicas required by the rhods-dashboard deployment.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
roles_ocp_workloads/ocp4_workload_openshift_ai

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_openshift_ai/tasks/workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_openshift_ai/tasks/workload.yml
@@ -62,7 +62,7 @@
             name: rhods-dashboard
             namespace: redhat-ods-applications
           spec:
-            replicas: "{{ ocp4_workload_openshift_ai_dashboard_replicas | int }}"
+            replicas: {{ ocp4_workload_openshift_ai_dashboard_replicas | int }}
 
 - name: Get the OpenShift AI dashboard route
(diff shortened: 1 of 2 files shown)


ansible/roles_ocp_workloads/ocp4_workload_openshift_ai/tasks/workload.yml
```diff
@@ -33,6 +33,37 @@
     msg: r_ds_cluster
     verbosity: 2
 
+- name: Configure rhods-dashboard deployment replicas
+  when:
+    - ocp4_workload_openshift_ai_deploy_dsc | bool
+    - ocp4_workload_openshift_ai_dashboard == "Managed"
+    - ocp4_workload_openshift_ai_dashboard_replicas is defined
```
juliaaano, line 38: I believe only this condition would be necessary:

ocp4_workload_openshift_ai_dashboard_replicas is defined

Adding the other ones creates a unnecessary coupling. I think if this last condition is true, the other first two condition will always be true.

Not a big deal, fell free to adjust if you wish.

stencell: thanks for the feedback, i have removed them as you suggested.

PR #9590, opened 2025-10-08
[ocp4_workload_cert_manager] Add variable to define the timeout

agonzalezrh:
##### SUMMARY

Currently it is set to a high fix value, 1200. This PR sets to a more reasonable value, 600.

As we have implemented retries (long time ago) and fallback (recently)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ocp4_workload_cert_manager role


stencell (commented):
@wkulhanek should review this.

If you want to set this as a var, fine. You should take the current value as the default, though. Don't cut it in half by default, let people override it if they want to.

wkulhanek (approved):

PR #9613, opened 2025-10-24
Load provision-user-data.yaml in other actions

fridim:
- destroy, start, stop, lifecycle

Use a variable `agnosticd_load_provision_user_data` to control this behavior.
Set it to false for now as the default.


jkupferer (changes requested):
User data items are not variables for AgD, they are an interface for upstream things like babylon and showroom. We need to discuss further.

PR #9709, opened 2026-01-23
Fix workload ocp4_workload_openshift_data_foundation - add support for 4.19, 4.20

ryanj:
Add support for ODF channels: `stable-4.19`, `stable-4.20`

##### SUMMARY
⚠️ ODF is currently unavailable on ocp-4.20

This change adds support for two new ODF channels while maintaining backward compatibility with previous releases

I considered these follow-up tasks (instead of configuring based on the odf channel info), but they were not easy to solve:
* Check the subscription CSV to determine support for StorageSystems CRD (deprecated in stable-4.19 and later releases)
* Find the ConsolePlugin api_version by fetching the related CRD

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`ansible/roles_ocp_workloads/ocp4_workload_openshift_data_foundation`

##### ADDITIONAL INFORMATION
This change has been tested with the following ODF channels: `stable-4.18`, `stable-4.19`, `stable-4.20`
via: https://catalog.demo.redhat.com/catalog?item=babylon-catalog-dev/rhdp.ocp4-rad-ea-rs-wksp.dev

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_openshift_data_foundation/tasks/workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_openshift_data_foundation/tasks/workload.yml
@@ -75,8 +75,7 @@
     definition: "{{ lookup('template', 'subscription.yaml.j2') | from_yaml }}"
 
 - name: "Config for channel {{ ocp4_workload_openshift_data_foundation_channel }}"
-  when:
-    - ocp4_workload_openshift_data_foundation_channel not in ["stable-4.19", "stable-4.20"]
+  when: ocp4_workload_openshift_data_foundation_channel not in ["stable-4.19", "stable-4.20"]
   ansible.builtin.set_fact:
     _console_plugin_api_version: "console.openshift.io/v1alpha1"
     _storage_systems_enabled: true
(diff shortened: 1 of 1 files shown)


ansible/roles_ocp_workloads/ocp4_workload_openshift_data_foundation/tasks/workload.yml
```diff
@@ -85,6 +92,9 @@
   until:
   - r_crd_storagesystems.resources is defined
   - r_crd_storagesystems.resources | list | length == 1
+  when:
```
wkulhanek: How about
when: _storage_systems_enabled | default(false) | bool

wkulhanek (approved):

ansible/roles_ocp_workloads/ocp4_workload_openshift_data_foundation/tasks/workload.yml
ryanj: that looks better - thanks for the tip

wkulhanek (approved):

wkulhanek (approved):

PR #9718, opened 2026-02-10
Fix infra-ec2-create-inventory failure when no instances exist

rut31337:
##### SUMMARY

Add guard condition to the bastion lookup task to handle empty instance lists. Commit 19f9d023 (#9354) refactored the bastion lookup but removed the when condition that previously skipped the task when no instances matched, causing '| first' to fail with 'No first item, sequence was empty' on environments without EC2 instances (e.g., sandbox open-environment configs).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

roles-infra/infra-ec2-create-inventory

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/ansible/roles-infra/infra-ec2-create-inventory/tasks/main.yml
+++ b/ansible/roles-infra/infra-ec2-create-inventory/tasks/main.yml
@@ -44,6 +44,7 @@
        | map(attribute='tags.internaldns')
        | sort | first
       }}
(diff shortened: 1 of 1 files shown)


perryrivera1 (approved):
lgtm

bbethell-1 (approved):

agonzalezrh (approved):

PR #9734, opened 2026-03-02
ocp4_workload_external_odf: scale NooBaa pvPool to match worker node count

prakhar1985:
## Problem

When clusters are provisioned with more than 3 worker nodes, the NooBaa default backing store \`pvPool.numVolumes\` stays at the operator default and does not scale with the cluster. This causes Quay and other object storage consumers to fail when they cannot provision enough buckets.

## What this does

Exposes a single variable \`ocp4_workload_external_odf_noobaa_pv_pool_num_volumes\` that catalog items can set to control the NooBaa pvPool size. When not set (default \`0\`), the role leaves the operator default untouched — no behaviour change for existing users.

When set, the role waits for the BackingStore to be ready and patches \`pvPool.numVolumes\` accordingly.

## Variable

| Variable | Default | Description |
|---|---|---|
| \`ocp4_workload_external_odf_noobaa_pv_pool_num_volumes\` | \`0\` (no-op) | Number of pvPool volumes. Set to match worker node count. |
| \`ocp4_workload_external_odf_noobaa_deploy_timeout\` | \`600\` | Seconds to wait for BackingStore to be ready. |

## Example usage in an AgnosticV catalog item

\`\`\`yaml
ocp4_workload_external_odf_noobaa_pv_pool_num_volumes: "{{ worker_instance_count | int }}"
\`\`\`

## Test plan

- [ ] Deploy without setting the variable — verify no behaviour change (operator default used)
- [ ] Deploy with \`ocp4_workload_external_odf_noobaa_pv_pool_num_volumes: 6\` — verify BackingStore pvPool has 6 volumes
- [ ] Verify Quay provisions successfully when pvPool matches worker count

> This is a possible solution. @alberto-gonzalez — can you validate in integration before we merge?


prakhar1985:
Closing — the actual role lives in agnosticd/core_workloads. Opening PR there instead.

PR #9735, opened 2026-03-06
ec2_infrastructure_deployment.yml - Fix new ansible bool logic for "C…

YoNoSoyVictor:
Task: `create_ssh_provision_key`

Ansible changed how the condition is evaluated and now it needs an actual bool. Other tasks were changed already e.g.

```
- name: Locate environment SSH key
      include_role:
        name: locate_env_authorized_key
      when: instances | default([]) | length > 0

    - name: Create keypair in ec2
      include_role:
        name: infra-ec2-ssh-key
      when: instances | default([]) | length > 0
```

but this one was left behind

### DIFF
--- a/ansible/cloud_providers/ec2_infrastructure_deployment.yml
+++ b/ansible/cloud_providers/ec2_infrastructure_deployment.yml
@@ -43,7 +43,7 @@
       include_role:
         name: create_ssh_provision_key
       when:
-        - instances | default([]) | length
+        - instances | default([]) | length > 0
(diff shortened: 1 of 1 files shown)


bbethell-1 (approved):

yordangit (approved):

stencell:
Why is this being changed? This has been fine for years, it seems. Can you provide an example of something that is broken that requires this change?

fridim (approved):

PR #9776, opened 2026-04-16
fix: infra-openshift-cnv-resources network config not valid JSON

prakhar1985:
## Problem

`infra-openshift-cnv-resources/tasks/create_network.yaml` defines `config`
as a Python-style dict string (single quotes):

\`\`\`yaml
config: "{'cniVersion':'0.3.1','type':'ovn-k8s-cni-overlay',...}"
\`\`\`

Applying `| to_json` to a **string** just wraps it in JSON string quotes
rather than converting the content to JSON. Multus webhook rejects the
resulting NetworkAttachmentDefinition:

\`\`\`
admission webhook "multus-validating-config.k8s.io" denied the request:
configuration string is not in JSON format
\`\`\`

## Fix

Define `config` as a proper YAML dict. `| to_json` on a dict produces
valid JSON that Multus accepts.

## Affects

lb1404 and any other catalog item using `infra-openshift-cnv-resources`
with CNV network attachment definitions.

### DIFF
--- a/ansible/roles-infra/infra-openshift-cnv-create-inventory/tasks/create_inventory.yaml
+++ b/ansible/roles-infra/infra-openshift-cnv-create-inventory/tasks/create_inventory.yaml
@@ -42,7 +42,7 @@
         sessionAffinity: None
         type: NodePort
   register: r_expose_bastion
-  when: local_bastion|default(False)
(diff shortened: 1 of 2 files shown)


agonzalezrh:
Use EE chain 4-10

PR #9780, opened 2026-04-22
fix(showroom): remap .ssh ownership for ansible-runner container (rootless podman uid mapping)

prakhar1985:
## What breaks

When \`showroom_ansible_runner_api: true\` is set, every lab that uses the ZT runner to execute solve/validate playbooks **fails on every module** with:

\`\`\`
fatal: [node]: UNREACHABLE! => {
  "msg": "Failed to connect to the host via ssh: Can't open user config file /app/.ssh/config: Permission denied"
}
\`\`\`

Solver and validator cannot connect to any lab node. Every click of **Check** or **Solve** in the showroom UI fails immediately.

## Root cause

The \`.ssh\` directory and config file are created owned by \`showroom:showroom\` (uid=1888, mode \`700\`/\`600\`) and volume-mounted into the \`ansible-runner-api\` container at \`/app/.ssh\`.

The container runs as **uid=1001** (\`default\` user in the image). In rootless podman, the showroom user's UID namespace maps:

| container uid | host uid |
|---|---|
| 0 | 1888 (showroom) |
| 1 | 231072 (first subuid) |
| 1001 | 232072 (first subuid + 1000) |

The container process (uid=232072 on host) cannot read files owned by uid=1888 with \`700\`/\`600\` permissions. Every SSH attempt by Ansible fails before a connection is even opened.

## Fix

After writing the SSH config, run \`podman unshare chown -R 1001:1001\` on the \`.ssh\` directory. \`podman unshare\` executes inside the showroom user's user namespace, so uid=1001 inside translates to the correct host uid (232072) that the container process actually runs as.

Gated on \`showroom_ansible_runner_api\` — no-op for deployments not using the runner.

## Tested on

- \`ocpvdev01.rhdp.net\` ZeroTouch single-pod deployment (GUID \`fg8gg\`)
- After fix: all modules connect to lab nodes and run validation checks correctly


prakhar1985:
Closing — the agnosticd showroom role is no longer actively used for new deployments. The fix has been applied to agnosticd/showroom#65 (vm_workload_showroom) which is the active path.

PR #9802, opened 2026-06-02
fix(infra_vmware_ibm): detach FCD disks before VM deletion

rut31337:
## Summary
- Detaches and deletes First Class Disk (FCD) attachments from VMs before attempting VM deletion
- Fixes "Invalid virtual machine state - incomplete disk information" errors during destroy

## Problem
OCP clusters deployed on VMware IBM create persistent volumes as FCDs via the vSphere CSI driver. During destroy:
1. The OCP cluster is torn down, which may delete the FCD objects but leave the backing VMDKs
2. The VM config still references the `fcd/` backed disks
3. `community.vmware.vmware_guest` with `state: absent` fails with "Invalid virtual machine state" because it can't resolve the FCD metadata for these disks

## Fix
Before the "Remove the VMs" step, the role now:
1. Queries disk info for each VM (`vmware_guest_disk_info`)
2. Identifies disks with `fcd/` in the backing filename
3. Detaches and deletes those disks (`vmware_guest_disk` with `state: absent, destroy: true`)

Handles both cases:
- **FCD still registered + attached**: detach removes attachment, destroy deletes the VMDK
- **FCD deleted, orphaned VM reference**: detach cleans up stale config, destroy removes leftover VMDK

No-op for VMs without FCD disks.

## Test plan
- [x] Tested locally against vcsnsx-vc with sandbox-48v2l (2 VMs, 4 FCD disks)
- [x] Detach succeeded on all 4 disks, VM deletion succeeded on both VMs
- [ ] Verify no regression on VMs without FCD disks (standard sandbox deployments)

### DIFF
--- a/ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_instances.yaml
+++ b/ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_instances.yaml
@@ -6,6 +6,47 @@
 - name: Stop instances
   include_tasks: stop_instances.yaml
 
+- name: Get disk info for each VM
+  community.vmware.vmware_guest_disk_info:
+    datacenter: "{{ vcenter_datacenter }}"
+    name: "{{ item.guest_name }}"
+    validate_certs: "{{ vmw_validate_certs | default(false) | bool }}"
+  loop: "{{ r_vmc_vms.virtual_machines | default([]) }}"
+  loop_control:
+    label: "{{ item.guest_name }}"
+  register: r_disk_info
+
+- name: Build list of FCD disks to detach
+  ansible.builtin.set_fact:
+    _fcd_disks_to_remove: >-
+      {{
+        _fcd_disks_to_remove | default([]) + (
+          item.guest_disk_info | dict2items
+          | selectattr('value.backing_filename', 'search', 'fcd/')
+          | map('combine', {'_vm_name': item.item.guest_name})
+          | list
+        )
+      }}
+  loop: "{{ r_disk_info.results }}"
+  loop_control:
+    label: "{{ item.item.guest_name }}"
+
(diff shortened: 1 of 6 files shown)


ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_instances.yaml
```diff
@@ -6,6 +6,47 @@
 - name: Stop instances
   include_tasks: stop_instances.yaml
 
+- name: Get disk info for each VM
+  community.vmware.vmware_guest_disk_info:
+    datacenter: "{{ vcenter_datacenter }}"
+    name: "{{ item.guest_name }}"
+    validate_certs: "{{ vmw_validate_certs | default(false) | bool }}"
```
jkupferer: `vmw_validate_certs` should have default in `defaults/main.yml` with documentation rather than inline filter.

ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_instances.yaml
```diff
@@ -6,6 +6,47 @@
...
+  loop: "{{ r_disk_info.results }}"
+  loop_control:
+    label: "{{ item.item.guest_name }}"
+
+- name: Detach FCD disks before VM deletion
+  when: _fcd_disks_to_remove | default([]) | length > 0
+  community.vmware.vmware_guest_disk:
+    datacenter: "{{ vcenter_datacenter }}"
+    name: "{{ item._vm_name }}"
+    validate_certs: "{{ vmw_validate_certs | default(false) | bool }}"
```
jkupferer: Same comment, inline default here indicates missing documentation of variable in `defaults/main.yml`

jkupferer (approved):

PR #9805, opened 2026-06-05
Add cleanup logic to ocp4_workload_open_ssh_from_bastion

bbethell-1:
## Summary
Add proper cleanup logic to the `ocp4_workload_open_ssh_from_bastion` workload to prevent destroy failures.

## Problem
The workload creates VPC infrastructure to enable SSH from bastion to cluster nodes:
- VPC Peering Connections
- Custom routes in route tables
- Security group rules allowing SSH

However, `remove_workload.yml` was empty - no cleanup logic existed. This can cause cluster destroy failures because:
1. VPC Peering Connections can block VPC deletion
2. Custom routes in route tables can prevent route table cleanup
3. Modified security groups can cause dependency violations

## Changes
Implemented `remove_workload.yml` with proper cleanup:
1. ✅ Remove SSH security group rules from cluster controlplane/node SGs
2. ✅ Remove routes from both bastion and cluster route tables  
3. ✅ Delete VPC peering connection
4. ✅ Add error handling (`failed_when: false`) to handle already-deleted resources
5. ✅ Follow same VPC/SG discovery logic as `workload.yml` for consistency

## Testing
- Code follows same pattern as workload.yml
- Cleanup is idempotent (safe to run multiple times)
- Error handling prevents failures if resources already deleted

## Related
- Related to rhpds/clusterplatform-agnosticv#58 which added this workload to the catalog
- Addresses potential destroy failures similar to those seen with other SSH access workloads

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9806, opened 2026-06-05
fix: add explicit route cleanup before deleting VPC peering

bbethell-1:
## Problem

Deployments using `sandbox-ocp4-0.3.3` are still experiencing VPC `DependencyViolation` errors during destroy despite the cleanup logic from #9805.

**Root Cause:** The cleanup deletes the VPC Peering Connection but does NOT explicitly remove the routes pointing to it from route tables. These routes can still block VPC deletion.

## Solution

Add explicit route cleanup BEFORE deleting the VPC peering connection:

1. Get bastion and cluster route tables
2. Remove routes from bastion route tables pointing to cluster CIDR
3. Remove routes from cluster route tables pointing to bastion CIDR
4. THEN delete VPC peering connection
5. THEN VPC can be deleted cleanly

## Testing

- Observed failure on sandbox2238 (jgkgf) using 0.3.3
- VPC deletion stuck with DependencyViolation
- This fix ensures routes are removed before peering deletion

## Related

- Follow-up to #9805
- Addresses production destroy failures
- Will be tagged as `sandbox-ocp4-0.3.4`

### DIFF
--- a/ansible/roles_ocp_workloads/ocp4_workload_open_ssh_from_bastion/tasks/remove_workload.yml
+++ b/ansible/roles_ocp_workloads/ocp4_workload_open_ssh_from_bastion/tasks/remove_workload.yml
@@ -71,7 +71,7 @@
 
         ###########################################################################
         #
-        # Step 2: Remove routes from route tables
+        # Step 2: Get VPC Peering Connection info for route cleanup
         #
         ###########################################################################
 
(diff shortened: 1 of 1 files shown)


rhjcd (approved):

perryrivera1 (approved):
lgtm

PR #9845, opened 2026-07-08
Fix jinja2 lstrip_blocks syntax in cloud_template.j2

YoNoSoyVictor:
Newer Jinja2 (with updated execution environments) requires a boolean, not the string "True".

Example error: https://aap2-prod-us-east-2.aap.infra.demo.redhat.com/#/jobs/playbook/2625120/output


bbethell-1 (approved):

bbethell-1:
@fridim  Bonjour - please as we need Core!!!

fridim (approved):

PR #9827, opened 2026-06-29
infra-openshift-cnv-resources / Create network -> Use valid json

YoNoSoyVictor:
The current configuration causes errors. See [1]:

```
TASK [infra-openshift-cnv-resources : Create network default] ******************
Monday 29 June 2026  11:26:18 +0000 (0:00:00.264)       0:00:12.447 *********** 
FAILED - RETRYING: [localhost]: Create network default (6 retries left).
FAILED - RETRYING: [localhost]: Create network default (5 retries left).
FAILED - RETRYING: [localhost]: Create network default (4 retries left).
FAILED - RETRYING: [localhost]: Create network default (3 retries left).
FAILED - RETRYING: [localhost]: Create network default (2 retries left).
FAILED - RETRYING: [localhost]: Create network default (1 retries left).
fatal: [localhost]: FAILED! => {"attempts": 6, "changed": false, "msg": "Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"admission webhook \\\\\"multus-validating-config.k8s.io\\\\\" denied the request: configuration string is not in JSON format\",\"code\":400}\\n'", "reason": "Bad Request"}
```

Relevant snippet being `configuration string is not in JSON format`.

In order to fix it, the config variable should be defined as a proper YAML dictionary, not a string.

https://aap2-prod-us-east-2.aap.infra.demo.redhat.com/#/jobs/playbook/2605006/output

Affected item: [Red Hat Openshift Single Node](https://catalog.demo.redhat.com/catalog/babylon-catalog-prod?item=babylon-catalog-prod/zt-rhelbu.zt-openshift-sno.prod&utm_source=webapp&utm_medium=share-link)

PR #9855, opened 2026-07-09
Fix Multus network config JSON serialization in infra-openshift-cnv-resources

rhjcd:
## Summary

- Converts the `config` variable in `create_network.yaml` from a Python dict string literal to a proper YAML dictionary
- The Python dict syntax with single quotes was not producing valid JSON when passed through `to_json`, causing Multus validating webhook to reject NetworkAttachmentDefinition creation
- Error: `admission webhook "multus-validating-config.k8s.io" denied the request: configuration string is not in JSON format`
- This is the same fix as PRs #9776 and #9827 which have been open since April and June respectively
- Tagged as `zero-touch-prod-1.1.1` for immediate use by zerotouch items that depend on CNV network creation

## Test plan

- [ ] Verify NetworkAttachmentDefinition creation succeeds on CNV clusters with Multus validating webhook enabled
- [ ] Test with zt-ocp-lab-developer-cnv which provisions VMs with network attachment

### DIFF
--- a/ansible/roles-infra/infra-openshift-cnv-resources/tasks/create_network.yaml
+++ b/ansible/roles-infra/infra-openshift-cnv-resources/tasks/create_network.yaml
@@ -1,16 +1,7 @@
 ---
 - name: Create network {{ _network.name }}
   kubernetes.core.k8s:
-    definition:
-      apiVersion: k8s.cni.cncf.io/v1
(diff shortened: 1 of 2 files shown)


agonzalezrh:
@rhjcd have you tested with a old EE and new EE to confirm it works with both ansible versions?

PR #9860, opened 2026-07-09
Added become: false block to create_ssh_provision_key

YoNoSoyVictor:
(Do Not merge yet) Added a `become: false` block to `create_ssh_provision_key` so it works when run on AAP EE (New EE's do not have sudo anymore).

See https://prod0.apps.ocpv-infra01.dal12.infra.demo.redhat.com/#/jobs/playbook/43044/output 

Looks like a big change because I indented all lines to include them in the block, but all tasks remain unchanged. If any other items start failing we can revert.

### DIFF
--- a/ansible/roles-infra/create_ssh_provision_key/tasks/main.yaml
+++ b/ansible/roles-infra/create_ssh_provision_key/tasks/main.yaml
@@ -1,46 +1,50 @@
 ---
-- include_tasks: checks.yaml
-- name: Generate SSH keys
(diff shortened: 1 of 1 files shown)


bbethell-1 (approved):

YoNoSoyVictor:
This now passes the previous failing tasks: https://prod1.apps.ocpv-infra02.wdc07.infra.demo.redhat.com/#/jobs/playbook/30535/output

PR #9866, opened 2026-07-13
roles-infra/infra-openshift-cnv-resources/tasks/create_network.yaml - Fix overlapping config variable and add better config formatting

YoNoSoyVictor:
Fixes deployment issues such as https://prod0.apps.ocpv-infra01.dal12.infra.demo.redhat.com/#/jobs/playbook/47915/output

AAP passes a config extra var (catalog name like zero-touch-base-rhel) that was shadowing the task's config var, so the NAD ended up with garbage instead of CNI JSON. Renamed it to network_config and switched to a dict for better readability.


YoNoSoyVictor:
Works now: https://prod1.apps.ocpv-infra02.wdc07.infra.demo.redhat.com/#/jobs/playbook/35536/output

yordangit (approved):

wkulhanek (approved):

PR #9872, opened 2026-07-16
Add retry logic to assisted-installer wait_for_hosts

rhjcd:
Same fix as #9871 (host-ocp4-assisted-scale) applied to the host-ocp4-assisted-installer role.

The `wait_for_hosts` task can fail on transient Assisted Installer API errors. Adding `until/retries/delay` makes it retry up to 3 times with 20s between attempts.

Companion v2 change needed at: https://github.com/rhpds/agnosticd-v2/blob/5744e00fe05787e752334db87b0497cf8f5970ba/ansible/roles/host_ocp4_assisted_installer/tasks/setup_infrastructure.yml#L153


rhjcd:
@agonzalezrh same retry fix as #9871 but for the installer role. The v2 equivalent also needs this change at setup_infrastructure.yml#L153.

agonzalezrh (approved):

PR #9878, opened 2026-07-18
Replace unarchive with tar in ROSA destroy to fix sudo error

rhjcd:
## Summary

- The `ansible.builtin.unarchive` module internally calls `sudo` for its stat check even when `become: false` is set at the play level, causing every ROSA-based destroy job to fail with `sudo: command not found` inside the EE container
- Replaces the destroy-path `unarchive` task with a direct `tar` command, which avoids the module's internal stat behavior entirely
- The provision-path `unarchive` is unchanged (it uses `become: true` and needs root for ownership)

## Impact

- 140+ destroy failures per month on `ocp-dr-trident-binder` alone
- Affects all `rosa-consolidated` based catalog items
- Jira: [RHDPOPS-23828](https://redhat.atlassian.net/browse/RHDPOPS-23828)

## Error

```
TASK [Unzip rosa-linux.tar.gz]
ansible.errors.AnsibleError: Failed to get information on remote file (/tmp):
  /bin/sh: sudo: command not found
```

Confirmed on both prod0 (DAL12) and prod1 (WDC07) controllers across multiple guids.

## Test plan

- [ ] Deploy a `rosa-consolidated` based item on integration (e.g. trident-base)
- [ ] Destroy it and verify the ROSA CLI tar extraction succeeds
- [ ] Confirm destroy completes without `sudo: command not found`

### DIFF
--- a/ansible/configs/rosa-consolidated/install_rosa_cli.yml
+++ b/ansible/configs/rosa-consolidated/install_rosa_cli.yml
@@ -20,13 +20,9 @@
 
 - name: Unzip rosa-linux.tar.gz (destroy)
   when: ACTION != 'provision'
-  ansible.builtin.unarchive:
-    src: /tmp/rosa-linux.tar.gz
-    dest: "{{ rosa_binary_path }}"
-    remote_src: true
-    mode: u=rwx,go=rx
-    extra_opts:
-    - --no-same-owner
+  ansible.builtin.command:
+    cmd: tar xzf /tmp/rosa-linux.tar.gz --no-same-owner -C {{ rosa_binary_path }}
+    creates: "{{ rosa_binary_path }}/rosa"
 
 - name: cleanup archive file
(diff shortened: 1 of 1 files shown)


agonzalezrh (changes requested):
ocp-dr-trident-1.2.2 is a old not including Patrick's PR: https://github.com/redhat-cop/agnosticd/commit/c9c0a2dfbf5c0cdb08d90736319dca8767a76e9c

Not related about unarchive

rhjcd:
You're right — Patrick's fix in #9854 (c9c0a2d) already handles this properly by splitting provision/destroy paths. Our PR is redundant.

The fix just needs a new tag (e.g., ocp-dr-trident-1.3.1) cut after c9c0a2d, then the agnosticv prod.yaml scm_ref bumped from ocp-dr-trident-1.3.0. Closing this in favor of a tag bump.

PR #9820, opened 2026-06-25
[sandbox] [infra-dns] [infra_vmware_ibm_resources] Update vmware sandbox to support nsupdate (dnssandbox)

agonzalezrh:
##### SUMMARY

To keep consistent, the cluster name switches from guid to cluster-gluid, updated infra_vmware_ibm_resources to allow to use nsupdate 

##### ISSUE TYPE

- Feature Pull Request
- 
##### COMPONENT NAME
sandbox config
infra-dns role
infra_vmware_ibm_resources role

PR #9881, opened 2026-07-20
Skip Route53 DNS cleanup when credentials are not defined (sandbox-vmware)

rhjcd:
## Summary

- Adds `when: route53_aws_access_key_id is defined` guard to the Route53 DNS deletion task in `infra_vmware_ibm_resources`
- Prevents destroy failures when the catalog item uses DDNS instead of Route53

## Root cause

The `sandbox-vmware` catalog item uses DDNS for DNS management. The `dynamic-dns.yaml` include explicitly sets `route53_aws_access_key_id` (and related vars) to `{{ undef() }}` to avoid conflicts. During destroy, `delete_additional_public_ips.yaml:50` unconditionally tries to use these Route53 credentials, hitting "Mandatory variable has not been overridden."

The VMC equivalent (`infra-vmc-resources`) correctly uses `nsupdate` for DNS cleanup. The IBM version incorrectly uses `amazon.aws.route53`.

## Impact

- ~62 destroy failures/month on sandbox-vmware
- ~$2,760/month in leaked VMware IBM resources
- VMs, folders, NAT, and public IPs all clean up successfully -- only the DNS step crashes
- Part of RHDPOPS-23826 destroy failure remediation epic

## Test plan

- [ ] Deploy sandbox-vmware on integration and trigger destroy
- [ ] Verify NSX-T cleanup (VMs, NAT, public IPs) still works
- [ ] Verify Route53 DNS step is skipped (not failed)
- [ ] Confirm no regression on configs that DO use Route53 (e.g., ocp4-cluster on VMware IBM)

### DIFF
--- a/ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_additional_public_ips.yaml
+++ b/ansible/roles-infra/infra_vmware_ibm_resources/tasks/delete_additional_public_ips.yaml
@@ -48,6 +48,7 @@
   register: _public_ip_request
 
 - name: Delete DNS for additional dns
+  when: route53_aws_access_key_id is defined
   amazon.aws.route53:
     state: absent
     aws_access_key_id: "{{ route53_aws_access_key_id }}"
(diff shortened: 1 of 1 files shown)


rhjcd:
Updated to mirror the full pattern from PR #9820 (create side):

1. **Guard Route53 task** with `when: route53_aws_access_key_id is defined` (was missing on delete side)
2. **Add nsupdate delete task** with `when: cluster_dns_server is defined` (matches the nsupdate create task added in #9820)

PR #9820 added nsupdate support to `create_additional_public_ips.yaml` but missed the delete counterpart. This completes that migration for the destroy path.

Root cause: `dynamic-dns.yaml` include explicitly sets `route53_aws_access_key_id` to `{{ undef() }}` for DDNS sandboxes, causing "Mandatory variable has not been overridden" on every destroy.

agonzalezrh (approved):

PR #9889, opened 2026-07-24
Support exclude parameter in CUDA repo configuration

bbethell-1:
## Summary
- Passes through the `exclude` field from repo config to `yum_repository` module in the `setup_nvidia_cuda` role
- Allows consumers to exclude specific packages (e.g. `cccl-13*` which obsoletes `cuda-cccl-12-8` and breaks CUDA 12.x installations)
- Uses `default(omit)` so existing configurations are unaffected

## Context
NVIDIA released `cccl-13-3` which obsoletes `cuda-cccl-12-8`, causing depsolve failures during RHAIIS provisioning. The CUDA repo needs an `exclude=cccl-13*` directive to prevent the obsoleting package from being considered. This change enables that by passing the `exclude` parameter through to the `yum_repository` module.

## Test plan
- [ ] Deploy RHAIIS catalog item with `exclude: "cccl-13*"` in repo config
- [ ] Verify CUDA toolkit and all dependencies install without depsolve errors
- [ ] Verify existing catalog items without `exclude` are unaffected


YoNoSoyVictor:
image
This worked and deployed the item successfully

YoNoSoyVictor (approved):

PR #9897, opened 2026-08-03
Fix Azure DevOps cleanup loop_control.label template error

bbethell-1:
## What Changed
Fixed loop_control.label template error in Azure DevOps resource cleanup task

## Why
Task was failing with: `Failed to template loop_control.label: 'dict object' has no attribute 'name'`

## Files Changed
- `ansible/roles/open-env-azure-remove-user-from-subscription/tasks/empty_resource_group.yml`:
  - Line 35: `{{ resource.name }}` → `{{ resource.name | default(resource.id | default('unnamed resource')) }}`

## Fix Details
The label template now uses a default filter chain:
1. Try `resource.name` first
2. Fall back to `resource.id` if name is missing
3. Fall back to `'unnamed resource'` if both are missing

This prevents template errors while still providing useful labels when attributes are available.

## Tested
- [x] Destroy job: https://prod1.apps.ocpv-infra02.wdc07.infra.demo.redhat.com/#/jobs/playbook/92745/output
- [ ] Integration deploy via rhpds/agnosticv#27661
- [ ] Cleanup task handles missing resource.name

## Impact
Fixes Azure subscription cleanup failures when Azure DevOps resources don't have name attribute

## Jira
GPTEINFRA-17534

### DIFF
--- a/ansible/roles/open-env-azure-remove-user-from-subscription/tasks/empty_resource_group.yml
+++ b/ansible/roles/open-env-azure-remove-user-from-subscription/tasks/empty_resource_group.yml
@@ -32,7 +32,7 @@
   loop: "{{ r_resources.response }}"
   loop_control:
     loop_var: resource
-    label: "{{ resource.name }}"
+    label: "{{ resource.name | default(resource.id | default('unnamed resource')) }}"
 
 - name: Warn if Azure DevOps account could not be deleted (billing still linked)
   when:


YoNoSoyVictor:
To my understanding, this works to unblock the task by stopping Ansible from crashing when resource.name is missing.

The underlying issue is that we are targeting the wrapper, `azure_rm_resource_info` returns response: `[{"value": []}]`. We’re looping that wrapper object `[{"value": []}]`, not real resources `[]`.

A better approach would be to normalize before looping, and then loop the list which is what the tasks were aiming to do to begin with.

bbethell-1:
## Note: Workaround, Not Root Fix

This PR unblocks the immediate crash, but doesn't fix the underlying issue.

**Root cause:** The code loops `r_resources.response` which returns the wrapper `[{"value": []}]` instead of the actual resource list.

**Current fix:** Adds default filters to prevent crash when `resource.name` is missing
**Proper fix:** Normalize the data structure before looping:

```yaml
- name: Normalize resources list
  set_fact:
    resources_list: "{{ r_resources.response[0].value | default([]) }}"

- name: Delete Azure DevOps resources
  loop: "{{ resources_list }}"
  # ... resource.name will now exist
```

This would properly loop over the actual resources, not the wrapper object.

Happy to update this PR with the proper fix if we want to do it now, or we can merge this as-is to unblock and file a follow-up issue for the refactor.

bbethell-1:
Let's merge and tag this for prod - it's well tested in current state and unblocks the immediate issue.

For the proper fix (normalizing the resources list structure), we can tackle that after tagging.

YoNoSoyVictor (approved):

PR #9907, opened 2026-08-09
Fix workload ocp4_workload_s3_model_copy_to_minio

ryanj:
##### SUMMARY

Fix boto3 error when attempting to copy models to s3 using awscli

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ocp4_workload_s3_model_copy_to_minio

##### ADDITIONAL INFORMATION
Error message found in the logs of `copy-models-to-minio` job via the Java4AI workshop:
```ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.```

This change fixes the boto3 dependency issue and encourages curl to follow redirects when needed.


oladapooloyede (approved):
Tested the fix and it worked well

rhjcd (approved):
lgtm

PR #9914, opened 2026-08-12
parametrize images and pre-pull in coco

esposem:
##### SUMMARY

Parametrize a couple of things as they are going to change over time and be optional

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ocp4_workload_coco_on_aro
##### ADDITIONAL INFORMATION

```paste below

```


rickgcv:
The logic looks good. One thing: the default for `ocp4_workload_coco_on_aro_coco_tools_image` in `defaults/main.yml` should be `0.5.1` rather than keeping `0.3.0` — anyone using the role outside of agnosticv will otherwise silently get the old image.

rickgcv (approved):
LGTM

bbethell-1 (approved):

PR #9923, opened 2026-09-07
open-environment-azure-subscription: disable sudo for the bastion lab user

rut31337:
## Summary
Disables sudo for the interactive bastion user in the `open-environment-azure-subscription` config. An ARO lab user does not need root on the bastion; leaving passwordless sudo enabled lets any attendee escalate to root (`sudo su -`) and read whatever the bastion holds. Part of PSIRTSUPT-23145 remediation.

## Changes
- **`post_software.yml`** — new final bastion play that drops a validated `/etc/sudoers.d/zz-disable-<user>-sudo` override revoking sudo for the interactive user. It runs last (all prior provisioning still needs sudo), uses `visudo -cf` validation, and overrides (rather than deletes) the provisioning-supplied grant so it survives cloud-init regeneration on reboot.
- **`default_vars.yml`** — `bastion_disable_user_sudo: true` (default on; set false for labs that need post-provision sudo).

### DIFF
--- a/ansible/configs/open-environment-azure-subscription/default_vars.yml
+++ b/ansible/configs/open-environment-azure-subscription/default_vars.yml
@@ -95,6 +95,11 @@ ansible_user: "{{ bastion_user_name }}"
 # Remove motd (register for insights) prompt from bastion
 bastion_remove_insights_motd: false
 
+# Disable sudo for the interactive bastion user at the end of provisioning.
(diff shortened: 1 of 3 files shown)


agonzalezrh (approved):

bbethell-1 (commented):
LgTm

